PR #13, opened 2019-02-04
Initially adding contributers to the cli-wg

spacekookie:
This PR adds a bunch of members from the CLI-WG as regular members.
Don't merge this yet before everybody has had a chance to update
their information to their liking.

Pinging @killercup retroactively

### DIFF
--- a/people/codesections.toml
+++ b/people/codesections.toml
@@ -1,2 +1,2 @@
-name = "Daniel Sockwell "
+name = "Daniel Sockwell"
 github = "codesections"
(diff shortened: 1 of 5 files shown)


spacekookie:
@pietroalbini I think you can merge this now

people/codesections.toml
```diff
@@ -0,0 +1,2 @@
+name = "Daniel Sockwell "
```
pietroalbini: Trailing whitespace.

PR #15, opened 2019-02-05
Update and rename oe.toml to komaeda.toml

komaeda:
`oe` has been deleted and this is my account now. This was already on the website before the structure update, not sure why it got reverted?


pietroalbini:
Woops, I probably forgot to sync that file. Sorry!

PR #18, opened 2019-02-06
Fix production team mailing list

pietroalbini:
r? @aidanhs


aidanhs:
@pietroalbini looks good

PR #29, opened 2019-03-04
Add Rust By Example

marioidival:
Add missing information of Rust By Example on team references

[old site](https://prev.rust-lang.org/en-US/team.html#Documentation-peers)
https://github.com/rust-lang/rust/blob/master/src/tools/publish_toolstate.py#L24

### DIFF
--- a/teams/docs.toml
+++ b/teams/docs.toml
@@ -23,3 +23,7 @@ extra-teams = ["reference"]
 [[lists]]
 address = "doc-team@rust-lang.org"
 extra-teams = ["reference"]
+
+[[lists]]
+address = "doc-team@rust-lang.org"
+extra-teams = ["rust-by-example"]
--- a/teams/rust-by-example.toml
+++ b/teams/rust-by-example.toml
@@ -6,7 +6,6 @@ leads = ["steveklabnik"]
 members = [
     "steveklabnik",
     "marioidival",
-    "projektir",
 ]
 
 [website]
--- a/teams/rust-by-example.toml
+++ b/teams/rust-by-example.toml
@@ -10,4 +10,4 @@ members = [
 
 [website]
 name = "Rust by Example team"
-description = "working on the rust by example"
+description = "maintaining and updating Rust By Example"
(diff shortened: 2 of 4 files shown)


marioidival:
ping @steveklabnik @QuietMisdreavus

teams/docs.toml
```diff
@@ -23,3 +23,7 @@ extra-teams = ["reference"]
 [[lists]]
 address = "doc-team@rust-lang.org"
 extra-teams = ["reference"]
+
+[[lists]]
+address = "doc-team@rust-lang.org"
+extra-teams = ["rust-by-example"]
```
QuietMisdreavus: Pretty sure you can just add this to the existing `doc-team@rust-lang.org` section:

```
extra-teams = ["reference", "rust-by-example"]
```

marioidival: Fixed! :)

pietroalbini:
projektir is missing.

QuietMisdreavus:
@projektir We're recreating the "Rust By Example" subteam in the new teams infrastructure, and it seems like we're missing an email from you. Did you want to be re-added to the docs-team mailing list?

steveklabnik:
So, part of the reason I originally moved this over was, around that time, it seemed like you all had stopped working on things. I see you're back at it again, so this seems good :)

projektir:
I appreciate being remembered, but, tbh, I think y'all should just remove me. I don't know when/if I'll be back to working on RBE again.

marioidival:
@pietroalbini @QuietMisdreavus

teams/rust-by-example.toml
```diff
@@ -0,0 +1,13 @@
...
+[people]
+leads = ["steveklabnik"]
+members = [
+    "steveklabnik",
+    "marioidival",
+]
+
+[website]
+name = "Rust by Example team"
+description = "working on the rust by example"
```
QuietMisdreavus: Probably a better description here would be something like "maintaining and updating Rust By Example".

QuietMisdreavus:
If the description is updated, this looks good to me. I can't merge PRs on this repo, though. :sweat_smile:

teams/rust-by-example.toml
marioidival: done!

steveklabnik:
Thanks!

PR #30, opened 2019-03-04
stop brson, alercah from getting mail

nikomatsakis:
Per their requests, this should stop them from getting mail from the `all@` mailing list:

- clear the e-mail fields from their people files
- in the case of @alercah, move from reference to alumni (I believe this is correct)

(cc @brson, @alercah, @steveklabnik)


steveklabnik:
Yes, that is correct.

PR #32, opened 2019-03-08
Devtools restructure

Manishearth:
See https://internals.rust-lang.org/t/tools-team-changes/9569

### DIFF
--- a/teams/devtools.toml
+++ b/teams/devtools.toml
@@ -1,19 +1,23 @@
 name = "devtools"
 
 [people]
-leads = []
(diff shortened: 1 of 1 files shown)


pietroalbini:
You hit two issues: first the error was broken when an user didn't exist, and also I mistakenly imported you as `manishearth` instead of `Manishearth` (it's case sensitive). I pushed two commits on master to fix those issues, if you capitalize your username in the members list this should be green.

Manishearth:
done!

PR #33, opened 2019-03-09
Add self to core, remove from mod team

Manishearth:
See https://internals.rust-lang.org/t/new-members-for-the-rust-core-team/9575

r? @nikomatsakis

I'm not sure if I should add @skade here too (I'm not sure if we need to mark observers separately).


skade:
@Manishearth added an issue for this https://github.com/rust-lang/team/issues/35

PR #34, opened 2019-03-10
Import bors permissions

pietroalbini:
This PR adds support for bors permissions in the schema and imports the current ones. The permissions are the same as the current `homu.toml.template`, with those exceptions:

* @ashleygwilliams gains review access to rustc since she's on the core team.
* @sgrif gains review access to rustc since he's on the infra team.
* @onur gains review access to rustc since they're on the rustdoc team.
* @rylev and @celaus gains review access to rustc since they're on the docs team.
* The whole infra team is allowed to review Crater PRs, instead of just Aidan, Mark and me.
* @killercup gains review access to clippy since he's in the clippy team.

Synchronization with the bors config is not yet implemented, but I plan to write that soon.

### DIFF
--- a/people/Aaronepower.toml
+++ b/people/Aaronepower.toml
@@ -1,3 +1,6 @@
 name = "Aaron Power"
 github = "Aaronepower"
 email = "theaaronepower@gmail.com"
+
+[permissions]
+bors.rust.review = true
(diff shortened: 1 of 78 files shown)


Manishearth:
The perms are split across two spots (teams and users) here, are you sure we want that? May be nicer to have an extra-reviewers section for teams? Slight suggestion, feel free to ignore.

PR #38, opened 2019-03-13
Add skade as a core team observer

Manishearth:
r? @skade @steveklabnik @pietroalbini

I'm not sure if I've done the TOML stuff correctly, make sure you wait for Pietro to sanity-check the format before merging.

### DIFF
--- a/teams/core-observers.toml
+++ b/teams/core-observers.toml
@@ -1,0 +1,15 @@
+name = "core-observers"
+subteam-of = "core"
+
+[people]
+leads = []
+members = [
+    "skade",
+]
+
+[permissions]
+bors.rust.review = true
+
+[website]
+name = "Core Team Observers"
+description = "observers on the core team"
--- a/people/Manishearth.toml
+++ b/people/Manishearth.toml
@@ -1,6 +1,3 @@
 name = "Manish Goregaokar"
 github = "Manishearth"
 email = "manishsmail@gmail.com"
-
-[permissions]
-bors.rust.review = true
--- a/teams/core.toml
+++ b/teams/core.toml
@@ -24,9 +24,11 @@ weight = 1000
 
 [[lists]]
 address = "core@rust-lang.org"
+extra-teams = ["core-observers"]
 
 [[lists]]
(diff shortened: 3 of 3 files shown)


teams/core-observers.toml
```diff
@@ -0,0 +1,14 @@
+name = "coore-observers"
+subteam-of = "core"
+
+[people]
+leads = []
+members = [
+    "skade",
+]
+
+[permissions]
```
pietroalbini: Just a note, the core team has the `bors.rust.review` permission. Not sure if you want to put that here.

teams/core-observers.toml
```diff
@@ -0,0 +1,14 @@
+name = "coore-observers"
```
pietroalbini: Wrong name.

pietroalbini:
Uuuh, you also need to remove the `bors.rust.permission` from yourself. This wasn't a problem when you first created the other PR, but a change got merged in the meantime that required it.

teams/core-observers.toml
Manishearth: I don't know either :)

Manishearth:
Ah, updated. Though I guess we should allow overlap this way: team memberships can change.

(But eventually I guess we should be creating buckets like journeypeople for these permissions so that individual perms aren't necessary anymore)

teams/core-observers.toml
nikomatsakis: I think it's fine. I would expect that -- like anyone with r+ -- there is an expectation that people will only use it where appropriate. But I could see a core team observer being delegated some task by core team that involves r+.

Manishearth: Added. Ready to merge, anyone want to push the button?

teams/core-observers.toml
```diff
@@ -0,0 +1,15 @@
...
+members = [
+    "skade",
+]
+
+[permissions]
+bors.rust.review = true
+
+[website]
+name = "Core Team Observers"
+description = "observers on the core team"
```
Manishearth: this text could be improved maybe

PR #40, opened 2019-03-20
Add rustfix WG

killercup:


Manishearth:
https://github.com/rust-lang/team/blob/master/teams/devtools.toml#L19-L30 should be updated too

Manishearth:
r=me with that change

killercup:
Good catch! Updated

Manishearth:
also, no lead?

Manishearth:
Doesn't have to have one, but I assumed you were the lead of the team :)

killercup:
Well, we haven't talked about having a leader. While I did lead the
meeting, I feel like we should at least discuss the topic before someone
(like me) takes over this team :D

Manishearth:
Ah. r=me either way, @pietroalbini can you merge this?

PR #50, opened 2019-03-30
Add rfcbot data to the Team API

pietroalbini:
This PR adds the data needed to integrate rfcbot with the repo in the Team API. The actual integration still needs to be done.

cc @Centril @anp


Centril:
Could you also add T-release while at it?

PR #53, opened 2019-04-09
Remove inactive reviewers

Manishearth:
When bors permissions were migrated to this repo, non-team members with review access were grandfathered over.

@pietroalbini did a bit of an audit of these and found a bunch of people with review permissions who haven't exercised them in at least a year. In the interest of scoping down permissions, we're removing review permissions from these folks.

There are still many non-team reviewers left who have been active in the last year. We're hoping to have some sort of way to fit this into the team structure (the compiler team is thinking about a concept of "journeypeople" which may help), however that is not the goal of this PR.

If you have had your permissions revoked due to this pull request and wish to become an active reviewer again, let us know!

r? @pietroalbini


eddyb:
The only person in this list who I think is active is @matklad.

matklad:
I've actually never knew I had r+, and I don't think I can use it effectively: I am still novice when it comes to compiler/stdlib.

PR #63, opened 2019-04-26
Add hashbrown to homu

Manishearth:
See also: https://github.com/rust-lang/rust-central-station/pull/204

r? @pietroalbini

cc @Amanieu


pietroalbini:
Thanks!

PR #68, opened 2019-05-13
Some text cleanup for the website

skade:
Fixes https://github.com/rust-lang/www.rust-lang.org/issues/499


pietroalbini:
Thanks!

PR #69, opened 2019-05-14
Add self to wg-grammar

CAD97:
Looks like we probably want to add at least @eddyb as well, potentially @matklad

r? @pietroalbini


pietroalbini:
@centril @qmx?

PR #70, opened 2019-05-23
Allow dumping website data as FTL files

Manishearth:
I'm working on internationalizing the website (https://github.com/rust-lang/www.rust-lang.org/pull/792), and for that eventually the website text data will have to come from an FTL file.

My implementation has fallback to the website data stored in this repo, but I need this tool so that I can generate an English FTL file that others can translate.

Since the team name is being used in Handlebars templates, I'm also forcing the internal identifier used for the team to be alphanumeric (plus dashes)

r? @pietroalbini


pietroalbini:
The PR looks good, but what's the update procedure when a new team is added or when a team's description is updated? The goal of this repo would be that any team-related change could be done by sending a PR here, does this add other steps?

The new command should also be documented in the readme.

Manishearth:
> The PR looks good, but what's the update procedure when a new team is added or when a team's description is updated? The goal of this repo would be that any team-related change could be done by sending a PR here, does this add other steps?

So if you add a team here it will Just Work on the website, but translators will not be able to translate it until you rerun dump-website.

Manishearth:
Basically, the _English_ site uses the team data (and any language without translations for team data). However the translation tools only understand FTL files, so we create this file for the benefit of translators.

Manishearth:
Done

PR #76, opened 2019-06-21
manage miri review permissions: t-compiler and RalfJung

RalfJung:
I hope I did this right. Also see https://github.com/rust-lang/rust-central-station/pull/265.

Cc @oli-obk


pietroalbini:
Yes, thanks!

PR #82, opened 2019-07-02
remove aturon from extra infra roles

aturon:


Manishearth:
cc @pietroalbini @aidanhs we should probably add more people here for redundancy

pietroalbini:
@alexcrichton we should have more than just yourself in those aliases so we're not blocked if you're on vacation. Should we add @aidanhs?

alexcrichton:
Sounds good to me yeah!

PR #84, opened 2019-07-04
wg-triage: show it on the website

pietroalbini:
The WG needs to be under the `operations` meta-team to be shown on the website.

r? @Mark-Simulacrum


pietroalbini:
Also added a check at the validation level to prevent subteams of subteams, which are not supported by the website.

PR #88, opened 2019-07-17
Add GitHub teams in the schema and the API

pietroalbini:
This PR is the first step of the synchronization of GitHub teams with the team repo. This adds all the relevant fields to the schema and the API, and "enables" synchronization for the release team (chosen because it's already synchronized so no accidental changes will be applied to GitHub).

After the PR is merged the next steps are sending a PR to the central station adding the synchronization tool (already working in a local branch), creating a bot account to do the work and then enabling synchronization for all teams. The last step will require coordination with the team leads to make sure no unwanted changes are applied.

cc #2

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -223,7 +223,7 @@ impl Team {
     pub(crate) fn github_teams(&self) -> Vec<(&str, &str)> {
         if let Some(github) = &self.github {
             let name = github
-                .name
+                .team_name
                 .as_ref()
                 .map(|n| n.as_str())
                 .unwrap_or(&self.name);
@@ -250,7 +250,7 @@ struct TeamPeople {
 #[derive(serde::Deserialize, Debug)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
 struct GitHubData {
-    name: Option<String>,
+    team_name: Option<String>,
     orgs: Vec<String>,
 }
 
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -7,11 +7,16 @@ use std::collections::HashSet;
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
 pub(crate) struct Config {
     allowed_mailing_lists_domains: HashSet<String>,
+    allowed_github_orgs: HashSet<String>,
 }
 
 impl Config {
     pub(crate) fn allowed_mailing_lists_domains(&self) -> &HashSet<String> {
         &self.allowed_mailing_lists_domains
+    }
+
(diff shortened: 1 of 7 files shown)


src/schema.rs
```diff
@@ -228,6 +247,13 @@ struct TeamPeople {
     include_all_team_members: bool,
 }
 
+#[derive(serde::Deserialize, Debug)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+struct GitHubData {
+    name: Option<String>,
```
Mark-Simulacrum: Could this struct be something like `GithubTeam` or at least a comment added on `name` saying that this is the team name?

src/static_api.rs
```diff
@@ -48,6 +48,9 @@ impl<'a> Generator<'a> {
             members.sort_by_key(|member| member.github.to_lowercase());
             members.sort_by_key(|member| !member.is_lead);
 
+            let mut github_teams = team.github_teams();
+            github_teams.sort();
```
Mark-Simulacrum: What do we gain by this? Determinism?

teams/release.toml
```diff
@@ -25,6 +25,9 @@ crater = true
 perf = true
 bors.rust.review = true
 
+[github]
+orgs = ["rust-lang"]
```
Mark-Simulacrum: As a relatively simple additional test of this functionality, could we also add `rust-lang-nursery` here and automatically create the release team there as well?

src/static_api.rs
pietroalbini: Yeah, all the things outputted by the API are currently deterministic. Other than generally being nicer, the synchronization code on RCS will run only if the API content changed, and having it deterministic avoids unnecessary runs.

teams/release.toml
pietroalbini: I wanted this PR to be a noop, that's why I chose the release team on `rust-lang`. Do you think we should start right away with making changes?

src/schema.rs
pietroalbini: Maybe it's best to rename it:

```toml
[github]
team-name = "release-alt"
orgs = ["rust-lang"]
```

Mark-Simulacrum: Yes, that works too!

teams/release.toml
Mark-Simulacrum: I guess it does make sense to leave this change for later. Maybe we can make that the first test though -- since it's relatively low risk and tests team creation.

src/schema.rs
pietroalbini: Renamed it to `team-name`.

PR #92, opened 2019-07-20
Sync the clippy team with GitHub

pietroalbini:
Changes to the teams that are going to be applied:

```
rust-lang/clippy: user Manishearth is missing, adding them...
rust-lang/clippy: user oli-obk is missing, adding them...
rust-lang/clippy: user killercup is missing, adding them...
rust-lang/clippy: user llogiq is missing, adding them...
```

r? @Manishearth

### DIFF
--- a/teams/clippy.toml
+++ b/teams/clippy.toml
@@ -13,6 +13,9 @@ members = [
 [permissions]
 bors.clippy.review = true
 
+[github]
+orgs = ["rust-lang"]
+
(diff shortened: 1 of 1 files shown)


phansch:
While we're at it, would it make sense to update the GitHub team with [current contributors](https://github.com/rust-lang/rust-clippy/graphs/contributors)? I'm not sure who's responsible for that, though.

PR #94, opened 2019-07-27
Fix compiler working group links

carols10cents:
As reported [on URLO](https://users.rust-lang.org/t/broken-links-on-compiler-team-page-on-rust-website/30727), most of the "[team] repository" buttons for the compiler working groups [on this page](https://www.rust-lang.org/governance/teams/compiler) are broken.

Looks like in https://github.com/rust-lang/compiler-team/pull/103 the compiler team switched from using github as a website to using github pages as a website :)

### DIFF
--- a/teams/wg-async-await.toml
+++ b/teams/wg-async-await.toml
@@ -9,4 +9,4 @@ members = ["nikomatsakis", "cramertj"]
 [website]
 name = "Async-await Implementation"
 description = "Implementing async-await"
-repo = "https://github.com/rust-lang/compiler-team/blob/master/working-groups/async-await"
+repo = "https://rust-lang.github.io/compiler-team/working-groups/async-await/"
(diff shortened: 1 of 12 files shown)


spastorino:
@carols10cents the docs part of the url was introduced unintentionally. We should remove them from compiler-team repo see https://github.com/rust-lang/compiler-team/pull/103#issuecomment-515128273

carols10cents:
Aha, thanks! I'll wait until that gets fixed then I'll update this PR.

teams/wg-async-await.toml
```diff
@@ -9,4 +9,4 @@ members = ["nikomatsakis", "cramertj"]
 [website]
 name = "Async-await Implementation"
 description = "Implementing async-await"
-repo = "https://github.com/rust-lang/compiler-team/blob/master/working-groups/async-await"
+repo = "https://rust-lang.github.io/compiler-team/docs/working-groups/async-await/"
```
spastorino: URLs now look like https://rust-lang.github.io/compiler-team/working-groups/async-await/

spastorino:
@carols10cents now yes, the docs part is out of URLs. The right changes in this PR would be what you have without the docs part.

carols10cents:
Fixed! This should be ready to merge now :)

spastorino:
Ready to merge, CI failures are unrelated.

PR #100, opened 2019-08-26
Update roster with the latest information

therealprof:
Taken directly from  from https://github.com/orgs/rust-embedded/people


therealprof:
Superseded by #101

PR #104, opened 2019-09-01
Move clippy reviewers to team, add Yaah, remove Georg

Manishearth:
We don't use the clippy "team" much for anything special, but folks with review should be on it, they all deserve it :)

This also moves @birkenfeld to alumni since he's been inactive for two years, but he's free to rerequest r+ whenever he wishes.

This also adds @yaahc to the team/reviewers, she's been doing excellent work on the cargo side lately and should be a part of this.

r? @oli-obk

### DIFF
--- a/people/birkenfeld.toml
+++ b/people/birkenfeld.toml
@@ -1,6 +1,4 @@
 name = "Georg Brandl"
 github = "birkenfeld"
 github-id = 144359
-
-[permissions]
-bors.clippy.review = true
+email = "georg@python.org"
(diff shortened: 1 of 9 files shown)


people/birkenfeld.toml
```diff
@@ -2,5 +2,3 @@ name = "Georg Brandl"
 github = "birkenfeld"
 github-id = 144359
 
-[permissions]
-bors.clippy.review = true
```
oli-obk: needs a mail address, too (according to CI)

Manishearth: oh oops

Manishearth: Done

PR #106, opened 2019-09-03
teams/rust-by-example: synchronize with GitHub

pietroalbini:
This will create a new `rust-by-example` team in the `rust-lang` GitHub organization with steveklabnik and marioidival. Then we should configure the rust-by-example repo to authorize the team instead of individual members.

r? @steveklabnik

### DIFF
--- a/teams/rust-by-example.toml
+++ b/teams/rust-by-example.toml
@@ -8,6 +8,9 @@ members = [
     "marioidival",
 ]
 
+[github]
+orgs = ["rust-lang"]
+
 [website]
 name = "Rust by Example team"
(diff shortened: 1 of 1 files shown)


steveklabnik:
r+! (I don't think we actually use bors here, right? Should I just merge?)

Manishearth:
Yes

pietroalbini:
Permissions tweaked on the `rust-lang/rust-by-example` repository to avoid using collaborators and instead give write access to the new team. @marioidival you should have received an automatic invitation to join the `rust-lang` org, please accept it to gain write access to the RBE repo back!

PR #115, opened 2019-09-16
Switch out michaelwoerister for aturon in wg-parallel-rustc

michaelwoerister:
r? @nikomatsakis 
cc @aturon


nikomatsakis:
It seems like this switch is not going to happen after all. Closing for now.

PR #116, opened 2019-09-16
Allow rust-timer to trigger bors try

Mark-Simulacrum:


Mark-Simulacrum:
This is in preparation for work to enable rust-timer to automatically enqueue a try build and then trigger a perf run once the try build is complete. The workaround would be to force users to write something like `@bors try @rust-timer queue` which seems not great, but workable, if we don't want to do this.

Mark-Simulacrum:
Came up with a slight alternative approach, will pivot to that.

PR #117, opened 2019-09-18
add JoshTriplett as lang team lead

nikomatsakis:
As announced in [this internals thread][thread].

[thread]: https://internals.rust-lang.org/t/announcing-josh-triplett-as-co-lead-of-lang-team/10965

r? @pietroalbini


pietroalbini:
Not sure why I'm r?ed but it looks fine!

PR #122, opened 2019-09-25
Create mailing list for all hands planning

Manishearth:
Unsure if this is the appropriate section, but it seemed correct.

r? @nikomatsakis


nikomatsakis:
The workflow error seems weird. Anyway, this looks good to me -- but r? @pietroalbini to be sure it's the right location.

pietroalbini:
Well that was weird. Pushed an empty commit to this branch and a build started successfully. Contacted GitHub Support about this.

PR #123, opened 2019-09-26
Add icebreakers-llvm team

spastorino:
We've [discussed on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/185694-t-compiler.2Fwg-meta) about adding icebreakers-llvm as a team with minimal everything (no github, no website) as can be seen on the toml file, instead of doing this as a working group.

From what we've quickly seen on team repo by doing what I've done in the PR this group would end included in `all@`. What would be the best way to avoid that?.

/cc @nikomatsakis @pietroalbini @Mark-Simulacrum (this is also related to https://github.com/rust-lang/triagebot/issues/169)

### DIFF
--- a/teams/icebreakers-llvm.toml
+++ b/teams/icebreakers-llvm.toml
@@ -1,0 +1,9 @@
+name = "icebreakers-llvm"
+marker-team = true
+
+[people]
+leads = []
+members = [
+    "nagisa",
+    "nikic",
+]


pietroalbini:
So, the way I'd go to exclude them from `all@` would be to create another state similar to `wg = true`. Bikeshedding is really welcome, maybe we should go with something like `marker-team = true` (I hate that name btw)?

You'll need to add [a new field to the schema](https://github.com/rust-lang/team/blob/3be7654fe4eb607cdd8e072de3053132d0732899/src/schema.rs#L110-L111) and use it [to tweak how team members are collected](https://github.com/rust-lang/team/blob/3be7654fe4eb607cdd8e072de3053132d0732899/src/schema.rs#L163). A [validation rule](https://github.com/rust-lang/team/blob/master/src/validate.rs) to prevent `wg = true` with `marker-team = true` would also be nice. The most annoying part of this will be [to expose the new state in the API](https://github.com/rust-lang/team/blob/3be7654fe4eb607cdd8e072de3053132d0732899/rust_team_data/src/v1.rs#L6-L11).  Unfortunately it's an enum at the moment and I'm pretty sure [some integrations](https://github.com/rust-lang/team#rust-teams-structure) are exaustively matching it (at least the website does). You'll need to audit them to make sure they won't break, and ideally add a `__NonExaustive` to prevent this issue from happening in the future.

Just find a better name for `marker-team = true` please :stuck_out_tongue:

spastorino:
@pietroalbini I've added some code, check it out and let me know if I'm lacking some stuff.

rust_team_data/src/lib.rs
```diff
@@ -1 +1,3 @@
+#![feature(non_exhaustive)]
```
spastorino: I guess this is not acceptable for a repo like this ...

spastorino:
So this PR is back to just adding the team, the infra work is done in https://github.com/rust-lang/team/pull/135/files

teams/icebreakers-llvm.toml
```diff
@@ -0,0 +1,9 @@
+name = "icebreakers-llvm"
+marker_team = true
```
pietroalbini:
```suggestion
marker-team = true
```

pietroalbini:
Integrations to update:

* [ ] Crater: https://github.com/rust-lang/crater/pull/471
* [ ] Perf
* [ ] rfcbot
* [ ] GitHub teams membership
* [ ] Mailing lists
* [ ] Website

PR #131, opened 2019-10-03
Add website and GH sync for the security team

pietroalbini:
This adds the security team to the governance section of the website and enables synchronization with GitHub's `@rust-lang/security`.

Fixes https://github.com/rust-lang/security-team/issues/4
r? @alexcrichton


alexcrichton:
Oh oops sorry I didn't realize github sync was enabled through here, that makes more sense!

PR #134, opened 2019-10-03
add simulacrum, centril to compiler-contributors

nikomatsakis:
Add @Mark-Simulacrum and @Centril to the "compiler contributors" team, both of whom have contributing more and more to the compiler lately. A big ❤️ thank you ❤️ to both of them! 🎉


teams/compiler-contributors.toml
```diff
@@ -12,6 +13,7 @@ members = [
   "nikic",
   "nnethercote",
   "scalexm",
+  "Mark-Simulacrum",
```
spastorino: Seems like your intention was to alphabetically sort names, but given that's not simulacrum but Mark-Simulacrum should be placed a bit above :smile:

teams/compiler-contributors.toml
```diff
@@ -4,6 +4,7 @@ subteam-of = "compiler"
 [people]
 leads = []
 members = [
+  "centril",
```
JohnTitor:
```suggestion
  "Centril",
```

teams/compiler-contributors.toml
nikomatsakis: ha thanks, I originally had him in there "simulacrum"...

spastorino: Yeah, I guessed that that was the issue :).

PR #140, opened 2019-10-10
add nikomatsakis to LLVM ICE-breakers

nikomatsakis:
This PR is meant to serve as an example! To add yourself to the LLVM ICE-breaker team, just create a commit like this one, but adding your own username to the file instead of `nikomatsakis`.

Also, if you are not already a member of a Rust team then -- in addition to adding your name to this file -- you have to checkout the repository, run the command below and commit the new file it created

```bash
cargo run add-person $your_user_name
```

### DIFF
--- a/teams/icebreakers-llvm.toml
+++ b/teams/icebreakers-llvm.toml
@@ -6,4 +6,5 @@ leads = []
 members = [
     "nagisa",
     "nikic",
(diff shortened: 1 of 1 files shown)


nikomatsakis:
(This PR is just meant to serve as an example, so I'm going to close it without merging.)

PR #150, opened 2019-10-16
Give smarnach bors permissions on crates.io.

smarnach:
@sgrif This change gives me `bors r+` permissions for crates.io.


pietroalbini:
Can we just give r+ perms to the whole team?

PR #155, opened 2019-10-17
add escstatic-morse to compiler contributors

nikomatsakis:
They're working to make compile-time evaluation more expressive by enabling `if`, `match` and other control flow in constants. As one of their first major contributions, they implemented a dataflow analysis to validate the bodies of `const`s and `const fn`s (rust-lang/rust#64470).

Congratulations @ecstatic-morse, and thanks!

cc @rust-lang/compiler

### DIFF
--- a/teams/compiler-contributors.toml
+++ b/teams/compiler-contributors.toml
@@ -6,6 +6,7 @@ leads = []
 members = [
   "Centril",
   "davidtwco",
+  "esctatic-morse",
   "flodiebold",
   "lqd",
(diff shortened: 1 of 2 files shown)


teams/compiler-contributors.toml
```diff
@@ -6,6 +6,7 @@ leads = []
 members = [
   "Centril",
   "davidtwco",
+  "esctatic-morse",
```
ecstatic-morse:
```suggestion
  "ecstatic-morse",
```

PR #167, opened 2019-10-24
sync lang team

nikomatsakis:
This synchronizes @rust-lang/lang as defined on team with the github repository.

This will cause

* @huonw
* @solson

to be removed, and @scottmcm to be added.

I'm also removing the shepherds subteam, as it is not really the way we're operating anymore. (I've discussed this with the participants / lang-team already.)

cc https://github.com/rust-lang/wg-governance/issues/21

### DIFF
--- a/teams/alumni.toml
+++ b/teams/alumni.toml
@@ -22,7 +22,10 @@ members = [
     "jseyfried",
     "pcwalton",
     "peschkaj",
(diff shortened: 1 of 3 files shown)


huonw:
Bye 😄

I hadn't realised I was still part of the team... but it explains the plethora of notifications I kept getting. Thanks!

PR #170, opened 2019-10-25
Nicolette is resigning from the moderation team 😢

carols10cents:


matthieu-m:
I find it sad to see her name completely removed.

Should there be an alumni section to acknowledge past members of the teams and their contributions?

pietroalbini:
Well, we have an alumni team.

carols10cents:
OOPS! Fix incoming to move Nicolette to Alumni instead!

PR #172, opened 2019-10-25
Add mark-i-m as co-lead and remove nikomatakis from learning wg

spastorino:
I'd like to propose @mark-i-m as WG learning co-lead. Mark, please approve if you want to be part of lead group :).

r? @nikomatsakis

PR #173, opened 2019-10-25
Add email for mark-i-m

mark-i-m:
r? @spastorino


spastorino:
I think we would rather close this one and I can add the email to #172

PR #180, opened 2019-11-08
Add vertexclique to wg-localization

vertexclique:
Add `vertexclique` to wg-localization.


Manishearth:
Team should be renamed to internationalization, but yeah

PR #186, opened 2019-11-27
SimonSapin: use my other email address

SimonSapin:
I’m curious what lead to this one being used here.


pietroalbini:
It was the one on the old mailing lists.

PR #193, opened 2019-12-09
add ffi-unwind project group

nikomatsakis:
For now, we have to use the wg- prefix because team repo has to be
taught about project groups.

cc @BatmanAoD @acfoltzer

PR #194, opened 2019-12-10
Add safe-transmute project group

rylev:
Adding new project group in similar way as #193.

cc @nikomatsakis @joshtriplett


teams/wg-safe-transmute.md
```diff
@@ -0,0 +1,18 @@
+name = "safe-transmute"
```
nikomatsakis: Huh, I thought that these had to be named wg-foo ...

pietroalbini: It's supposed to...

rylev: @pietroalbini @nikomatsakis whoops that's my fault. Strange that the CI passes....Should I change it?

nikomatsakis: Yes, might as well. It is curious that CI passes though, I got errors when I ran `cargo run -- check` locally

XAMPPRocky:
Thank you for your PR! One slight problem with this PR is that the file is using the `.md` extension instead of `.toml`. 😄

PR #196, opened 2019-12-11
Add inline-asm project group

Amanieu:
Adding new project group in similar way as #193 and #194.

cc @nikomatsakis @joshtriplett


XAMPPRocky:
@Amanieu Thank you for your PR! One slight problem with this PR is that the file is using the `.md` extension instead of `.toml`. 😄

nikomatsakis:
current status: I'm waiting here for https://github.com/rust-lang/rfcs/pull/2836 to merge

PR #200, opened 2019-12-16
Remove @Kimundi from rfcbot polls

SimonSapin:
Per email with them.


SimonSapin:
r? @alexcrichton
CC @rust-lang/libs

PR #211, opened 2020-01-15
Wg governance update

nikomatsakis:
This PR 

* Adds @valgrimm to the governance wg, as they've been quite active lately 🎉 
* Replaces @skade as lead with @XAMPPRocky and myself, as @skade has been busy of late and we've been kind of carrying that ball


skade:
Thanks for this PR.

PR #217, opened 2020-01-21
Archive some teams and add alumni

XAMPPRocky:
Archives some of the teams that no longer active.

r? @nikomatsakis


pietroalbini:
Are we worried about the `ecosystem@` and `production@` mailing lists disappear?

Manishearth:
I don't think they were ever in use? But perhaps we should have a bounces list just in case.

XAMPPRocky:
For future proofing i think we'd want to have a bounces list regardless. As even if they are unused, eventually we'll have one that has been used, that we want to archive.

Manishearth:
Yes, I'm proposing a general bounces list, not one specifically for these :)

nikomatsakis:
Seems good to me. I've got a branch trying to modernize the 'networking working group' setup to reflect the people who are most active now, I've just been debating how precisely to set it up.

nikomatsakis:
I don't quite understand what you all mean by a "bounces" list -- what is that and what should we do to set it up?

nikomatsakis:
Does that block me from merging this PR? :)

XAMPPRocky:
@nikomatsakis An email bounce list, so that any emails sent to the mailing list addresses after the team is removed are sent *somewhere* rather being lost to the ether. I don't what it would take to set that up. I don't know if these email addresses are worth that, as I don't know how active they were. I don't think it should be a blocker for merging.

nikomatsakis:
OK, I'm going to just go ahead and merge this. Based on earlier comments it doesn't sound like anyone objects.

PR #220, opened 2020-01-24
Move Kimundi to alumni status

alexcrichton:
Kimundi reached out over email and mentioned they'd like to be moved to
alumni status, so this is doing so.


Mark-Simulacrum:
r=me

PR #222, opened 2020-01-31
Add zulip ID field to people

Mark-Simulacrum:
This will be used by triagebot for authentication; we also provide a mapping
from Zulip ID to the GitHub user ID which is the canonical "identifer" of a
person in the Rust organization.

r? @pietroalbini


pietroalbini:
This looks good! How can someone discover their ID?

Mark-Simulacrum:
I'm not sure of the best process (the API gives it to you, but it's unclear to me that implies automatic population). Currently the bot will tell you the ID when complaining about authentication failing. I think it can be verified by looking at the auto-generated user12322@ email when mousing over users in Zulip, but we can also just trust users to get it right.

PR #232, opened 2020-02-07
Remove myself from a number of teams

alexcrichton:
Follow-up from
https://internals.rust-lang.org/t/scaling-back-my-involvement-in-rust/11754

### DIFF
--- a/teams/core.toml
+++ b/teams/core.toml
@@ -5,7 +5,6 @@ leads = []
 members = [
     "Mark-Simulacrum",
     "aidanhs",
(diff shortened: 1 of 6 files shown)


pietroalbini:
Thanks for everything you've done Alex! :heart: 

Did you mean to remove yourself from infra-admins as well?

alexcrichton:
I thought for a second it might be good to keep that around in case there's any stray emails and/or things I should help take care of, but on second though y'all can always alert me so I think it's ok!

PR #237, opened 2020-02-07
Add sinato to Cleanup Crew ICE-breakers

sinato:
I like Rust. I have no deep knowledge of the compiler yet, but I will do my best


pietroalbini:
Thanks, and welcome onboard!

PR #239, opened 2020-02-07
Added Noah-Kennedy to people, llvm and cleanup icebreakers.

Noah-Kennedy:
I would like to join the two icebreaker teams. I have not worked much with the compiler or with LLVM, but I am very interested in learning about both.


nikomatsakis:
Added as part of https://github.com/rust-lang/team/pull/244

PR #251, opened 2020-02-11
Sync Core Team Permissions

XAMPPRocky:
While looking through github's teams I noticed that the core team permissions aren't synced between rust-lang/team and GitHub. I'm not sure if this is intentional, but it seems like an oversight.

r? @nikomatsakis

### DIFF
--- a/teams/core.toml
+++ b/teams/core.toml
@@ -18,6 +18,9 @@ members = [
 bors.rust.review = true
 bors.team.review = true
 
+[github]
+orgs = ["rust-lang", "rust-lang-nursery"]
+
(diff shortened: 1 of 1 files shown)


Manishearth:
I _suspect_ we don't want to automate the syncing of a group with admin permissions just so that things can't go horribly wrong. cc @pietroalbini

pietroalbini:
`@rust-lang/core` is not special and doesn't have admin permissions. Organization owners are managed separately.

Manishearth:
Ah, then we should probably land this

PR #253, opened 2020-02-11
Record the Async Foundations WG

nikomatsakis:
So, the "wg-async-await" compiler group has kind of morphed and combined with the "Async Foundations WG" that was announced in [this blog post](https://blog.yoshuawuyts.com/async-ecosystem-wg/). 

To start, I included folks who have been hacking on the compiler -- but in building this, I found (for the first time) [this list of folks](https://rustasync.github.io/team/async-foundations/) who were working on futures 0.3 and the book.  I've not heard from those folks though for all I know they are still active! We should figure out final scope and grow participant list as needed.


teams/wg-async-foundations.toml
```diff
@@ -0,0 +1,19 @@
+name = "wg-async-foundations"
+wg = true
```
nikomatsakis: oops, the whole point of this was to sync with a github team, I should add that. ( I want to use give that team ownership of https://github.com/rust-lang/wg-async-foundations )

PR #261, opened 2020-02-16
add myself to UCG

RalfJung:
Also making myself co-lead. I think this matches @nikomatsakis' intent when he proposed I could be something like the UCG "shepherd/liaison/ambassador" to the lang team, or so.


nikomatsakis:
Yes, definitely, though we should discuss membership and organization here again at some point. But I feel like I'm always saying that.

Issue #146, opened 2019-10-15
Update Rust Team Information


XAMPPRocky:
## Introduction
Hello everyone, The governance working group has been looking to formalise some of the current structure of Rust Programming Language organisation that we have found to be currently unclear and under or undocumented. One example of this is that some of the team's current membership is out of date with [`rust-lang/team`]. So we're asking that team leads (or relevant members) update their roster and to synchronise their team permissions with GitHub. This will reduce some of current overhead in managing our infrastructure.

## How To Update Team Information.
1. Clone the [`rust-lang/team`] repository.
2. Visit your team's information available in the [`teams/` directory][team-dir].
	* Refer to the [TOML schema][team-schema] for how to configure the file.
3. Check that all of the names in `members` matches what you expect. Add any recent/missing team members, and remove any inactive members.
    * When adding members, also ensure that they are also present in the [`people/` directory][people-dir].
	* The command `cargo run add-person <github-username>` inside the team repo will load publicly available information.
	* If the member is not available, please contact them directly, asking them to add their information to [`rust-lang/team`].
4. Make sure your team contains something like the following code snippet. Importantly that you have the `rust-lang` org in your org. This will synchronise your team information here with the team information on GitHub. Your team may already have this and if so no action is required. 
```toml
[github]
orgs = ["rust-lang", "rust-lang-nursery"]
```
6. Run `cargo run check` inside the team repo to ensure the changes you made are valid.

[team-dir]: https://github.com/rust-lang/team/tree/master/teams
[team-schema]: https://github.com/rust-lang/team/blob/master/docs/toml-schema.md#teams
[people-dir]: https://github.com/rust-lang/team/tree/master/people
[`rust-lang`]: https://github.com/rust-lang/
[`rust-lang/team`]: https://github.com/rust-lang/team
[`rust-lang-nursery`]: https://github.com/rust-lang-nursery
[`rust-central-section`]: https://github.com/rust-lang/rust-central-station

## Potential Questions
> How do I decide whether a person should be added or removed from my team?

The Goverance Working Group's purpose is to provide guidance, not decisions. It is up to each team's discretion to decide their roster and what qualifies for membership.

> How important is it to update this?

Having you and your team's information up to date and synchronised with GitHub will help reduce some overhead in adminstrating Rust's infrastructure, and will help in the future lay groundwork for proposals for formalising Rust's access policy.

> How can I see the difference between GitHub's team membership and `rust-lang/team`?

[...]

PR #262, opened 2020-02-16
add Miri team consisting of Oli and me

RalfJung:
Cc @oli-obk 

t-compiler also has Miri r+ rights, and I guess there is no good reason to change that. Still, saying that Miri is a subteam of t-compiler is kind-of arbitrary. Does it matter much?

I was not sure about the `repo` key, as Clippy doesn't have it.

### DIFF
--- a/teams/miri.toml
+++ b/teams/miri.toml
@@ -1,0 +1,14 @@
+name = "miri"
+subteam-of = "compiler"
+
+[people]
+leads = ["RalfJung", "oli-obk"]
+members = ["RalfJung", "oli-obk"]
+
+[permissions]
+bors.miri.review = true
+
+[website]
+name = "Miri"
+description = "design and implementation of the Miri interpreter"
+repo = "https://github.com/rust-lang/miri"
--- a/people/RalfJung.toml
+++ b/people/RalfJung.toml
@@ -1,6 +1,7 @@
 name = "Ralf Jung"
 github = "RalfJung"
 github-id = 330628
+email = "post@ralfj.de"
 
 [permissions]
 perf = true
--- a/people/RalfJung.toml
+++ b/people/RalfJung.toml
@@ -5,4 +5,3 @@ github-id = 330628
 [permissions]
 perf = true
 bors.rust.review = true
-bors.miri.review = true


Mark-Simulacrum:
I don't think it matters much in practice whether it's a compiler subteam (but we also have no other good team to put it under, as it's still I believe primarily intended as a compiler part than a separate tool). Perhaps dev-tools will eventually be a good fit if miri-the-tool becomes more prominent in its development but for now I think compiler makes sense.

Merging, we can always change later if needed.

teams/miri.toml
```diff
@@ -0,0 +1,14 @@
+name = "miri"
+subteam-of = "compiler"
+
+[people]
```
nikomatsakis: do we want to sync this with a github team?

RalfJung: Not sure. Do we?^^ What would that do for us?

XAMPPRocky: @RalfJung If it's synced with a team then you can give that team permissions to repos instead of individuals. So for example if there's a new member, once the PR adding them here is merged they will have the same permissions as every other member. Where as right now you would have to do that manually add them to each repository.

RalfJung: That sounds sensible. But I also have no idea which permissions I have why.^^ I can edit issues both in rust-lang/rust and rust-lang/miri and rust-lang/unsafe-code-guidelines (and maybe more?).

nikomatsakis: Seems fine as is.

nikomatsakis: Well, the main thing is that if there is a distinct mir repository etc, we're trying to move away from granting permissions to individuals and move towards granting permissions to teams. So maybe if there are distinct miri repositories we might want to have a team for that. But it's not a big deal.

RalfJung: I mean if you want to change https://github.com/rust-lang/miri/ to be entirely team-based in terms of permissions, I am fine with that. Let me know if there is anything I can/should do to make that happen.

XAMPPRocky: @RalfJung You can follow these instructions in #146 to add github syncing. It was written for teams but it works the exact same for subteams I believe.

RalfJung: I'm no admin in the Miri repo, so I doubt I can set up any kind of syncing... also I don't think the relevant team even exists already on github so some org admin would probably have to create it?

XAMPPRocky: @RalfJung Yes sorry, the only thing you need to do is make the pull request adding github syncing, once that's merged someone else will update the miri permissions. I believe your team will be created once you've enabled it on GitHub.

RalfJung: I don't understand what I can "enable on GitHub" without having any admin powers.

RalfJung: I opened a PR at https://github.com/rust-lang/team/pull/281. But it's all very confusing, most of https://github.com/rust-lang/team/issues/146 and in particular everything in "Potential Questions" makes little sense in my situation (creating a new team, not adapting an existing one).

PR #127, opened 2019-09-30
Fix team names and descriptions

LEARAX:
The grammar of team names and descriptions on the governance page of rust-lang.org is very inconsistent. This PR attempts to correct this, by changing each description to start with a present progressive verb, applying consistent capitalization rules, and ending most of the team names with either "working group" or "team" (moderator groups and observers were left as exceptions).

PR #267, opened 2020-02-19
Fix inconsistencies that slipped through the cracks

LEARAX:
I don't intend to spam this repository with shallow pull requests, but it looks like a few minor inconsistencies with the working groups were either introduced since I opened #127 or were simply overlooked. I believe I've fixed them now.

### DIFF
--- a/teams/wg-async-foundations.toml
+++ b/teams/wg-async-foundations.toml
@@ -17,6 +17,6 @@ members = [
 orgs = ["rust-lang", "rust-lang-nursery"]
 
 [website]
-name = "Async Foundations working group"
+name = "Async foundations working group"
 description = "Pursuing core language and library support for async-await"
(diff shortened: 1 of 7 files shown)


LEARAX:
Some of my changes here are subjective; I thought that they made it seem more consistent/professional when looking at the site. If it's preferred that I revert them, let me know.

Manishearth:
I don't think the working group names should be lower cased. The "Working Group" perhaps should be _uppercased_

LEARAX:
> I don't think the working group names should be lower cased. The "Working Group" perhaps should be _uppercased_

Should the whole title be like that, then?

Manishearth:
> Should the whole title be like that, then?

Yes

nikomatsakis:
These do look like good changes -- I don't have a strong opinion about the capitalization of Working Group vs working group.

PR #277, opened 2020-02-23
Add QuietMisdreavus as alumni

XAMPPRocky:
This PR should match the current state of the docs.rs and related teams
based on talking with jyn514 and QuietMisdreavus' farewell post.

https://quietmisdreavus.net/self/2020/02/17/rust-ghost-signing-off/

r? @pietroalbini @jyn514

### DIFF
--- a/teams/docs-rs.toml
+++ b/teams/docs-rs.toml
@@ -2,13 +2,12 @@ name = "docs-rs"
 subteam-of = "devtools"
 
 [people]
-leads = ["QuietMisdreavus"]
+leads = ["jyn514", "pietroalbini"]
 members = [
-    "QuietMisdreavus",
     "GuillaumeGomez",
+    "jyn514",
     "onur",
     "pietroalbini",
-    "jyn514",
 ]
 
 [github]
@@ -27,3 +26,6 @@ discord-name = "#docs-rs"
 
 [[lists]]
 address = "docs-rs@rust-lang.org"
+
+[[lists]]
+address = "help@docs.rs"
--- a/teams/alumni.toml
+++ b/teams/alumni.toml
@@ -8,6 +8,7 @@ members = [
     "Kimundi",
     "MajorBreakfast",
     "Nemo157",
+    "QuietMisdreavus",
     "TimNN",
     "alercah",
     "arielb1",
--- a/teams/docs.toml
+++ b/teams/docs.toml
@@ -5,7 +5,6 @@ leads = ["steveklabnik"]
 members = [
     "steveklabnik",
     "GuillaumeGomez",
-    "QuietMisdreavus",
     "rylev",
     "celaus",
 ]
@@ -30,11 +29,3 @@ extra-teams = ["reference"]
 [[lists]]
 address = "doc-team@rust-lang.org"
 extra-teams = ["reference", "rust-by-example"]
-
-[[lists]]
-address = "help@docs.rs"
-include-team-members = false
-extra-people = [
-    "QuietMisdreavus",
-    "onur",
-]
--- a/teams/rustdoc.toml
+++ b/teams/rustdoc.toml
@@ -5,7 +5,6 @@ subteam-of = "devtools"
 leads = ["GuillaumeGomez"]
 members = [
     "GuillaumeGomez",
-    "QuietMisdreavus",
     "ollie27",
     "kinnison",
 ]


teams/docs-rs.toml
```diff
@@ -27,3 +25,6 @@ discord-name = "#docs-rs"
 
 [[lists]]
 address = "docs-rs@rust-lang.org"
+
+[[lists]]
+address = "help@docs.rs"
```
jyn514: What does this email address do?

XAMPPRocky: It was an email that was in the `docs` team that only included misdreavus and onur. I'm just adding it here to redirect any email that could be sent to that address can be sent to the rest of the team instead.

jyn514: Will emails sent to that address get forwarded to `docs-rs@rust-lang.org`? Since misdreavus and onur left it looks like no one has access to that address anymore.

XAMPPRocky: I believe it will just create another mailing list that's consists of the the same members as `docs-rs@rust-lang.org`. So this will essentially be an alias for that list.

pietroalbini: Yes, and TIL we had such mailing list.

teams/docs-rs.toml
```diff
@@ -2,13 +2,11 @@ name = "docs-rs"
 subteam-of = "devtools"
 
 [people]
-leads = ["QuietMisdreavus"]
+leads = ["jyn514", "pietroalbini"]
 members = [
-    "QuietMisdreavus",
     "GuillaumeGomez",
-    "onur",
```
pietroalbini: Let's keep onur on the team until we get a chance to talk with them and discuss whether they want to still be on the team.

XAMPPRocky: I would prefer to do the opposite. Remove them now, reach out to them, and if they want to still be on the team add them back at that point. Looking at their GitHub activity they haven't been active for awhile. A person being moved off a team into alumni doesn't mean that person isn't welcome to be active and be on the team again, this just puts the onus and work on them to add themselves back, rather than on us to keep track of them.

If this works I can send the email.

nrc: I don't think this is the right thing to do. Although long term, we want our teams to match reality, there is very little cost in keeping non-active people on teams in the short-term. People often have commitments outside of Rust and we should not judge people's involvement by their GitHub activity. It is common courtesy to talk to people before adding or removing from teams. I worry that this is the open-source equivalent of finding out that you've been fired because your corporate login stops working :-)

"A person being moved off a team into alumni doesn't mean that person isn't welcome to be active" - I know you have good intentions here, but I think that for any person in this position, it will feel extremely discouraging and demotivating.

"puts the onus and work on them to add themselves back" - I think the Rust community should strive to be as encouraging and kind as possible in this kind of situation.

"... us to keep track of them" - on the other hand, I consider this part of the leadership role for any team leader and the core team.

nrc: (Given Misdreavous's post I think it is fine to move her)

XAMPPRocky: > I don't think this is the right thing to do. Although long term, we want our teams to match reality, there is very little cost in keeping non-active people on teams in the short-term. People often have commitments outside of Rust and we should not judge people's involvement by their GitHub activity. It is common courtesy to talk to people before adding or removing from teams. I worry that this is the open-source equivalent of finding out that you've been fired because your corporate login stops working :-)

I mentioned the github activity thing off handly which was a mistake. I also talked to @jyn514 who said that they hadn't been involved since they had joined.

I'm happy to block removing the person if we were to set some reasonable limits both on how long we should wait to reach out, and then a limit on how long they have to respond. This repo can often be out of date by several months, I'm not here removing people who took a month or two off.

I want to push back on this being something we want *"long term"* though, when organising the all hands one of the problems we had was making sure that the right people were being contacted about the all hands, and this is managed entirely through the team repo. Having a roster that is not accurate is something that *has* already affected us.

> "... us to keep track of them" - on the other hand, I consider this part of the leadership role for any team leader and the core team.

I don't know if that's ever been communicated as something leaders should do, or worked out in practice. The Governance WG created a [tracking issue][ti] five months ago to update teams, and it some teams still haven't updated their rosters (one of which is core funnily enough).

[ti]: https://github.com/rust-lang/team/issues/146

> "A person being moved off a team into alumni doesn't mean that person isn't welcome to be active" - I know you have good intentions here, but I think that for any person in this position, it will feel extremely discouraging and demotivating.

I think I mostly have different perspective. Being on Rust team is a great privilege and it is motivating to be on the recognised as part of the team. Being on a Rust team is also an *obligation*. Being in this repo is signing yourself for a lot of noise in your email and on GitHub, I don't want people to feel like they have to reply or fill their inbox just because they were contributing for some period in the past.

I think your comparison to corporations is interesting, as I believe we too often think of open source as a corporate ladder. That people only as contribute more and stick around they get "promoted" to new positions and repeat. When really what I believe happens in reality is we have a more of a stewarding model, where people come, contribute, get responsibility, reach their peak, hand it off to other people, and move on.

A lot of other open source projects have limits, and I believe Rust does too, only that they're currently from invisible factors like burn-out. I also believe a lot of people who might feel burnt out or busy with their this way are reluctant to make a PR and remove themselves from team since can be seen as admitting defeat that they aren't living up the obligations, versus "We haven't heard from you in awhile, we can add you back at any time."

This is anecdotal but I have seen people file issues or try to reach out to inactive members because they are still listed on the website as part of the team. People shouldn't be doing that, and we shouldn't be creating un-needed noise for people who want to scale back their involvement.

<hr>

I would also be open to someway of reflecting activity in the team toml structure so that members can still have their permissions and still be shown as part of the team, so that it doesn't feel like you have your login revoked.

I think overall we could be doing a lot better work on our alumni section and archiving and recognising people for their work in the Rust lang organisation. It's important that we have a current up to date membership but also that we don't forget about the people who worked and stopped who don't have the "celebrity" status that some ex-members have.

Sorry for the long reply. A lot of this is out of the scope of this single PR, but your response got me thinking :)

XAMPPRocky: Another thing that's not really important in this case, but is more important for other teams is that inactive members slow down RFCs as one often the biggest blockers on PRs is waiting for enough people to have ticked their box. I don't know rfcbot handles if team members leave their team in any way. Of course I'm not arguing that RFCs should be rushed but when they stall out it's often very unclear how to move the RFC forward.

nrc: Thanks for the reply. I think this is an area where we should rely on the team leads to make decisions. We don't (and probably can't) keep track of people's involvement, and this is part of the team leads role, to ensure that team members are present, motivated, and doing well. I think that if the team lead wants to remove a person that is fine. I would expect the team lead to keep this roster up to date within reason (with nudging from core/governance if needed).

I'm not sure if it is useful to give a hard time limit, since people tend to drift away rather than have a set date where they disappear.

> as I believe we too often think of open source as a corporate ladder. That ...

This is really interesting, thanks!

> I don't know if that's ever been communicated as something leaders should do, or worked out in practice

Sorry, I meant that it is up to team leads to keep track of who is still actively involved and who is not, and for them to decide when a person is not on a team. I do think they should also keep this roster up to date, but that is secondary to keeping track of the people themselves. I don't think that is written down but I hope it is something that leaders learn from mentoring or experience.

> I think overall we could be doing a lot better work on our alumni section and archiving and recognising people for their work in the Rust lang organisation

Strong agree. The alumni section was created a long time ago when there was only one alumnus and only a few teams. It could be much improved for the current state of affairs.

> inactive members slow down RFCs

Team leaders can tick boxes for members they know are no longer involved. RFCBot is a helper, not an authority so I hope this does not cause issues.

Moving forward with this, it seems that @pietroalbini @jyn514 are the new docs-rs team leaders, so we can leave it up to them to contact Onur, and if they think it is necessary. They should probably have a discussion offline and then send a PR if/when they are ready.

In the meantime, we can merge Misdreavous's part of this PR.

Thanks for being proactive in keeping things up to date!

XAMPPRocky: > I'm not sure if it is useful to give a hard time limit, since people tend to drift away rather than have a set date where they disappear.

Sorry I kind of mixed my points here. The main point being a limit on reasonable contact. For example if you ghosted tomorrow and were unresponsive for 6 months, we'd attempt to contact you, and if you haven't responded in another 3 months we send a final reach out, and if there hasn't been anything 3 months after that we'd remove you as inactive. In this particular example you'd have to inactive and unresponsive for a full year to be removed, but it could be longer or shorter.

>  I think that if the team lead wants to remove a person that is fine. I would expect the team lead to keep this roster up to date within reason (with nudging from core/governance if needed).

I believe whatever *"within reason"* is, the current state of team is far outside of that, this is not the only case where I have talked to leads who said they have other members or leads who have never participated in their tenure, and they don't remove them because no one has told anything.

I think this if this is the expectation it should be actually documented somewhere as something leads should do and clearly communicated, as team leads aren't currently pulling their administrative load. I would prefer to *not* spend most my time being proactive and nudging every team lead to have an up to date roster, or reach out to every inactive member to see if they want to stay on.

XAMPPRocky:
I've updated the PR to not remove onur based on the conversation above.

teams/docs-rs.toml
nrc: > I think this if this is the expectation it should be actually documented somewhere as something leads should do and clearly communicated, as team leads aren't currently pulling their administrative load.

This is a good point and I agree. Following up with the core team...

PR #281, opened 2020-02-27
teams/miri: add github clause

RalfJung:
I *think* this is what @XAMPPRocky [asked me to do](https://github.com/rust-lang/team/pull/262#discussion_r384749112), but it is not clear at all to me. https://github.com/rust-lang/team/issues/146 lists a dozen things to do and most of them make no sense as the team doesn't exist yet and I have no admin powers.

### DIFF
--- a/teams/miri.toml
+++ b/teams/miri.toml
@@ -12,3 +12,6 @@ bors.miri.review = true
 name = "Miri"
 description = "design and implementation of the Miri interpreter"
 repo = "https://github.com/rust-lang/miri"
+
+[github]
+orgs = ["rust-lang"]


pietroalbini:
Let me try to clarify things a bit. Our goal is to have the permissions on the miri repository be synchronized with the team repo, so when people are added or removed here the permissions are automatically updated. To do that, two things are needed:

1. A GitHub team synchronized with the team repo needs to be created.
2. The permissions on the miri repository needs to be changed to allow this new GitHub team to interact with it.

This PR does the first point: once we enable GitHub synchronization like you did with this, a bot will automatically create a GitHub team named `@rust-lang/miri` on its own and add y'all to it. Then, for the second step, someone with admin access will need to change the permissions on the miri repository.

pietroalbini:
Merged, the team should be created in a few minutes. I'll tweak the permissions on the miri repository later today.

RalfJung:
Awesome. :)

If you find that there were some individuals besides @oli-obk and me having access (that are not in the team), let me know. I suppose we might as well use this opportunity to make sure that the miri repo uses teams exclusively for permissions.

pietroalbini:
Ok, switched the miri repository to use teams! Also pinged you on Zulip for the other people who have access.

PR #287, opened 2020-03-04
Add support for declaring team alumni

pietroalbini:
This PR adds support for declaring team alumni, which will be properly shown on the website. Team alumni will be automatically included in the `alumni` team. Team alumni will be included in the API.

r? @XAMPPRocky

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -42,6 +42,11 @@ members = [
     "craterbot",
     "rust-timer",
 ]
+# Past members of the team. They will not be considered as part of the team,
+# but they will be recognized on the website.
+alumni = [
+    "buildbot",
+]
 
 [permissions]
 # Optional, see the permissions documentation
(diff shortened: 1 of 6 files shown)


XAMPPRocky:
Overall this looks good. Though I'm not too familiar with the team crate's structure to feel super confident. The one pice of feedback I have is that these new fields  should be documented. Even `include_all_alumni` even if just to say to not use it.

pietroalbini:
Documented the `alumni` key. I'd prefer not to document the special one for the alumni team though, in the past we never documented those ad-hoc things.

XAMPPRocky:
@pietroalbini Well I mean even documenting it as a comment would be fine, it doesn't have to be in the public api. It should mention it's tended to be temporary. We don't know how long it will take to completely move away from needing it or who will be looking at this in the future.

pietroalbini:
Ok, documented everything.

PR #288, opened 2020-03-05
Remove skade from governance group

nikomatsakis:
@skade has asked to step back from the governance working group. Thanks for your efforts, @skade!


XAMPPRocky:
@nikomatsakis Maybe we should add them to `alumni` instead of removing them? We have it implemented already on team and an open PR on the website. https://github.com/rust-lang/www.rust-lang.org/pull/1025

nikomatsakis:
Ah, I didn't know that was a thing already.

PR #298, opened 2020-03-13
Archive Docs and Docs subteams

XAMPPRocky:
This PR archives the docs team and it's sub teams, adding any members not on other teams to alumni.

As with the previous archival it should be noted that this will mean that mailing lists associated with these teams will stop working, as we do not have a bounce list set up. Someone on the team will have to weigh in on whether that should be a blocker.

r? @steveklabnik


steveklabnik:
So, the reference and RBE teams should really be converted to wgs. The email address thing shouldn't be a big deal, they're never used.

XAMPPRocky:
@steveklabnik Ok, should they have a parent team?

steveklabnik:
It is not clear to me, to be honest. Do we always require wgs to have a parent team?

pietroalbini:
If they don't have a parent team they'll show up as top-level teams in the website.

Manishearth:
@steveklabnik we have the security response WG as toplevel already. I think it's fine. We need to fix up the domain vs normal WG distinction, but as previously discussed we're reevaluating domain WGs anyway (https://github.com/rust-lang/wg-governance/issues/46)

XAMPPRocky:
Well the reference team has been migrated to lang, and I think we should figure out the current state of "Rust By Example" group, and if this team is still active, before moving forward with the PR.

XAMPPRocky:
cc @marioidival 

Just looking at rust-by-example repo it seems you're still actively maintaining the project.

marioidival:
What is the real plan for RBE, will there be no more maintenance?

steveklabnik:
No, it still needs a ton of work! It should just be a WG.

camelid:
@steveklabnik What team would an RBE WG fall under though? It wouldn't be the docs team, since the whole point is that it no longer exists :)

steveklabnik:
Historically, I don't think there's been a parent team requirement. Many WGs don't have them.

camelid:
> If they don't have a parent team they'll show up as top-level teams in the website.

^ is this an issue though

marioidival:
I'm fired from Docs/RBE? :sweat_smile:

camelid:
I think we're going to create an RBE working group.

PR #309, opened 2020-04-01
add Aaron1011, Jonas to the compiler contributors

nikomatsakis:
Thanks for all your hard work, @aaron1011 and @jonas-schievink! <3


nikomatsakis:
cc @rust-lang/compiler

PR #318, opened 2020-04-14
Move fmoko to alumni

fmoko:
I'm retiring from the community team, this change formalizes that.

The future of the RustBridge team (which I lead) is currently unclear, so I'm not touching it yet, that'll go in a separate PR.

cc @rust-lang/community

### DIFF
--- a/teams/alumni.toml
+++ b/teams/alumni.toml
@@ -21,6 +21,7 @@ members = [
     "dwijnand",
     "edunham",
     "erickt",
+    "fmoko",
     "frewsxcv",
     "huonw",
     "jonathandturner",
--- a/teams/alumni.toml
+++ b/teams/alumni.toml
@@ -21,7 +21,6 @@ members = [
     "dwijnand",
     "edunham",
     "erickt",
-    "fmoko",
     "frewsxcv",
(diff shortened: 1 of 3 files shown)


JohnTitor:
Could you add the `alumni` entry then add you there, like compiler team? I think it's common style to move someone to alumni nowadays:

https://github.com/rust-lang/team/blob/45013d114d89480c33cab2a1a99828e4aacb8aee/teams/compiler.toml#L17-L21

fmoko:
@JohnTitor yep, updated

teams/alumni.toml
```diff
@@ -21,6 +21,7 @@ members = [
     "dwijnand",
     "edunham",
     "erickt",
+    "fmoko",
```
pietroalbini, line 24: There is no need to add you explicitly to the alumni team, being in `alumni = ["..."]` inside a team automatically puts you here as well.

JohnTitor:
Updates from the comm-team meeting: we moved the items about rustbridge including mailing lists to next month.

badboy:
ping @pietroalbini - this can be merged now

PR #324, opened 2020-04-20
Return dump-team and dump-list in sorted order

dtolnay:
Rather than random HashSet order. This matches the sorting done in dump-permission and makes it somewhat easier to visually scan and diff.


Mark-Simulacrum:
Thanks!

PR #329, opened 2020-04-28
Add @taiki-e to wg-async-foundations

tmandry:
@taiki-e has been maintaining pin-utils, and is also the author of the pin-project crate.


Mark-Simulacrum:
Looks like you'll need to create @taiki-e as a person, see https://github.com/rust-lang/team#adding-a-person-to-the-repository

PR #330, opened 2020-04-30
crates.io: Add @pichfl as team member

Turbo87:
/cc @pichfl


pichfl:
I'm honored to be a part of this.

PR #342, opened 2020-05-26
removed jakevossen5 from clenup crew

jakevossen5:
I want to continue contributing, but the emails from this have been a bit too much but I don't want to disable GitHub emails. Thanks!


Mark-Simulacrum:
Thanks for your help!

spastorino:
Idea: what if we add a group of the people that are occasionally pinged?. Like we could have the main group that's always pinged and we could have a ping of people that is pinged like once every four times or something like that.

Trying to find a solution because it's sad that have people that want to help signing off :).
Would something like this or similar help you @jakevossen5 ?.

jakevossen5:
I agree that it is sad to sign off. 

I like the idea - but I don't think that is quite the right solution. The problem wasn't getting pinged all the time, it is mostly the fact that I have the GitHub setting turned on that sends me an email every time someone comments on an issue I am mentioned on. I like this setting and want to keep it on, but as you can imagine, those threads get quite long. I probably could have been less lazy and unsubscribed from the threads that I was not interested in / wasn't going to work on, but it was always easier to just archive than to unsubscribe so the emails kept flowing... 

This would be a more complicated solution, but if instead of commenting on the issue at hand when you ping cleanup crew - it creates a new issue / adds another message on an existing issue (aka cleanup-crew tracker). That way you would be notified and if you want to work on that issue you would go to the issue and hit subscribe, otherwise you only got that one email. Not sure how elegant that solution is though.

spastorino:
Hmm yeah, this maybe gets a bit more complicated than what we want. At some point we could also have a twitter account sending tweets about new issues and things like that. I guess for now what one can do is follow [this link](https://github.com/rust-lang/rust/labels/ICEBreaker-Cleanup-Crew) from time to time and check what's in there meanwhile we see if there's some solution for this problem. Thanks and hope you're able to follow in some way.

PR #356, opened 2020-06-09
Update wg-traits team

jackh726:


jackh726:
So not sure why CI is failing here. wg-traits doesn't have any mailing lists

nikomatsakis:
That does seem curious. Maybe there is a mailing list of all WG devs? But then skinny121 was already a member... maybe @pietroalbini has an idea what's going on.

jackh726:
My thought was maybe there's something special for alumni? But I couldn't find anything.

pietroalbini:
Apparently they're added to the `all@` mailing list...

pietroalbini:
Can we add the new members in this PR and move the past members to alumni in a separate PR? I don't want to block adding new members for too long.

jackh726:
I updated to only add new people

PR #362, opened 2020-06-15
Add RISC-V group

tblah:
cc @denisvasilik @Disasm

On https://github.com/rust-lang/rust/pull/72973 @nikomatsakis  suggested creating a RISC-V group (similar to the ARM and Windows groups). 

r? @nikomatsakis

### DIFF
--- a/teams/risc-v.toml
+++ b/teams/risc-v.toml
@@ -1,0 +1,10 @@
+name = "riscv"
+marker-team = true
+
+[people]
+leads = []
+members = [
+    "denisvasilik",
+    "Disasm",
+    "tblah",
+]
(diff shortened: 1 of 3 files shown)


nikomatsakis:
Thanks for this. I think we're still waiting a bit to figure out the infra team side of things though.

tblah:
Can this be merged now that https://github.com/rust-lang/rust/pull/72973 is merged?

teams/risc-v.toml
```diff
@@ -0,0 +1,10 @@
+name = "riscv"
```
pietroalbini, line 1:
```suggestion
name = "risc-v"
```

The name here should be the same as the file name :)

PR #387, opened 2020-07-21
rust-timer should be able to invoke itself

Mark-Simulacrum:


pietroalbini:
Let's be sure to have safeguards to avoid spam :)

Mark-Simulacrum:
Yeah, I plan to basically force that issue by not posting comments which invoke themselves (i.e., creating an acyclic graph of events so to speak).

PR #401, opened 2020-08-11
Added hargoniX to the WG and eldruin and hargoniX to the resources team

therealprof:
Signed-off-by: Daniel Egger <daniel@eggers-club.de>

### DIFF
--- a/people/hargoniX.toml
+++ b/people/hargoniX.toml
@@ -1,0 +1,4 @@
+name = 'Henrik Böving'
+github = 'hargoniX'
+github-id = 33270164
+email = false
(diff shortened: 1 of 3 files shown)


pietroalbini:
Thanks!

RalfJung:
This PR broke rustc CI (https://github.com/rust-lang/rust/pull/75450). Is there some CI (or so) we could do in this repo to prevent that from happening?

pietroalbini:
I don't think so. A proper fix would be to change toolstate on the rustc side to only allow assigning teams to tools, so we would not have problems with individuals leaving teams.

PR #402, opened 2020-08-11
Synchronize wg-llvm with GitHub

Mark-Simulacrum:
cc @cuviper @nikic -- I've added this which should make @rust-lang/wg-llvm actually pingable (something I've at least personally wanted for a long time now).

cc @nikomatsakis @pnkfelix as t-compiler leads

r? @pietroalbini to make sure I did it right

### DIFF
--- a/teams/wg-llvm.toml
+++ b/teams/wg-llvm.toml
@@ -6,6 +6,9 @@ wg = true
 leads = ["nagisa"]
 members = ["nagisa"]
 
+[[github]]
+orgs = ["rust-lang"]
+
 [website]
 name = "LLVM working group"
 description = "Working with LLVM upstream to represent Rust in its development"


teams/wg-llvm.toml
```diff
@@ -6,6 +6,9 @@ wg = true
 leads = ["nagisa"]
 members = ["nagisa"]
```
cuviper, line 7: Did you intend to extend the members too?

Mark-Simulacrum: Oh I completely didn't look at the members list, I definitely think we should do that! I think adding @nikic and @cuviper is right? Are there others who should be included?

pietroalbini:
Looks good.

teams/wg-llvm.toml
cuviper: Three of us make a good start, better than putting it all on @nagisa. :slightly_smiling_face:

PR #413, opened 2020-08-19
Add zulip-stream data to teams

Alexendoo:
Should be most of them, but I may have missed a couple

### DIFF
--- a/teams/cargo.toml
+++ b/teams/cargo.toml
@@ -25,8 +25,7 @@ ping = "rust-lang/cargo"
 [website]
 name = "Cargo team"
 description = "Designing and implementing the official Rust package manager, Cargo"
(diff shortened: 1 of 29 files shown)


pietroalbini:
Thanks for doing this :heart: 

A thing I'd prefer to avoid is having `.2F` in the names instead of `/`, as that's just an implementation detail of the website. If you send a PR [to change the website to replace `/` with `.2F` in the Zulip stream name](https://github.com/rust-lang/www.rust-lang.org/blob/67c9ef963e78481fa978f502fad3dd4a1973a264/src/teams.rs#L99-L104) I'll be happy to merge both this and the website one :)

pietroalbini (approved):

pietroalbini:
Thanks! Let's merge this once the website is deployed.

PR #418, opened 2020-08-27
Add a test suite to the team repository

pietroalbini:
Up until now there was no test suite for the team repository, and that started to bite us recently (for example with the inclusion of working group leads in the leads team).

This PR adds a test suite to the team repository similar to compiletest: inside `tests/static-api` there is another "team repository" with some dummy data, and the test suite ensures its output of the static API matches the expected one committed to the repository. If the output doesn't match a diff is displayed, and the `tests/bless.sh` script can be used to bless the new output.

The test suite doesn't fully cover everything yet, but it's a start: I'll followup with other PRs in the near future.


tests/snapshot.rs
```diff
@@ -0,0 +1,117 @@
...
+        // Redirects stderr to stdout to be able to print the output in the correct order.
+        let res = self.stderr_to_stdout().stdout_capture().unchecked().run()?;
+        print!("{}", String::from_utf8_lossy(&res.stdout));
+
+        if !res.status.success() {
+            failure::bail!("command returned a non-zero exit code!");
+        }
+        Ok(())
+    }
+}
```
Mark-Simulacrum, line 117: It seems a bit unfortunate to hand-roll this, is there a reason to do that over using some of the existing tooling? e.g., rust-analyzer and rustc are currently using https://docs.rs/expect-test/0.1.0/expect_test/, and @matklad has commented that there are a number of others (e.g., https://docs.rs/insta/0.16.1/insta/).

I would guess none of those would 100% fit what we have here though, mostly interested in whether you've looked at them or if the impl in this PR just seemed simpler.

Mark-Simulacrum:
I think I would prefer to use dummy (e.g., bot -- including existing) accounts for the testing. While I don't plan to leave anytime soon, I do semi-frequently run things like `rg <username>` in the team repo to find out what teams a user is on, and having users belong to test-only teams seems annoying for that.

tests/snapshot.rs
matklad: Yeah, if we want to literally compare files on disk, than hand-rolling is the best option I know.

If we, however, go for a less integrated test of a thing that returns something like HashMap<Path, String>, we should be able to feed it to some existing snapshot testing library.

Although, given the number of files here, and the added reassurance of the integrated test, hand rolling might be the best option indeed.

tests/snapshot.rs
```diff
@@ -0,0 +1,117 @@
...
+            files.insert(entry.path().strip_prefix(dir)?.to_path_buf());
+        }
+    }
+
+    // Check whether any file is different
+    let mut failed = false;
+    for file in &files {
+        let expected = std::fs::read_to_string(dir_expected.join(file))
+            .ok()
+            .unwrap_or_else(String::new);
```
matklad, line 44: unwrap_or_default might work?

tests/snapshot.rs
pietroalbini: Yeah, I'd prefer to test the actual output of `cargo run static-api`.

tests/snapshot.rs
pietroalbini: Done.

pietroalbini:
@Mark-Simulacrum good point!

I changed all the users in tests with dummy users, and added a `--skip` flag to `cargo run check` to skip individual validation checks (since we need to disable the `validate_github_usernames` check).

teams/leaderless.toml
```diff
@@ -0,0 +1,5 @@
+name = "leaderless"
+
+[people]
+leads = []
+members = ["pietroalbini"]
```
Mark-Simulacrum, line 5: Is this intentional or an artifact? It seems to not be used anywhere else...

Mark-Simulacrum:
This looks good to me with CI fixed.

teams/leaderless.toml
pietroalbini: Woops, it was supposed to be inside the test suite.

PR #419, opened 2020-08-28
Add parselib wg

matklad:


Mark-Simulacrum:
It would be great to have some more members here, but I think this is a good initial start at least.

@pnkfelix seemed to feel we *had* already formed in this a recent conversation, so I'm going to go ahead and just merge this.

PR #422, opened 2020-08-31
Add the inside-rust-reviewers marker team

pietroalbini:
This PR adds the `inside-rust-reviewers` marker team, which is composed by all team leads and project group leads. This is part of the work needed to address https://github.com/rust-lang/core-team/issues/18.

I chose to create a dedicated "reviewers" marker team instead of creating a `project-group-leads` team because that allows adding extra people as reviewers without having to change the CODEOWNERS file on the blog.


Mark-Simulacrum:
r=me when ready.

PR #435, opened 2020-09-22
Adjust compiler performance group

Mark-Simulacrum:
This is just a sketch of who I know for sure is actively working in this area -- happy to expand this set. I was sparked to do this by wanting to add @wesleywiser with commit/write bit to rust-lang/rustc-perf.

r? @nikomatsakis 
cc @wesleywiser


wesleywiser (approved):

nikomatsakis:
Seems good as a start :) but we need to revisit the compiler-team groups in general

PR #444, opened 2020-09-30
Add myself to libc bors permissions

joshtriplett:


joshtriplett:
cc @alexcrichton

alexcrichton (approved):

JohnTitor:
Hm, we currently don't use bors on libc as Alex suggested, do we use it again?

joshtriplett:
Fair enough. I'd like to go ahead and get added in case that changes, but I'm not looking to propose changing that. :)

Mark-Simulacrum:
I am personally against expanding the set of permissions like this for a hypothetical future - that said, it would be great to create a libc team/group (at least marker) rather than managing these permissions per-person.

joshtriplett:
Happy to see this happen via a group instead.

PR #455, opened 2020-10-15
Remove spastorino from cleanup crew

spastorino:
I've added myself to this group at the very beginning as a way to test things out and never removed myself, so here I am :).


skade:
Sooo... you forget to clean up?

spastorino:
Rebased cc @pietroalbini

PR #465, opened 2020-10-28
Move sgrif to infra alumni

pietroalbini:
@sgrif did a lot of great work on infrastructure in the past, especially around crates.io, but she doesn't have the time to dedicate to the infra team anymore. Thanks Siân, you'll be missed!

cc @rust-lang/infra


sgrif:
For the record my pronouns are they/them

PR #467, opened 2020-10-30
Remove the operations team

pietroalbini:
The operations team was supposed to be an umbrella for the infrastructure and release teams, but it never had any actual use, meeting or role. This removes the team, putting both infra and release as top-level teams on the website.

r? @Mark-Simulacrum

### DIFF
--- a/teams/infra.toml
+++ b/teams/infra.toml
@@ -1,5 +1,4 @@
 name = "infra"
-subteam-of = "operations"
 
 [people]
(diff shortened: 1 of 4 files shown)


pietroalbini:
[I also pushed a commit to the website repository making sure links will not be fully broken](https://github.com/rust-lang/www.rust-lang.org/commit/6592a146984b21a73eff6318d3f07551c0fef74a).

PR #476, opened 2020-11-09
Add workingjubilee to library reviewers

KodrAus:
So that @workingjubilee can `r+` all the PRs!


KodrAus:
I’ve also updated the Portable SIMD team doc now to list @workingjubilee as a project lead. I meant to do this originally.

PR #477, opened 2020-11-09
Remove myself from the libs team

LukasKalbertodt:
Unfortunately, working as part of the libs team does not currently fit into my life :( 
So I'm already leaving after less than a year. It was a great experience though and I hope to maybe come back some day! And I'm certainly not leaving the Rust community, so you'll see me around. 

Regarding the PRs that I am currently assigned to: I can still review those, but I can't promise how quickly that will happen, so feel free to reassign. (Also, by then I won't be able to merge.) The PRs that I'm the author of I will continue to update, so no change there.

PS: I would not have added myself to the alumni list, but @KodrAus said that I should. If anyone disagrees I can of course change that.

CC @rust-lang/libs

### DIFF
--- a/teams/library-reviewers.toml
+++ b/teams/library-reviewers.toml
@@ -9,7 +9,6 @@ members = [
     "joshtriplett",
     "kennytm",
     "KodrAus",
-    "LukasKalbertodt",
     "m-ou-se",
     "Mark-Simulacrum",
     "sfackler",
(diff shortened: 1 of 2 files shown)


KodrAus:
Thanks for your awesome work @LukasKalbertodt and good luck! ❤️ I hope we will still see you around Rust in the meantime

PR #489, opened 2020-12-03
src: fix include-project-group-leads = true not working

pietroalbini:
Discovered this while looking through the source code.

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -198,7 +198,10 @@ impl Team {
 
     pub(crate) fn members<'a>(&'a self, data: &'a Data) -> Result<HashSet<&'a str>, Error> {
         let mut members: HashSet<_> = self.people.members.iter().map(|s| s.as_str()).collect();
-        if self.people.include_team_leads || self.people.include_wg_leads {
+        if self.people.include_team_leads
+            || self.people.include_wg_leads
+            || self.people.include_project_group_leads
+        {
             for team in data.teams() {
                 let should_include = match team.kind {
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
This will update the inside rust reviewers marker team and wg-leads. I think I would not want wg-leads to include *project* groups - maybe we should drop it from there before merging this.

pietroalbini:
Hmm, why do you think so? The `if` I changed only gates the execution of the `for` loop, but whether to add team leads or not is decided by the `match` below it.

Mark-Simulacrum:
I mean that we include-project-group-leads is explicitly present in those files: https://github.com/rust-lang/team/blob/master/teams/wg-leads.toml#L8 and https://github.com/rust-lang/team/blob/master/teams/inside-rust-reviewers.toml#L8. The latter makes sense to me, the first seems unexpected.

pietroalbini:
Ok, but this PR doesn't change that. The `include-project-group-leads = true` in those files was already working, as there was another `include-*` matched by the if statement. This PR only fixes `include-project-group-leads = true` not working when there is not any other `include-*.

Mark-Simulacrum:
Right, that makes sense, I misread this PR. Ok, this seems fine to me to merge - it does seem like we could also just get rid of this if entirely and that would maybe be better in some sense.

pietroalbini:
Yeah I totally agree it's confusing. I'll refactor the code in a bit.

pietroalbini:
@Mark-Simulacrum tried pushing a refactoring that should hopefully make this more clear.

PR #494, opened 2020-12-10
Give myself perf permissions

jackh726:
For #76814 and other wg-traits-related work


Mark-Simulacrum:
Going to go ahead and merge, but generally we should think about giving wg-traits as a whole these rights perhaps.

PR #498, opened 2020-12-15
Add rylev to compiler-perf team and give compiler-perf team timer permissions

rylev:


Mark-Simulacrum:
We should also grant bors.rust.try?

wesleywiser (approved):

PR #502, opened 2020-12-27
Add v1/discord-map.json endpoint

technetos:
This PR adds a new endpoint `/v1/discord-map.json` to map discord ids to github ids.  

This is part of an effort to automatically manage role assignment in the Rust-Lang discord.


pietroalbini:
Synchronizing team membership with Discord sounds exciting! What's your overral plan for implementing it?

In general, I'd prefer to avoid creating a `discord-map.json` endpoint in the API, and instead I'd represent it inside `Team` (like we do with GitHub teams right now).

technetos:
The plan is to sync once a day and when a user joins the discord.  Additionally a manual `?sync {user}` command would sync the users roles.   

#### Syncing 
If that users discord_id exists in `discord-map.json`, then lookup their github_id in `teams.json` and assign the appropriate roles.  

> In general, I'd prefer to avoid creating a discord-map.json endpoint in the API, and instead I'd represent it inside Team (like we do with GitHub teams right now).

Without `discord-map.json`, how would a user on Discord be associated to a user on GitHub?

pietroalbini:
I would not handle the synchronization in discord-mods-bot. The bot currently has limited privileges (it can assign roles, but it's right above the "talk" role so in practice it can't assign any of the team roles), and expanding the privileges would require the same code review practices as this repository (basically only core would be able to merge PRs to the bot), as otherwise people with write access to the discord-mods-bot repository would be able to escalate their privileges.

I would instead add a new module in [rust-lang/sync-team](https://github.com/rust-lang/sync-team), which is designed to run with elevated privileges. That will also synchronize changes daily *and* as soon as a PR is merged in this repository.

> Without discord-map.json, how would a user on Discord be associated to a user on GitHub?

In the teams schema I'd add something like this:

```json
{
    "discord": {
        "roles": [
            {
                "name": "core-team",
                "color": "#00f",
                "members": [
                    111111111111111111111,
                    222222222222222222222
                 ]
             }
        ]
    }
    ...
}
```

PR #515, opened 2021-01-12
Add optional email encryption support

pietroalbini:
There are cases (especially for the foundation) where we want the members of a mailing list not to be public. This commit implements optional encryption support to temporarily address the problem while the infra team works on a more permanent approach.

Another PR to [rust-lang/sync-team](https://github.com/rust-lang/sync-team) will need to be opened before we can start using encrypted email addresses.

r? @Mark-Simulacrum

### DIFF
--- a/rust_team_data/src/email_encryption.rs
+++ b/rust_team_data/src/email_encryption.rs
@@ -1,0 +1,113 @@
+//! This module implements the encryption scheme used to safely include private email addresses in
+//! the team repository. It generates encrypted content that looks like this:
+//!
+//! ```text
+//! encrypted+3eeedb8887004d9a8266e9df1b82a2d52dcce82c4fa1d277c5f14e261e8155acc8a66344edc972fa58b678dc2bcad2e8f7c201a1eede9c16639fe07df8bac5aa1097b2ad9699a700edb32ef192eaa74bf7af0a@rust-lang.invalid
+//! ```
+//!
+//! The hex-encoded part of the email address is a concatenation of a 24-byte random nonce and the
+//! XChaCha20Poly1305-encrypted email address. Utilities are provided to both encrypt and decrypt.
+
+use chacha20poly1305::aead::{Aead, NewAead};
+use chacha20poly1305::{Key, XChaCha20Poly1305, XNonce};
+
+const PREFIX: &str = "encrypted+";
+const SUFFIX: &str = "@rust-lang.invalid";
+const KEY_LENGTH: usize = 32;
+const NONCE_LENGTH: usize = 24;
+
+/// Encrypt an email address with the provided key.
+pub fn encrypt(key: &str, email: &str) -> Result<String, Error> {
+    // Generate a random nonce every time something is encrypted.
+    let mut nonce = [0u8; NONCE_LENGTH];
+    getrandom::getrandom(&mut nonce).map_err(Error::GetRandom)?;
+    let nonce = XNonce::from_slice(&nonce);
+
+    let mut encrypted = init_cipher(key)?
+        .encrypt(&nonce, email.as_bytes())
+        .map_err(|_| Error::EncryptionFailed)?;
+
+    // Concatenate both the nonce and the payload, as both will be needed for decryption.
+    let mut payload = nonce.to_vec();
+    payload.append(&mut encrypted);
+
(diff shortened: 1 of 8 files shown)


rust_team_data/src/email_encryption.rs
```diff
@@ -0,0 +1,111 @@
...
+    let mut nonce = [0u8; NONCE_LENGTH];
+    getrandom::getrandom(&mut nonce).map_err(Error::GetRandom)?;
+    let nonce = GenericArray::from_slice(&nonce);
+
+    let mut encrypted = init_aes(key)?
+        .encrypt(&nonce, email.as_bytes())
+        .map_err(|_| Error::EncryptionFailed)?;
+
+    // Concatenate both the nonce and the payload, as both will be needed for decryption.
+    let mut payload = nonce.to_vec();
```
Mark-Simulacrum, line 29: If we're converting to a vec, why the GenericArray::from_slice? Can we just initially create the nonce via `vec!`?

rust_team_data/src/email_encryption.rs
```diff
@@ -0,0 +1,111 @@
+//! This module implements the encryption scheme used to safely include private email addresses in
+//! the team repository. It generates encrypted content that looks like this:
+//!
+//! encrypted+a371d1d65dcc697d000826e4bd59dcbfb16ad73eceb1139caaaab795737bb3a54403f0ec69f3d40c8965a5f4bc52f67c94ec538a8a2fea710cabdaaf27661f9a62b0121db7d229@rust-lang.invalid
+//!
+//! The hex-encoded part of the email address is a concatenation of a 12-byte random nonce and the
+//! AES256-GCM-encrypted email address. Utilities are provided to both encrypt and decrypt.
+
+use aes_gcm::aead::{generic_array::GenericArray, Aead, NewAead};
+use aes_gcm::Aes256Gcm;
```
Mark-Simulacrum, line 10: Was there some particular reason to choose this particular crate and cipher?

rust_team_data/src/email_encryption.rs
pietroalbini: `GenericArray` is required by the `aes_gcm` crate's API.

rust_team_data/src/email_encryption.rs
pietroalbini: AES256-GCM is a secure AEAD cipher, and it's definitely good for our use case. There is no particular reason why I chose it compared to other secure AEAD ciphers.

The crate is part of the [RustCrypto project](https://github.com/RustCrypto/) and received [a security audit](https://research.nccgroup.com/2020/02/26/public-report-rustcrypto-aes-gcm-and-chacha20poly1305-implementation-review/). In general I find the maintenance policy of RustCrypto crates to be better for downstream users compared to *ring*.

rust_team_data/src/email_encryption.rs
```diff
@@ -0,0 +1,111 @@
+//! This module implements the encryption scheme used to safely include private email addresses in
+//! the team repository. It generates encrypted content that looks like this:
+//!
+//! encrypted+a371d1d65dcc697d000826e4bd59dcbfb16ad73eceb1139caaaab795737bb3a54403f0ec69f3d40c8965a5f4bc52f67c94ec538a8a2fea710cabdaaf27661f9a62b0121db7d229@rust-lang.invalid
```
camelid, line 4: Maybe you want to put this in a code block so the docs are rendered correctly?

rust_team_data/src/email_encryption.rs
Mark-Simulacrum: We discussed a little on Zulip, and after some consideration it felt like we should probably switch to https://crates.io/crates/chacha20poly1305 - some random googling seems to say that AES256-GCM is not the best thing to use (though likely it doesn't matter for us); libsodium's docs also recommend the use of this primitive.

https://doc.libsodium.org/secret-key_cryptography/aead/chacha20-poly1305/xchacha20-poly1305_construction

pietroalbini: Switched algorithm.

rust_team_data/src/email_encryption.rs
pietroalbini: Done!

pietroalbini:
This is again ready for review.

Mark-Simulacrum (approved):

PR #524, opened 2021-01-23
create the const generics project group

nikomatsakis:
We've had an active const generics project group for some time, but we never created an entry in the team repo for it!

As of opening this PR, the repo doesn't exist, so the link is dangling, but we should create one (based on the [project group template], with this group as write access).

[project group template]: https://github.com/rust-lang/project-group-template/


pietroalbini:
Created the repository!

PR #533, opened 2021-02-04
Remove lenary from RISC-V

lenary:
I no longer work on the RISC-V LLVM target, and I cannot help to triage RISC-V issues any longer.

### DIFF
--- a/people/lenary.toml
+++ b/people/lenary.toml
@@ -1,6 +1,0 @@
-name = 'Sam Elliott'
-email = "sam@lenary.co.uk"
-github = 'lenary'
-github-id = 14548
-discord-id = 646368775599423489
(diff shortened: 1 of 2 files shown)


camelid:
I don't think you need to remove your `people/` file; you may join other teams in the future.

LeSeulArtichaut:
> I don't think you need to remove your `people/` file; you may join other teams in the future.

No, it looks like you can't have that: https://github.com/rust-lang/team/blob/420979ae2ebed613f6c7955e90891a8ea9a82dc9/src/validate.rs#L186-L187

camelid:
Scratch that then. Sorry about that!

lenary:
Yeah, I only force-pushed because I got a build failure.

PR #538, opened 2021-02-17
Sync Discord Roles

technetos:
This PR adds support for syncing team members discord roles with their membership in the teams api.  

In addition, this PR adds discord-role entries to the docs-rs, infra, and community teams.  

A team that
  1. Has a `[discord-role]` in the team config
  2. Has members who have their `discord-id` in the people config

will have their members automatically updated on discord with that corresponding teams role.  

Examples of discord role in team config
```toml
[discord-role]
name = "community-team" // required
role-id = 1234 // required
color = "#000000" // optional
```

Blocks:

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -354,6 +354,7 @@ impl Team {
 pub(crate) struct DiscordRole {
     name: String,
     role_id: usize,
+    color: Option<String>,
 }
 
 impl DiscordRole {
@@ -363,6 +364,10 @@ impl DiscordRole {
 
     pub(crate) fn role_id(&self) -> usize {
         self.role_id
+    }
+
+    pub(crate) fn color(&self) -> Option<&str> {
+        self.color.as_ref().map(|s| &s[..])
     }
 }
 
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -335,6 +335,15 @@ impl Team {
         Ok(result)
     }
 
+    pub(crate) fn discord_ids(&self, data: &Data) -> Result<Vec<usize>, Error> {
+        Ok(self
+            .members(data)?
+            .iter()
+            .flat_map(|name| data.person(name).map(|p| p.discord_id()))
+            .flatten()
+            .collect())
+    }
+
     pub(crate) fn is_alumni_team(&self) -> bool {
         self.people.include_all_alumni
     }
@@ -355,6 +364,11 @@ impl DiscordRole {
     pub(crate) fn role_id(&self) -> usize {
         self.role_id
     }
+}
+
+#[derive(Eq, PartialEq, Debug)]
+pub(crate) struct DiscordTeam {
+    pub(crate) members: Vec<usize>,
 }
 
 #[derive(Eq, PartialEq)]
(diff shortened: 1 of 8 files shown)


Manishearth:
Wow, thanks for working on this!! :clap:

src/schema.rs
```diff
@@ -330,11 +335,47 @@ impl Team {
...
     pub(crate) fn is_alumni_team(&self) -> bool {
         self.people.include_all_alumni
     }
 }
 
+#[derive(serde_derive::Deserialize, Debug)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub(crate) struct DiscordRole {
+    name: String,
+    role_id: usize,
```
pietroalbini, line 356: Can we avoid specifying the Role ID here? Ideally a newly created team would just add something like:

```toml
[[discord-role]]
name = "team-name"
```

...and the new role would automatically be created.

src/schema.rs
```diff
@@ -169,6 +169,7 @@ pub(crate) struct Team {
     website: Option<WebsiteData>,
     #[serde(default)]
     lists: Vec<TeamList>,
+    discord_role: Option<DiscordRole>,
```
pietroalbini, line 172: I would have a `Vec<DiscordRole>` instead of `Option<DiscordRole>` just in case.

pietroalbini (commented):
Thanks for the PR!

In the past we added validation rules (in `src/validation.rs`) to ensure everyone on a team had access to the service when configuring the service for the team. In Discord's case, this would mean that every member of a team with a Discord role would have to have a Discord ID in their profile.

This is the behavior we want for every team but `Teams and WGs`, and we should probably add  a `allow-members-without-discord-id = true` to that `[[discord-role]]` (possibly with a better name).

src/schema.rs
technetos: how would you deal with the role hierarchy?  
image

would a new team go to the bottom?

pietroalbini: A possible solution would be to add new roles after the last role managed by the team repo. So, for example, if you have "managed1", "managed2", "unmanaged1", "managed3", "unmanaged2", new roles would be added after "managed3". if no managed roles are present defaulting to the bottom would be acceptable.

technetos: > Can we avoid specifying the Role ID here? Ideally a newly created team would just add something like:
> 
> ```toml
> [[discord-role]]
> name = "team-name"
> ```
> 
> ...and the new role would automatically be created.

In discord, names of roles are not uniquie.  That is, 2 roles can have the same name and different ids.  Managing roles exclusively by name means we would have to ensure the uniqueness.

technetos: @sebasmagri can you send me or post here your discord-id? Im not 100% sure if your actually on discord and I want to make sure I have the right user.

technetos:
@sebasmagri can you send me or post here your discord-id? Im not 100% sure if your actually on discord and I want to make sure I have the right user.

sebasmagri:
> @sebasmagri can you send me or post here your discord-id? Im not 100% sure if your actually on discord and I want to make sure I have the right user.

It's `sebasmagri#9756` :-)

src/schema.rs
pietroalbini: Right now we practically have uniqueness in the roles. I don't expect we would create multiple roles with the same name.

pietroalbini (approved):

pietroalbini:
Thanks

PR #559, opened 2021-03-26
add zulip-id

nikomatsakis:


Mark-Simulacrum:
(Was going to fix the clippy lint to unblock this PR directly, but it looks like there's more fall out than I had expected; it should happen in a different PR.)

Mark-Simulacrum (dismissed):

Mark-Simulacrum (approved):

PR #594, opened 2021-06-23
Move sfackler to libs-api alumni

joshtriplett:
Based on Zulip discussion with @sfackler. Thanks for all your work on Rust libraries!


joshtriplett:
cc @rust-lang/libs-api

Manishearth (approved):

PR #602, opened 2021-07-06
Move out external email addresses

Mark-Simulacrum:
r? @pietroalbini


teams/core.toml
```diff
@@ -50,38 +50,6 @@ address = "user-logos@rust-lang.org"
 [[lists]]
 address = "legal@crates.io"
```
pietroalbini, line 51: We might want to move this away from core too? Dunno if it's ever been used. Otherwise looks good to me.

Note for onlookers: still used address are not being deleted, but they're now managed outside the team repository.

Mark-Simulacrum: It's not something I'm sure about either but I'd prefer to not touch it just yet.

pietroalbini (approved):

PR #613, opened 2021-08-04
Update status of ecstatic-morse

ecstatic-morse:
Remove me from const-eval but add me to the Polonius team.

I haven't been useful to const-eval in a long time, but hope to be moderately useful elsewhere.

cc @lqd


lqd:
👍

PR #649, opened 2021-10-08
Add subcommand for listing teams

rylev:
I'd like to start cleaning up the teams repo so that we can fully use it for the things it's supposed to be used for (e.g., permissions, team overview, etc.). This is going to be a long process, but one of the first things is to make sure we have ways to visualize team data. 

I've added a subcommand for viewing all teams/working groups/project groups at different levels of granularity. This can definitely be improved but this is a good start for being able to see this data more clearly.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,7 +8,9 @@ mod schema;
 mod static_api;
 mod validate;
 
-use crate::{data::Data, schema::Email};
+use data::Data;
+use schema::{Email, Team, TeamKind};
+
 use failure::{err_msg, Error};
 use log::{error, info, warn};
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -37,6 +39,26 @@ enum Cli {
     StaticApi { dest: String },
     #[structopt(name = "show-person", help = "print information about a person")]
     ShowPerson { github_username: String },
+    #[structopt(name = "dump-teams", help = "Lists all teams")]
+    DumpTeams {
```
camelid, line 43: Maybe rename this to `show-teams` since it's closer in output format to `show-person`?

PR #650, opened 2021-10-08
Adding embedded-wg team definitions

ryankurte:
Adds the remainder of the embedded-wg teams based on membership [here](https://github.com/rust-embedded/wg#organization) per https://github.com/rust-embedded/wg/issues/442, towards moving team mailinglists over to central hosting.

cc. @pietroalbini

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -4,6 +4,7 @@ allowed-mailing-lists-domains = [
     "docs.rs",
     "rust-embedded.org",
     "rustconf.com",
+    "teams.rust-embedded.org",
 ]
 
 allowed-github-orgs = [
--- a/config.toml
+++ b/config.toml
@@ -4,7 +4,6 @@ allowed-mailing-lists-domains = [
     "docs.rs",
     "rust-embedded.org",
     "rustconf.com",
-    "teams.rust-embedded.org",
 ]
 
 allowed-github-orgs = [
--- a/teams/wg-embedded-tools.toml
+++ b/teams/wg-embedded-tools.toml
@@ -1,0 +1,20 @@
+name = "wg-embedded-tools"
+subteam-of = "wg-embedded"
+kind = "working-group"
+
+[people]
+leads = []
+members = [
+    "almindor",
+    "Disasm",
+    "thejpster",
+    "dkhayes117",
+]
+
+[website]
+name = "Embedded Tools Team"
+description = "Develops and maintains core embedded tools"
+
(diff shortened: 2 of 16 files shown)


config.toml
```diff
@@ -4,6 +4,7 @@ allowed-mailing-lists-domains = [
     "docs.rs",
     "rust-embedded.org",
     "rustconf.com",
+    "teams.rust-embedded.org",
```
pietroalbini, line 7: I think we can add this later when/if we actually move mailing lists over.

pietroalbini (dismissed):
Sorry for taking long to review! Just a small change.

config.toml
ryankurte: i can remove it if you prefer, but my intent is to move the mailing lists over asap (though i am not sure about the rest of this process?)

adamgreig:
Any chance we could get this merged, possibly without the new URL? We've had some changes to the teams since the PR, but it's probably easier to apply them on top of this rather than resolve a merge afterwards.

(or @ryankurte maybe you could update the PR with the changes? we need to remove thejpster from risc-v, add him to tools, and add YuhanLiin to msp430)

adamgreig:
Friendly ping for anyone to review these changes please?

@ryankurte, could you add @newam to the cortex-m team in this PR, too?

teams/wg-embedded-tools.toml
```diff
@@ -0,0 +1,20 @@
...
+kind = "working-group"
+
+[people]
+leads = []
+members = [
+    "almindor",
+    "Disasm",
+    "thejpster",
+    "dkhayes117",
+]
```
Emilgardis, line 12: This member list seems wrong

therealprof:
@ryankurte Can you address the wrong and missing team members?

teams/wg-embedded-tools.toml
ryankurte: huh, good catch, thanks! i must have got addled somewhere in the manual transcription.

therealprof (approved):
@pietroalbini Any chance to get this merged?

rylev (approved):
Sorry for taking so long. There's kind of a lack of a clear policy on how we want to handle working groups going forward which led to my hesitation to review this particular change (sorry. I should have communicated more clearly). That being said, since this is simply an addition to an existing working group and doesn't fundamentally change the structure of the project, I think it's fine to merge.

PR #653, opened 2021-10-16
Create new log team

Thomasdezeeuw:
To maintain the log crate https://github.com/rust-lang/log.

I've added @KodrAus, @sfackler and myself to it. If either of you don't want to be part of the team let me know.

### DIFF
--- a/teams/log.toml
+++ b/teams/log.toml
@@ -1,0 +1,20 @@
+name = "logs"
+subteam-of = "libs"
+
+[people]
+leads = []
+members = [
+    "KodrAus",
+    "sfackler",
+]
+
+[permissions]
+bors.libc.review = true
+
+[[github]]
+orgs = ["rust-lang"]
+
+[website]
+name = "Log crate team"
+description = "Developing and maintaining the log crate"
+repo = "https://github.com/rust-lang/log"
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
> This hasn't been discussed by the existing `log` maintainers or the libs team yet.

You're right I'm doing this based on @KodrAus suggestion in https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Log.20crate.20maintainers/near/257767087.

teams/log.toml
```diff
@@ -0,0 +1,21 @@
...
+[people]
+leads = []
+members = [
+    "KodrAus",
+    "sfackler",
+    "Thomasdezeeuw",
+]
+
+[permissions]
+bors.libc.review = true
```
JohnTitor, line 13: I don't think we need this, copy/paste error?

Thomasdezeeuw: Yes, will fix when/if the team is approved.

Thomasdezeeuw (commented):
Ah, I didn't know there was a process for creating a new team. What would be the way forward to get that started?

Thomasdezeeuw:
I'm guessing this isn't happening, so I'm going to close this.

PR #665, opened 2021-11-05
remove nikomatsakis from wg-security-response

nikomatsakis:
I'm stepping back from this WG to make more time for other things. ❤️

### DIFF
--- a/teams/wg-security-response.toml
+++ b/teams/wg-security-response.toml
@@ -8,6 +8,9 @@ members = [
     "pietroalbini",
     "steveklabnik",
 ]
+alumni = [
+    "nikomatsakis"
+]
 
 [[github]]
 team-name = "security"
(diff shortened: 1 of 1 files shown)


teams/wg-security-response.toml
```diff
@@ -5,7 +5,6 @@ kind = "working-group"
 leads = []
 members = [
     "cuviper",
-    "nikomatsakis",
```
pietroalbini, line 8: Do you want to add yourself to alumni?

pietroalbini (dismissed):

teams/wg-security-response.toml
nikomatsakis: Ah, sure. Forgot that existed!

pietroalbini (approved):

PR #666, opened 2021-11-08
Remove mbrubeck from the moderation team.

mbrubeck:
I am retiring from the moderation team.

### DIFF
--- a/teams/mods.toml
+++ b/teams/mods.toml
@@ -3,7 +3,6 @@ name = "mods"
 [people]
 leads = []
 members = [
-    "mbrubeck",
     "BurntSushi",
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks for everything you've done @mbrubeck. It has been an honor. :-)

lqd:
The mods team can have an `alumni` field, right ? So that Matt is not removed from the web site (if they would want that, of course)

mbrubeck:
I am still in the "Discourse moderators" list that appears on the same page on the web site.  That's fine with me for now.

jntrnr (approved):

PR #698, opened 2022-01-26
Rename wg-async-foundations to wg-async

tmandry:
See also https://github.com/rust-lang/wg-async/pull/267

### DIFF
--- a/teams/wg-async.toml
+++ b/teams/wg-async.toml
@@ -1,4 +1,4 @@
-name = "wg-async-foundations"
+name = "wg-async"
 kind = "working-group"
 
 [people]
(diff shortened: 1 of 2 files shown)


teams/wg-async.toml
```diff
@@ -38,7 +38,7 @@ alumni = [
 orgs = ["rust-lang", "rust-lang-nursery"]
 
 [website]
-name = "Async Foundations working group"
+name = "Async working group"
 description = "Pursuing core language and library support for async-await"
```
camelid, line 42: nit: I think `async/await` is a more standard spelling? The current spelling is probably fine though.

rylev (approved):
🎉

teams/wg-async.toml
rylev: I'm going to merge, but feel free to submit a change to this in a new PR

PR #707, opened 2022-02-24
Add Weihang Lo to the Cargo team

ehuss:
cc @rust-lang/cargo @weihanglo

### DIFF
--- a/people/weihanglo.toml
+++ b/people/weihanglo.toml
@@ -1,0 +1,4 @@
+name = 'Weihang Lo'
+github = 'weihanglo'
+github-id = 14314532
+email = 'me@weihanglo.tw'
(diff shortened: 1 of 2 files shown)


ehuss:
I'm not sure, would it make sense to set up the GitHub team to be synced with the definition here?

rylev (approved):

rylev:
@ehuss I have it on my todo list to do this for all teams where it's not the current case. @pietroalbini had set this up for some teams in the past.

camelid:
I think the

```toml
[[github]]
```

that's in the team file should mean it's already the case that it's auto-synced?

ehuss:
Oh, indead it is, thanks @camelid!

PR #712, opened 2022-03-08
Remove myself from the Cargo team

alexcrichton:
Lately I don't have quite the time I used to have to dedicate to Cargo
and I'm also lacking much of my prior enthusiasm, so I'm going to be
stepping down from the Cargo team. While I was here I figured I'd also
update the extra repos that I have permission to review and remove
myself there as well, I haven't looked at those in quite some time so
it's probably good to clean that up as well.

### DIFF
--- a/people/alexcrichton.toml
+++ b/people/alexcrichton.toml
@@ -3,7 +3,3 @@ github = "alexcrichton"
 github-id = 64996
 irc = "acrichto"
 email = "alex@alexcrichton.com"
-
-[permissions]
-bors.rustup-rs.review = true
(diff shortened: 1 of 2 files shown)


Eh2406:
Is there some reason you don't want to be on the `cargo alumni` list?

alexcrichton:
Oops no just didn't see that there!

jonhoo:
You will be missed Alex — thanks for everything!

hi-rustin:
Your kindness and patience with contributors has been an inspiration to me to keep contributing. Thank you for everything.

PR #720, opened 2022-03-22
Add a flag for whether a user has dev_desktop access.

oli-obk:
Right now all non-alumni team members have dev_desktop access.
The field is only relevant for the `all` meta-team.
We may use this field to disable access to users in the future

r? @pietroalbini 

This PR is used to enable https://github.com/rust-lang/simpleinfra/pull/79 to easily decide whether a user has access to the dev desktops, without requiring changes to the dev-desktops in the future (we can just modify the teams repo to switch flags to `false`)

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -41,6 +41,7 @@ permissions-bors-repos = [
 permissions-bools = [
     "perf",
     "crater",
+    "dev-desktop",
 ]
 
 permissions-crates-io-ops-bot-apps = [
--- a/teams/compiler.toml
+++ b/teams/compiler.toml
@@ -29,6 +29,7 @@ alumni = [
 [permissions]
 perf = true
 crater = true
+dev-desktop = true
 bors.rust.review = true
 bors.miri.review = true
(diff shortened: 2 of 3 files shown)


pietroalbini:
Rather than adding a new field to each member, I think it's best to handle this through the permissions system.

You can add `dev-desktop` here:

https://github.com/rust-lang/team/blob/40d27aee4b765d02b0a19ee25364d43d8824f315/config.toml#L41-L44

...and then add the `dev-desktop` permission like any other permission:

```toml
[permissions]
dev-desktop = true
```

On the API side, it will create a file at the following URL, looking [like this](https://team-api.infra.rust-lang.org/v1/permissions/perf.json):

```
https://team-api.infra.rust-lang.org/v1/permissions/dev-desktop.json
```

To start, maybe we could add it to `infra` and `compiler` as the first test group?

oli-obk:
oh great! this is what I was looking for

rylev (approved):
🎉

PR #730, opened 2022-04-05
Zulip groups endpoint

rylev:
Added as part of @pietroalbini's [feedback](https://github.com/rust-lang/sync-team/pull/3#issuecomment-1086582926) to https://github.com/rust-lang/sync-team/pull/3

Adds an endpoint to the Team API "zulip-groups.json" which provides all the Zulip groups defined in the teams toml. 

The definition in toml mirrors the "[[lists]]" schema (just with a "name" key instead of "address"). The members of the Zulip groups are either identified by a Zulip user id or an email.


rust_team_data/src/v1.rs
```diff
@@ -88,6 +88,23 @@ pub struct Lists {
     pub lists: IndexMap<String, List>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ZulipGroup {
+    pub name: String,
+    pub members: Vec<ZulipGroupMember>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
pietroalbini, line 97: Nit: it'd be nice to have a `#[serde(rename_all = "snake_case")]` to be consistent with the rest of the API.

src/schema.rs
```diff
@@ -302,6 +303,56 @@ impl Team {
...
+                let member = data
+                    .person(member)
+                    .ok_or_else(|| err_msg(format!("member {} is missing", member)))?;
+                if let Some(zulip_id) = member.zulip_id {
+                    list.members.push(ZulipGroupMember::Id(zulip_id));
+                }
+                if let Email::Present(email) = member.email() {
+                    list.members
+                        .push(ZulipGroupMember::Email(email.to_string()));
+                }
```
pietroalbini, line 346: We should probably add either the ID or the email (preferring the ID if present), not both. It's more intuitive to me that if an explicit ID is present it's going to be used rather than the email address. There could also be a *slight* chance that the ID and the email map to different accounts, which would be weird.

src/schema.rs
```diff
@@ -509,6 +560,20 @@ pub(crate) struct TeamList {
...
+#[derive(serde_derive::Deserialize, Debug)]
+#[serde(rename_all = "kebab-case", deny_unknown_fields)]
+pub(crate) struct RawZulipGroup {
+    pub(crate) name: String,
+    #[serde(default = "default_true")]
+    pub(crate) include_team_members: bool,
+    #[serde(default)]
+    pub(crate) extra_people: Vec<String>,
+    #[serde(default)]
+    pub(crate) extra_emails: Vec<String>,
```
pietroalbini, line 572: It'd be nice to have an `extra_zulip_ids` here in addition to `extra_emails`.

src/static_api.rs
```diff
@@ -146,6 +147,31 @@ impl<'a> Generator<'a> {
...
+    fn generate_zulip_groups(&self) -> Result<(), Error> {
+        let mut groups = IndexMap::new();
+
+        for group in self.data.zulip_groups()?.values() {
+            let members = group.members().to_vec();
+            groups.insert(
+                group.name().to_string(),
+                v1::ZulipGroup {
+                    name: group.name().to_string(),
+                    members: members
```
pietroalbini, line 159: The list of members here is not sorted, which could result in unreproducible ordering for the members list (in the test for example it's not ordered).

tests/static-api/_expected/v1/zulip-groups.json
```diff
@@ -0,0 +1,15 @@
+{
+  "groups": {
+    "T-foo": {
+      "name": "T-foo",
+      "members": [
+        {
+          "Email": "user1@example.com"
```
pietroalbini, line 7: Can you test adding the Zulip ID to one of those users?

pietroalbini (changes requested):
Thanks for the PR! This mostly looks good, I left some minor feedback.

pietroalbini:
There should also be some validation rules, at least to prevent conflicting group names across teams.

pietroalbini (approved):
Looks good! Feel free to merge :slightly_smiling_face:

PR #736, opened 2022-04-13
Add GitHub info for wg-nll

rylev:
The [types team](https://github.com/rust-lang/rfcs/pull/3254) is the likely successor of the work.

Note: I also added a `[[github]]` annotation to the wg defintion in the unlikely case that the wg is revived, since until now the permissions have been manually curated on GitHub.

cc @nikomatsakis @pnkfelix

### DIFF
--- a/teams/wg-nll.toml
+++ b/teams/wg-nll.toml
@@ -11,3 +11,6 @@ name = "Non-Lexical Lifetimes (NLL) working group"
 description = "Implementing the new MIR-based borrow checker and non-lexical lifetimes in Rust"
 repo = "https://rust-lang.github.io/compiler-team/working-groups/nll/"
 zulip-stream = "t-compiler/wg-nll"
+
+[[github]]
+orgs = ["rust-lang"]


jackh726:
I think it's probably just worth waiting until NLL stabilization. That's hopefully coming soon™

While I think most of the members here aren't active much anyways, the group as an *entity* is good to help organize the final stabilization push. (I'm mostly thinking the wg-nll zulip stream. The formal working group might very well be overkill.)

rylev:
I'll let @nikomatsakis weigh in here, but I am of the opinion that when in doubt we should keep the wg around. However, I believe Niko was keen on seeing this responsibility moved to the new types team.

jackh726:
Oh this definitely would fall under the types team. But I think this would still exist, either in the form of an initiative or a working group.

rylev:
@jackh726 that sounds reasonable. One additional thing to consider is that we will also be syncing Zulip user groups through the team repo (see https://github.com/rust-lang/sync-team/pull/3) so if we want @wg-nll to work on Zulip, we should still keep it around.

rylev:
Ok I shifted this to just add the GitHub info to the wg definition since that should be without controversy. We can still archive at a later point.

Mark-Simulacrum:
This will create a duplicate team with [wg-compiler-nll](https://github.com/orgs/rust-lang/teams/wg-compiler-nll/members), which has many more members. Happy to merge, but we should probably have a follow-up of deduplicating here one way or another. I'm not sure if there's repositories wg-compiler-nll has access to.

rylev:
@Mark-Simulacrum indeed. The plan is to remove the wg-compiler-nll group in favor of wg-nll. Before removing it, I would check with the members who are not in the working group but are in the old user group to ensure we reach consistency.

Mark-Simulacrum (approved):

PR #741, opened 2022-04-20
Add new marker team for Cloud Compute Program

JoelMarcey:
A dedicated team with access to the cloud desktops associated with the [Cloud Compute Program](https://foundation.rust-lang.org/news/2021-11-16-news-announcing-cloud-compute-initiative/).


JoelMarcey:
Should we also add a dedicated Zulip group for this as well? (e.g., `[[zulip-groups]]`)

JoelMarcey:
cc @yaahc
cc @Mark-Simulacrum

Mark-Simulacrum:
I don't think Zulip/GitHub are likely to see much use so I'd not bother configuring those. But they don't hurt either so either way is fine.

For onlookers, this is likely a temporary team - mostly for (very) alpha testing.

Modulo CI seems ok to me.

compiler-errors:
Is this the same thing as what @oli-obk added in team#720 ?

oli-obk:
Yea you should be able to just give individual users the dev-desktop permission, but I also don't mind having a separate team yo add people to

compiler-errors:
Ah, missed the part saying:

```
[permissions]
dev-desktop = true
```

Alright. Sorry for the noise, hehe.

JoelMarcey:
I thought having a specific, explicit team would just make things clearer and cleaner. And particularly easier to add folks and, if we ever need to, remove from just one place.

Mark-Simulacrum (approved):

PR #747, opened 2022-05-02
Add GitHub team for unsafe-code-guidelines wg

rylev:
Currently the [unsafe code guidelines GitHub team](https://github.com/orgs/rust-lang/teams/wg-unsafe-code-guidelines) is out of sync with the team repo definition. This will create a team automatically on GitHub that's kept in sync. Once this is merged, we would delete the old team in favor of the newly synced team.

However, if this is merged, this will remove a few people from the GitHub team: @comex, @gereeter, @strega-nil, @Mark-Simulacrum, and @ehsanmok. If removing any of these folks is incorrect, we just need to add them to the members list. 

cc @nikomatsakis, @avadacatavra, @RalfJung


ehsanmok:
I hope I can still remain in the wg despite struggling to find time :(

Mark-Simulacrum (approved):

RalfJung:
> If removing any of these folks is incorrect, we just need to add them to the members list.

We used the GH team as a notification group, which I guess is why it is not aligned with team membership as shown on the website.

rylev:
@RalfJung there's several ways to fix that I believe. We either make the missing folks members of the working group or we create a marker team with them.

PR #748, opened 2022-05-02
Add GitHub team for grammar wg

rylev:
Currently the [wg-grammar GitHub team](https://github.com/orgs/rust-lang/teams/wg-grammar) is out of sync with the team repo definition. This will create a team automatically on GitHub that's kept in sync. Once this is merged, we would delete the old team in favor of the newly synced team.

However, if this is merged, this will remove a few people from the GitHub team: @nikomatsakis, @qmx, @ehuss, @petrochenkov. If removing any of these folks is incorrect, we just need to add them to the members list. 

cc @eddyb

### DIFF
--- a/teams/wg-grammar.toml
+++ b/teams/wg-grammar.toml
@@ -7,7 +7,6 @@ leads = ["eddyb"]
 members = [
     "eddyb",
     "CAD97",
-   
 ]
 alumni = [
     "Centril",
@@ -17,3 +16,6 @@ alumni = [
 [website]
 name = "Grammar working group"
 description = "Working out the official, formal grammar for Rust and validating it against existing implementations"
+
+[[github]]
+orgs = ["rust-lang"]


rylev:
@Mark-Simulacrum thoughts on this? I'm not sure if the grammar wg is even still active?

ehuss:
The team is not active. I think it would be fine to merge this. If the group ever resurrects, we can fix the membership as needed.  

I guess there is a broader discussion of how to disband groups or teams. I'm not aware of any documentation on how that should happen?  I guess the repo would be archived, and the team somehow shut down here.  But I don't think that is necessary right now, I don't think anyone would object to leaving things open.  I would like to see things start up again, but I don't have the time to drive things.

rylev:
> I guess there is a broader discussion of how to disband groups or teams. I'm not aware of any documentation on how that should happen?

For the Team repo, I think the only thing that needs to be done is moving the wg into the archive folder. 

I'll merge this for now. We can do another pass to see if this wg (and potentially others) should be wound down.

Mark-Simulacrum (approved):

PR #753, opened 2022-05-10
Add myself to the Windows ping group

ChrisDenton:


ChrisDenton:
I realized I never actually added myself to the Windows ping group. I usually get pinged anyway if it's an area I may be able to help with. Still, it can't hurt.

rylev (approved):
🎉

PR #754, opened 2022-05-11
add digama0 to UCG WG

RalfJung:
They have been contributing constructively for quite a while, it's about time we recognize that.
Cc @digama0 @rust-lang/wg-unsafe-code-guidelines


RalfJung:
@digama0 is not in the rust-lang org yet, not sure if that means any extra process is required?

PR #756, opened 2022-05-11
add-person does not seem to need a GITHUB_TOKEN

RalfJung:
At least, I just ran this without setting any env var, and it worked fine...

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,7 @@ cargo run check
 ### Adding a person to the repository
 
 It's possible to fetch the public information present in a GitHub profile and
-store it in a person's TOML file. To do that you need to have the
-`GITHUB_TOKEN` environment variable setup with a valid personal access token,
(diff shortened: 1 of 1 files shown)


rylev (approved):
Tested this, and it indeed works without a token.

m-ou-se:
It works, but without a token there's a rate limit of 60 API calls per hour per IP address (instead of 5000 per hour per user).

```
$ cargo r add-person m-ou-se
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/rust-team add-person m-ou-se`
ERROR https://api.github.com/users/m-ou-se: Client Error: 403 Forbidden
```

rylev:
Might be better if we detect 403 errors, see if the appropriate token env variable is set, and if not say that they might need that env variable to make things work.

PR #759, opened 2022-05-20
Add vim and emacs marker teams

rylev:
This adds vim and emacs marker teams that mirror the teams in GitHub: [rust.vim](https://github.com/orgs/rust-lang/teams/rust-vim) and [rust-mode](https://github.com/orgs/rust-lang/teams/rust-mode). This is part of the effort to ensure we don't have any teams on GitHub that are not controlled by this repo. Here are some things we may want to consider perform merging:
* The use of marker teams is a bit odd. One could argue this should be a subteam of the IDEs team, but I'm not sure if that team is actually active right now. However, whether IDEs is active or not might not actually matter. cc @Xanewok 
* I've named the teams "vim" and "emacs" respectively instead of using the "rust.vim" and "rust-mode" names employed on GitHub. The new names are more understandable I believe, but they might be more general than we actually need. 
* This PR adds @MicahChalmer, @brotzeit, @chris-morgan, and @da-x to the team repo. However, the use of marker teams means that they are still not "officially" a part of the Rust project.

### DIFF
--- a/people/MicahChalmer.toml
+++ b/people/MicahChalmer.toml
@@ -1,0 +1,3 @@
+name = "Micah Chalmer"
+github = "MicahChalmer"
+github-id = 698400
--- a/people/brotzeit.toml
+++ b/people/brotzeit.toml
@@ -1,2 +1,3 @@
+name = "brotzeit"
 github = "brotzeit"
 github-id = 25367303
--- a/people/brotzeit.toml
+++ b/people/brotzeit.toml
@@ -1,0 +1,2 @@
+github = "brotzeit"
+github-id = 25367303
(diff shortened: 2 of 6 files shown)


Mark-Simulacrum (approved):

Xanewok:
What's the usage of mirroring the structure on our side? Better transparency on the official Rust team list or maybe we can ping them now?

rylev:
The main motive of this is we want all GitHub teams (through which repo permissions are established) to be controlled by the team repo. Now the vim and emacs repos permissions are controlled through the team repo instead of through the GitHub interface.

PR #760, opened 2022-05-22
Archive wg-traits and create types team

jackh726:
Also, welcome @compiler-errors (officially).

Zulip and github things should be renamed before this is merged.


rylev:
cc @wesleywiser and @pnkfelix since we'll want approval from the compiler team before merging.

The failure in CI is from [here](https://github.com/rust-lang/team/blob/9de1895b14f7515b4bc53a682628de01f8ca0cf6/src/validate.rs#L190-L222). We have a validation that all people in the Team repo are at least on one team. However, this does not take into account archived teams. I believe the right thing to do is to update this logic to all consider archived teams, but I'd like to check with @pietroalbini first.

jackh726:
Is the RFC *not* approval? And why compiler and not lang? (If it's because the labeled subteam, then it really should be both and this is just a limitation of the team repo)

Let me know if I need to make a PR to fix the archived team issue.

wesleywiser (dismissed):

teams/types.toml
```diff
@@ -0,0 +1,26 @@
+name = "types"
+subteam-of = "compiler"
```
compiler-errors, line 2: can we leave a comment here mentioning that this is both a subteam of lang and compiler, but filed under compiler due to technical reasons?

teams/types.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[website]
+name = "Types team"
+description = "Working to implement and formally define the semantics of the Rust language"
+repo = "https://github.com/rust-lang/types-team"
+zulip-stream = "t-types"
+
+[[github]]
+orgs = ["rust-lang", "rust-lang-nursery"]
+
```
rylev, line 24: You can also add a `zulip-group` here to automatically sync this team with a Zulip user group. For example: https://github.com/rust-lang/team/blob/9de1895b14f7515b4bc53a682628de01f8ca0cf6/teams/infra.toml#L51-L52

jackh726: This is a group, not a stream, right?

rylev: Correct. We don't yet have support for syncing topics.

jackh726:
Thoughts on me dropping `wg-traits.toml` from archive (or commenting out the members causing problems with a FIXME) to move this forward without blocking on fixing the archived teams check?

pietroalbini:
So, since what to do with archived teams is a bit of a bigger question what I'd do is keep the wg-traits team around but remove it from the website and all the actual permissions, ideally with a todo comment. That way we don't lose any data until we figure out what to do with archived teams.

jackh726:
Okay, I can move it back to a regular team. Do I just remove the `[[website]]` section?

pietroalbini:
That, and also `[[github]]` and `[permissions]`.

jackh726:
Updated for reviews

teams/wg-traits.toml
```diff
@@ -1,3 +1,7 @@
+# Note: this has effectively been replaced by the types team
+# This team remains until we figure out what to do with archived teams
```
rylev, line 2: nit: the issue is that we can't archive teams (we already have an archive folder). The issue is that archiving doesn't work if a person in the Team repo is *only* on an archived team. Can we make that explicit here so that no one is tempted in the future to "fix" this by moving this team to the archive folder?

rylev (changes requested):
Looks good, but I'd like us to address the one issue I found.

teams/wg-traits.toml
jackh726: Fixed

rylev (approved):
🎉

nikomatsakis:
This looks great, thanks @jackh726 for preparing it

jackh726:
Anything else that needs to be done for this to be merged?

PR #761, opened 2022-05-25
Add Zulip Groups

rylev:
This moves almost all Zulip User groups to be controlled by the Team repo. 

The following Zulip groups are still not controlled by the Team repo:
* T-compiler/meeting
* WG-profile-guided-optimization
* WG-pipelining
* mods
* WG-governance
* T-community/l10n
* WG-prioritization/alerts
* project-custom-linter

Additionally the WG-async-foundations user group will be renamed to WG-async to reflect the new name of the working group.

Because there has been some drift between Zulip user group membership and team membership this may remove (or add) certain people to these Zulip groups. This should be fairly rare and something we can correct 1 by 1.

This also adds a new type of validator: the Zulip validator which has access to the Zulip API. This means that the CI will need a username and token (much like the team sync repo).

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -90,12 +90,8 @@ pub(crate) fn validate(data: &Data, strict: bool, skip: &[&str]) -> Result<(), E
 
     let zulip = ZulipApi::new();
     if let Err(err) = zulip.require_auth() {
-        if strict {
-            return Err(err);
-        } else {
-            warn!("couldn't perform checks relying on the Zulip API, some errors will not be detected");
-            warn!("cause: {}", err);
-        }
+        warn!("couldn't perform checks relying on the Zulip API, some errors will not be detected");
+        warn!("cause: {}", err);
     } else {
         for check in ZULIP_CHECKS {
             if skip.contains(&check.name) {
@@ -626,8 +622,9 @@ fn validate_zulip_group_ids(data: &Data, errors: &mut Vec<String>) {
                     && matches!(member.email(), Email::Missing | Email::Disabled)
                 {
                     bail!(
-                        "person `{}` is a member of a Zulip user group but has no a Zulip id nor an enabled email address",
-                        member.github()
+                        "person `{}` in '{}' is a member of a Zulip user group but has no a Zulip id nor an enabled email address",
+                        member.github(),
+                        team.name()
                     );
                 }
             }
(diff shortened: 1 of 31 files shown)


Mark-Simulacrum:
Hm, I'm not sure we can scope Zulip tokens to read only, which would make placing it in PR CI relatively complicated. I'm not sure what the best way to work around that would be - maybe we can have (for example) triagebot have a read-only API, but that also seems complicated.

rylev:
@Mark-Simulacrum I made the Zulip validation optional. We can still validate Zulip ids locally but given that tokens need to be read/write admin tokens, I agree it's probably not a good idea to run this in CI. I think we'll eventually want to move to a model where all members of a team with an associated Zulip user group need to include a Zulip id in their person config, but we can do that in a separate PR. 

As such I have only updated the teams where we can find the Zulip id for *all* the members. I can work on adding Zulip info for the 30 or so people who are missing that info.

Mark-Simulacrum:
What is the failure mode if there's a missing zulip ID -- do we just skip that user, or stop processing that group entirely? (Or all groups)?

rylev:
> What is the failure mode if there's a missing zulip ID -- do we just skip that user, or stop processing that group entirely? (Or all groups)?

The user is just [filtered](https://github.com/rust-lang/team/blob/e40aaf320e86719df869be93e2d5d4bf4eac31c0/src/static_api.rs#L165) out. This means that from the sync-team tool's perspective that user is simply not on that team (for the purposes on syncing Zulip), and they just won't be included.

src/validate.rs
```diff
@@ -562,8 +622,9 @@ fn validate_zulip_group_ids(data: &Data, errors: &mut Vec<String>) {
                     && matches!(member.email(), Email::Missing | Email::Disabled)
                 {
                     bail!(
-                        "person `{}` is a member of a Zulip user group but has no a Zulip id nor an enabled email address",
-                        member.github()
+                        "person `{}` in '{}' is a member of a Zulip user group but has no a Zulip id nor an enabled email address",
```
Mark-Simulacrum, line 625:
```suggestion
                        "person `{}` in '{}' is a member of a Zulip user group but has no Zulip id nor an enabled email address",
```

Mark-Simulacrum (approved):

Mark-Simulacrum (approved):

PR #764, opened 2022-05-30
Check synced

rylev:
Adds a `check_synced` command which checks that the contents of the team repo are synced remotely. Currently this only checks GitHub team membership but can be extended to more things (e.g., Zulip user groups). 

I needed this functionality for work on moving all GitHub team management to the team repo.

### DIFF
--- a/src/check_synced.rs
+++ b/src/check_synced.rs
@@ -1,0 +1,67 @@
+//! Validates that remote data matches data in team repo.
+
+use crate::data::Data;
+use crate::github::{self, GitHubApi};
+use crate::schema;
+use log::error;
+use std::collections::HashMap;
+
+pub(crate) fn check(data: &Data) -> Result<(), failure::Error> {
+    const BOT_TEAMS: &[&str] = &["bors", "bots", "rfcbot", "highfive"];
+    let github = GitHubApi::new();
+    let mut remote_teams = github
+        .teams()?
+        .into_iter()
+        .filter(|t| !BOT_TEAMS.contains(&t.name.as_str()))
+        .map(|team| {
+            let members = github.team_members(team.id)?;
+            Ok((team.name.clone(), (team, members)))
+        })
+        .collect::<Result<HashMap<_, _>, failure::Error>>()?;
+
+    for team in data.teams() {
+        let local_teams = team.github_teams(&data)?;
+        let local_team = local_teams.into_iter().find(|t| t.org == "rust-lang");
+        let local_team = match local_team {
+            Some(t) => t,
+            None => continue,
+        };
+        match remote_teams.remove(local_team.name) {
+            Some((_, remote_members)) => {
+                check_team_members_match(local_team, remote_members);
+            }
+            None => error!("Team '{}' in team repo but not on GitHub", local_team.name),
+        }
+    }
+    for (remote_team_name, _) in remote_teams {
+        error!("Team '{}' on GitHub but not in team repo", remote_team_name)
+    }
+    Ok(())
+}
+
+fn check_team_members_match(
+    local_team: schema::GitHubTeam,
+    remote_members: Vec<github::GitHubMember>,
+) {
+    let mut local_members = local_team.members;
+    for remote_member in remote_members.iter() {
+        let pos = local_members
+            .iter()
+            .position(|(_, id)| id == &remote_member.id);
+        match pos {
+            Some(pos) => {
+                local_members.swap_remove(pos);
+            }
+            None => error!(
+                "'{}' is on GitHub '{}' team but not in team repo definition",
+                remote_member.name, local_team.name
+            ),
+        }
+    }
+    for (local_member_name, _) in local_members {
+        error!(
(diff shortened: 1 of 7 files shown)


src/check_synced.rs
```diff
@@ -0,0 +1,68 @@
...
+use log::error;
+use rayon::prelude::*;
+use std::collections::HashMap;
+
+pub(crate) fn check(data: &Data) -> Result<(), failure::Error> {
+    const BOT_TEAMS: &[&str] = &["bors", "bots", "rfcbot", "highfive"];
+    let github = GitHubApi::new();
+    let mut remote_teams = github
+        .teams()?
+        .into_par_iter()
```
Mark-Simulacrum, line 15: IIRC, GitHub recommends against concurrent requests to their API, but I suppose it seems OK for our purposes. Is this a major win in practice? I know the API can be rather slow, so it very well may be, but given the added dependency on rayon I'd like to make sure it's worth it (e.g., shaving dozens of seconds off).

src/check_synced.rs
```diff
@@ -0,0 +1,68 @@
...
+        .filter(|t| !BOT_TEAMS.contains(&t.name.as_str()))
+        .map(|team| {
+            let members = github.team_members(&team.name)?;
+            Ok((team.name.clone(), (team, members)))
+        })
+        .collect::<Result<HashMap<_, _>, failure::Error>>()?;
+
+    for team in data.teams() {
+        let local_teams = team.github_teams(&data)?;
+        let local_team = local_teams.into_iter().find(|t| t.org == "rust-lang");
```
Mark-Simulacrum, line 25: Hm, I think it would be worthwhile to have a const or similar with a list of orgs we validate like this. For example, rust-lang-ci and rust-lang-nursery likely should also be included here.

src/github.rs
```diff
@@ -146,8 +147,65 @@ impl GitHubApi {
         }
         Ok(result)
     }
+
+    /// Get all teams for the rust-lang org
+    pub(crate) fn teams(&self) -> Result<Vec<GitHubTeam>, Error> {
+        Ok(self
+            .prepare(true, Method::GET, "orgs/rust-lang/teams?per_page=100")?
```
Mark-Simulacrum, line 154: Hm, it seems likely that we should validate we're not hitting pagination limits here. We currently have 119 rust-lang/team teams (not all of which correspond to an entity on GitHub), so it's likely that we'll pretty soon run up against a 100 team limit here if we're not already.

I don't think it's necessarily true that we need to block on doing pagination, but we shouldn't silently skip teams.

src/github.rs
```diff
@@ -146,8 +147,65 @@ impl GitHubApi {
...
+                        "orgs/rust-lang/teams/{}/members?per_page=100&page={}",
+                        name, page_num
+                    ),
+                )?
+                .send()?
+                .error_for_status()?
+                .json()?;
+            let len = page.len();
+            members.extend(page);
+            if len < 100 {
```
Mark-Simulacrum, line 179: FWIW, the "proper" way to do this is via the Link header (https://docs.github.com/en/rest/guides/traversing-with-pagination), but that's a pain so this seems OK for now.

src/github.rs
```diff
@@ -146,8 +147,65 @@ impl GitHubApi {
...
+                break;
+            }
+            page_num += 1;
+        }
+        let invites: Vec<GitHubMember> = self
+            .prepare(
+                true,
+                Method::GET,
+                &format!("orgs/rust-lang/teams/{}/invitations?per_page=100", name),
```
Mark-Simulacrum, line 188: As before, I'd like to see validation on this that we're not running up against the limit.

pietroalbini:
I have thoughts on this, please don't merge until I had a chance to give feedback :slightly_smiling_face:

src/check_synced.rs
rylev: I agree. I didn't want to complicate things initially with other orgs, but we should definitely do this - though perhaps in a follow-up.

src/github.rs
rylev: Weirdly, these API endpoints don't seem to be returning `Link` headers:

```bash 
$ curl -I https://api.github.com/orgs/rust-lang/teams/compiler/members\?per_page\=100\&page\=1 -H "Authorization: token $MYTOKEN" | grep -i link
access-control-expose-headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset
```

A quick search shows several people noting that many endpoints don't return that header. 🤷‍♀️

rylev:
@pietroalbini any thoughts on this?

pietroalbini (approved):
At a glance looks correct. It seems a bit of a duplication of what sync-team does, but if it helps you in your work then it's fine to merge it!

PR #790, opened 2022-06-29
Add sunfishcode to libs-contributors.

m-ou-se:
Welcome, @sunfishcode! And thanks for all your contributions. :)

cc @rust-lang/libs @rust-lang/libs-contributors


rylev (approved):
🎉

PR #801, opened 2022-08-03
Add ability to encode and fetch repo information

rylev:
This is the start of being able to control GitHub repos from the team repo. 

This introduces one new configuration file type ("repo" located in the "repos" folder) as well as two new endpoint types: /v1/repos.json and /v1/repos/$REPO.json.

The ultimate idea is for this to be fed to sync-team so we can move away from having GitHub admins make changes that are not visible to the rest of the project. 

I've only tried to reproduce this repo in the configuration so far.

### DIFF
--- a/repos/team.toml
+++ b/repos/team.toml
@@ -1,0 +1,8 @@
+org = "rust-lang"
+name = "team"
+description = "Rust teams structure"
+bots = ["bors", "highfive", "rustbot", "rust-timer"]
+
+[access.teams]
+core = "admin"
+mods = "maintain"
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -644,6 +644,18 @@ pub(crate) struct Repo {
     pub access: RepoAccess,
 }
 
+impl Repo {
+    const VALID_ORGS: &'static [&'static str] = &["rust-lang"];
+
+    pub(crate) fn validate(&self) -> Result<(), Error> {
+        if !Self::VALID_ORGS.contains(&self.org.as_str()) {
+            bail!("{} is not a valid repo org", self.org);
+        }
+
+        Ok(())
+    }
+}
+
 #[derive(serde_derive::Deserialize, Debug)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
 pub(crate) struct RepoAccess {
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -640,7 +640,7 @@ pub(crate) struct Repo {
     pub org: String,
     pub name: String,
     pub description: String,
-    pub bots: Vec<String>, // TODO: should this be an enum?
+    pub bots: Vec<Bot>,
     pub access: RepoAccess,
 }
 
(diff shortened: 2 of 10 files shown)


repos/team.toml
```diff
@@ -0,0 +1,8 @@
+org = "rust-lang"
+name = "team"
+description = "Rust teams structure"
+bots = ["bors", "highfive", "rustbot", "rust-timer"]
+
+[access.teams]
+core = "admin"
+mods = "maintain"
```
Mark-Simulacrum, line 8: I think the access + bots definitions here are OK for now, but I do think we may want some kind of "standard template" for repositories that you can opt out of. For example, rustbot + highfive probably should have access to anything in the org.

repos/team.toml
```diff
@@ -0,0 +1,8 @@
+org = "rust-lang"
+name = "team"
+description = "Rust teams structure"
+bots = ["bors", "highfive", "rustbot", "rust-timer"]
```
Mark-Simulacrum, line 4: Why are bors and rust-timer in this list? (highfive and rustbot also aren't used, but seem more "OK).

src/schema.rs
```diff
@@ -633,3 +633,39 @@ fn default_true() -> bool {
...
     false
 }
+
+#[derive(serde_derive::Deserialize, Debug)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub(crate) struct Repo {
+    pub org: String,
+    pub name: String,
+    pub description: String,
+    pub bots: Vec<String>, // TODO: should this be an enum?
```
Mark-Simulacrum, line 643: Yes, I think so (or we want some validation at least). Each bot will likely end up with slightly different privileges -- or even move to a github app that gets invited to particular repos, so that it's very "non-stringy".

src/schema.rs
```diff
@@ -633,3 +633,39 @@ fn default_true() -> bool {
...
+    pub access: RepoAccess,
+}
+
+impl Repo {
+    const VALID_ORGS: &'static [&'static str] = &["rust-lang"];
```
Mark-Simulacrum, line 648:
```suggestion
    const VALID_ORGS: &[&str] = &["rust-lang"];
```

I think this should work?

src/validate.rs
```diff
@@ -653,6 +654,23 @@ fn validate_zulip_group_extra_people(data: &Data, errors: &mut Vec<String>) {
     });
 }
 
+/// Ensure working group names start with `wg-`
```
Mark-Simulacrum, line 657: Stale comment?

Mark-Simulacrum (commented):
Looks OK to me overall.

repos/team.toml
rylev: This reflects the current state of the things in GitHub. I agree they likely don't need to be there, but perhaps we can make that true by using this mechanism (once the changes to sync-team land).

src/schema.rs
rylev: Unfortunately, not: https://github.com/rust-lang/rust/issues/51370

Mark-Simulacrum: I... am surprised I've never hit this before, but I guess that makes some sense.

Mark-Simulacrum (approved):

PR #803, opened 2022-08-11
Add TaKO8Ki to compiler-contributors and wg-diagnostics

estebank:


TaKO8Ki:
cc @wesleywiser @pnkfelix

TaKO8Ki:
Could someone please review this? @rust-lang/core

pnkfelix:
I'm 👍 on this.

Mark-Simulacrum (approved):

TaKO8Ki:
I have not been added to the compiler-contributors team yet, did something happen during the synchronization?

Mark-Simulacrum:
Should be fixed now. I think we have a bug where sometimes sync-team runs before caching invalidates across GitHub's CDN, which means we end up with a no-op run rather than actually taking action on the results.

cc @rylev -- we may want to adjust our CI here to either sleep for a minute or some other mitigation tactic, though this happens pretty rarely and clears up on the next run (which happens every day regardless).

PR #804, opened 2022-08-18
Add branches to repo definitions

rylev:
This adds support for protected branches to the repo definitions. The definitions look like this:

```toml
# the use of "branch" instead of protected branch makes this 
# feature more general than just protecting branches
[[branch]] 
# Should we require the user specify at least one branch with 
# a `default` field set to true for the default branch?
name = "master" 
# This field is optional. I believe GitHub will accept these even
# if the job does not exist
ci-checks = ["CI"] 
```

No other fields than the ones above are supported. I believe all other branch options that rust-lang/team has set should be the default. They are:
* Require a pull request before merging
* Require 1 approval (no exceptions for admins or any specific teams or people)
* Dismiss stale pull request approvals when new commits are pushed
* All other options are the GitHub default

When this is merged the following can be added to sync-team:
* Ensuring the branch exists
* Ensuring the proper branch protection options are set for the branch
* Removing all other branch protections

### DIFF
--- a/repos/team.toml
+++ b/repos/team.toml
@@ -6,3 +6,7 @@ bots = ["bors", "highfive", "rustbot", "rust-timer"]
 [access.teams]
 core = "admin"
 mods = "maintain"
+
+[[branch]]
+name = "master"
+ci-checks = ["CI"]
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -144,6 +144,7 @@ pub struct Repo {
     pub description: String,
     pub bots: Vec<Bot>,
     pub teams: Vec<RepoTeam>,
+    pub branches: Vec<Branch>,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
@@ -167,4 +168,11 @@ pub enum RepoPermission {
     Write,
     Admin,
     Maintain,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct Branch {
+    pub name: String,
+    pub ci_checks: Vec<String>,
 }
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -642,6 +642,8 @@ pub(crate) struct Repo {
     pub description: String,
     pub bots: Vec<Bot>,
     pub access: RepoAccess,
+    #[serde(default)]
+    pub branch: Vec<Branch>,
 }
 
 impl Repo {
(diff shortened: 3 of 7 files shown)


Mark-Simulacrum:
> Require a pull request before merging

Seems OK.

> Require 1 approval (no exceptions for admins or any specific teams or people)

This is *mostly* fine, though can be annoying for some of our infrastructure, particularly when things are down. I think it's something we probably want to think about a little deeper; I think the default is right but we should probably consider allowing opt-out on a per-repo basis, even if otherwise managing through sync-team.

I'll also note that we currently have more custom settings on rust-lang/rust to allow release infrastructure to force push branches and such, so if we want support for managing those through here that'll need more configuration.

> Dismiss stale pull request approvals when new commits are pushed

This is typically more annoying than helpful. By analogy, after bors r+ on rust-lang/rust, we typically are OK with re-approvals after rebasing etc, so I think it makes sense to not automatically clear reviews elsewhere too. It's definitely a real pain with e.g. blog posts.

rylev:
> if we want support for managing those through here that'll need more configuration.

I absolutely think we'll need more configuration options, but I'd personally prefer to take it slow and not try to add all of those features at once. 

> This is typically more annoying than helpful.

While I can certainly see how in many cases dismissing reviews on new commits could be annoying, I can also imagine a lot of cases where it's not. I guess it comes down to whether you trust the author to not push substantial changes after a positive review is given. Do we have an idea of how often this setting tends to be set in rust-lang repos?

Mark-Simulacrum:
I think mostly never (all of that infra is hand managed today); the author shouldn't have write permissions if they're not trusted to make that call IMO. (In any case an approving review is required so it's not like major changes would be requested this way).

I wouldn't necessarily mind us pushing for this harder rule but I think it's not something we should enable at the same time as adding automation, since it'll slow down adoption which we want to be easy (ideally no reason to say no).

Mark-Simulacrum:
Yeah, totally happy to merge once we arrive at an agreed set of defaults - no need to add all options at once.

rylev:
Ok - not dismissing a review if someone pushes new commits is the default behavior. I don't think we need to make any changes to the team repo for that to be the case for repos controlled through this mechanism. The only thing we'll need to do here is make sure the documentation is clear, but I've not yet added that.

Mark-Simulacrum (approved):

Mark-Simulacrum:
Ah, right, I forgot that this is just adding the user facing side :)

Approving the PR, will let you merge when ready.

PR #805, opened 2022-08-19
Add more crates.io related zulip IDs

carols10cents:
The crates.io team is slowly working on migrating to Zulip for our official communications. 

I want to be able to point to this repo to say "these are the IDs that should have access to the private crates.io streams on zulip", so I'm adding the Zulip IDs that I can find. This includes people on the crates.io team and the crates.io on-call team.

To find your Zulip ID, go to "Personal Settings" -> "Profile" and click the "Preview profile" button, you should see "User ID" there.

## People whose Zulip IDs I found - Please check and comment if yours is incorrect!

- [x] @Turbo87 
- [x] @jonas-schievink 
- [x] @jtgeibel 
- [x] @listochkin
- [x] @locks
- [x] @skade 
- [x] @smarnach
- [x] @JohnTitor
- [x] @pichfl
- [x] @justahero
- [x] @Xylakant

## People on the teams who already have their Zulip ID in their teams profile (no action needed)

- me
- Pietro

### DIFF
--- a/people/Turbo87.toml
+++ b/people/Turbo87.toml
@@ -3,3 +3,4 @@ github = "Turbo87"
 github-id = 141300
 email = "tobias.bieniek@gmail.com"
 discord-id = 313625269405745155
+zulip-id = 282338
--- a/people/locks.toml
+++ b/people/locks.toml
@@ -3,3 +3,4 @@ github = "locks"
 github-id = 32344
 email = "rokusu@gmail.com"
 discord-id = 83124738867597312
+zulip-id = 459395
--- a/people/JohnTitor.toml
+++ b/people/JohnTitor.toml
@@ -3,6 +3,7 @@ github = "JohnTitor"
 github-id = 25030997
 email = "huyuumi.dev@gmail.com"
 discord-id = 362122860673892353
+zulip-id = 217081
 
 [permissions]
 bors.semverver.review = true
--- a/people/Xylakant.toml
+++ b/people/Xylakant.toml
@@ -1,3 +1,4 @@
 name = 'Felix Gilcher'
 github = 'Xylakant'
 github-id = 337823
+zulip-id = 534379
--- a/people/jonas-schievink.toml
+++ b/people/jonas-schievink.toml
@@ -3,3 +3,4 @@ github = "jonas-schievink"
 github-id = 1786438
 email = "jonasschievink@gmail.com"
 irc = "jschievink"
+zulip-id = 211727
--- a/people/jtgeibel.toml
+++ b/people/jtgeibel.toml
@@ -3,3 +3,4 @@ github = "jtgeibel"
 github-id = 22186
 email = "jtgeibel@gmail.com"
 discord-id = 451522714101088258
+zulip-id = 257369
--- a/people/justahero.toml
+++ b/people/justahero.toml
@@ -1,3 +1,4 @@
 name = 'Sebastian Ziebell'
 github = 'justahero'
 github-id = 1305185
+zulip-id = 529812
--- a/people/listochkin.toml
+++ b/people/listochkin.toml
@@ -1,3 +1,4 @@
 name = 'Андрей Листочкин (Andrei Listochkin)'
 github = 'listochkin'
 github-id = 405222
+zulip-id = 499958
--- a/people/pichfl.toml
+++ b/people/pichfl.toml
@@ -3,3 +3,4 @@ github = 'pichfl'
 github-id = 194641
 email = 'fp@ylk.gd'
 discord-id = 412702990072283138
+zulip-id = 527401
--- a/people/skade.toml
+++ b/people/skade.toml
@@ -3,3 +3,4 @@ github = "skade"
 github-id = 47542
 email = "flo@andersground.net"
 discord-id = 433394056073183253
+zulip-id = 215333
--- a/people/smarnach.toml
+++ b/people/smarnach.toml
@@ -3,3 +3,4 @@ github = "smarnach"
 github-id = 249196
 email = "sven@marnach.net"
 discord-id = 491002025950183434
+zulip-id = 219929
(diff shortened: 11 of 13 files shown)


Mark-Simulacrum:
> I want to be able to point to this repo to say "these are the IDs that should have access to the private crates.io streams on zulip", so I'm adding the Zulip IDs that I can find. This includes people on the crates.io team and the crates.io on-call team.

Note that we don't have automation for this currently -- we can adjust groups on Zulip automatically but I don't think those help with private stream membership.

people/Turbo87.toml
```diff
@@ -3,3 +3,4 @@ github = "Turbo87"
 github-id = 141300
 email = "tobias.bieniek@gmail.com"
 discord-id = 313625269405745155
+zulip-id = 282338
```
Turbo87, line 6: I don't seem to have write access here so I can't check the checkbox, but I can confirm that this ID is correct :)

weihanglo:
Sorry to ask in this PR. Just wonder what can we do with the zulip-id linking to us?

JohnTitor:
> I see multiple Zulip accounts that could be yours, which one is correct?

Sorry for the trouble! That's because I disabled an old account in the past, `217081` is my current user ID.

smarnach (approved):
My Zulip ID is correct, but I don't have permission to check the box for my name.

carols10cents:
> Note that we don't have automation for this currently -- we can adjust groups on Zulip automatically but I don't think those help with private stream membership.

Yes, I did not mean to imply that I thought there was automation, I just wanted to be able to say "hey Josh, add the zulip IDs from the repo to the private stream for me, we've checked they're correct".

carols10cents:
> Sorry to ask in this PR. Just wonder what can we do with the zulip-id linking to us?

I don't think much right now, it's just a way to verify which accounts actually belong to the team members.

carols10cents:
> Sorry for the trouble! That's because I disabled an old account in the past, `217081` is my current user ID.

No worries! I've added 217081 to your account. Thank you!

rylev:
> Note that we don't have automation for this currently 

This is half true. If you add the `[[zulip-groups]]` options to the team (see [here](https://github.com/rust-lang/team/blob/3c7b5a8a765c33a49f7562d9c16c56e59d73fc1b/teams/infra.toml#L53-L54) for an example), then a Zulip user group will be automatically synced. Then someone just needs to manually add that Zulip user group to the private chat (and not each individual user).

rylev:
Seems to work for me - or perhaps I'm misunderstanding what you mean exactly. 

image

Mark-Simulacrum:
Oh, I kept starting with an @. Yes, seems like it works. Thanks!

jtgeibel:
Thanks @carols10cents! I've confirmed my ID is correct.

jonas-schievink:
Mine is correct as well

listochkin:
Mine is correct, can't tick the box.

pichfl:
My id is 527401, sorry this took me a while to figure out.

carols10cents:
Welp I tried to add zulip groups for crates.io and on-call, but CI fails because @Xylakant and @justahero don't have zulip IDs. 

This is a practical blocker for the crates.io team moving everything over to Zulip; I don't want to split on-call work between two chat platforms.

I have sent an email to the on-call team at Ferrous and haven't received any response, positive or negative.

@Xylakant and @justahero Are you two still on the on-call team? Are you on the rust-lang zulip? Are you willing to be? If so, could you please send your zulip IDs as requested in this PR? Thanks!

rylev:
While the most straightforward solution is to get the last two remaining Zulip IDs, we can work around this in several ways:
* Add a feature to the team repo that allows you to specifically specify members you'd like to exclude from a Zulip user group. I prefer this over turning off the Zulip ID check. 
* Remove the Zulip group annotation for the on-call team and, at least temporarily, create the team manually on Zulip (many teams are still manually controlled as we haven't fully moved over to full Zulip automation).

people/locks.toml
```diff
@@ -3,3 +3,4 @@ github = "locks"
 github-id = 32344
 email = "rokusu@gmail.com"
 discord-id = 83124738867597312
+zulip-id = 459395
```
locks, line 6: This is correct! I can't check it off either.

carols10cents:
> While the most straightforward solution is to get the last two remaining Zulip IDs, we can work around this in several ways:

Yes, there are definitely technical solutions to this-- I was using this as a prompt for the social problem of making sure the on-call team is on board with this change ;)

justahero:
Hello, my Zulip id is `235105`.

rylev:
@carols10cents in the interest of getting this merged would you be ok using the `excluded_people` feature added in #833 to exclude @Xylakant from the Zulip user group? Also, @skade's Zulip ID looks correct (or at least it is an account they've used in the past), and so I think it would be fine to merge without their explicit 👍.

pietroalbini:
Felix/Xylakant's Zulip ID is 534379.

carols10cents:
Thank you @pietroalbini ! I think this is ready to merge now!

carols10cents:
Ok, I captured an instance of this happening on chat.zulip.org and I had Firefox's Network tab open, so I exported it as `.har`. Only problem was I had it and the tab open for about 24 hours before the problem... I was going to try to pare down the logs, but it's in JSON, and I don't have much time right now to figure out the right `jq` filter to get the relevant info out, so.... [here you go](https://drive.google.com/file/d/1Ac1oU6_lAAMmOvI2EgfmF7RiPhLzQPkE/view?usp=sharing). It's a 383MB `.tar` file, expands to a 715MB `.har` file. 

The different thing about this failed reload is that the loading images *did* load, as opposed to the screenshot of the beginning of this issue where the images were broken. 

The relevant requests will be near the end of the file.. I did perform a successful manual reload after the failed reload.

JohnTitor:
@Mark-Simulacrum Could you merge this? Thanks!

Mark-Simulacrum (approved):

PR #806, opened 2022-08-20
move solson to Miri alumni

RalfJung:
We have not heard from @solson in years, so this just reflects reality I think. I don't even know how to reach them... @solson if you are still around and reading this, please let me know if you have any objections. :)

@oli-obk any objections?

### DIFF
--- a/teams/miri.toml
+++ b/teams/miri.toml
@@ -3,7 +3,8 @@ subteam-of = "compiler"
 
 [people]
 leads = ["RalfJung", "oli-obk"]
-members = ["RalfJung", "oli-obk", "solson"]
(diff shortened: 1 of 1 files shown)


solson:
I've been around, following a lot of Miri developments (and rustc generally), but unfortunately haven't had the time or energy to contribute. I can be reached on here, by my listed email, on {users,internals}.rust-lang.org, and on the Rust Zulip, I think. I might be forgetting one.

RalfJung:
Ah, hi there Scott. :)  Thanks for all your work in getting Miri started!

rylev (approved):
♥

PR #809, opened 2022-08-23
zulip-id for Cargo Team members

weihanglo:
Check if it is correct. Thanks!

- [x] @Eh2406 
- [x] @ehuss 
- [x] @weihanglo


Eh2406:
It looks correct to me!

rylev (approved):

rylev:
@weihanglo as I mentioned [here](https://github.com/rust-lang/team/pull/805#issuecomment-1222241520), you may want to added `[[zulip-groups]]` annotation to the Cargo team toml configuration which will ensure that `T-cargo` on Zulip is always properly populated.

PR #810, opened 2022-08-24
Add zulip-groups annotation for Cargo Team

weihanglo:
This creates and syncs a `T-cargo` Zulip group, which is like

image

Refs: 

- https://github.com/rust-lang/team/pull/809#issuecomment-1225514064
- [TOML schema reference](https://github.com/rust-lang/team/blob/master/docs/toml-schema.md)

cc @rust-lang/cargo

### DIFF
--- a/teams/cargo.toml
+++ b/teams/cargo.toml
@@ -37,3 +37,6 @@ zulip-stream = "t-cargo"
 
 [[lists]]
 address = "cargo@rust-lang.org"
+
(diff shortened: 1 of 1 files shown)


rylev (approved):
Just a note that the `T-cargo` user group already exists on Zulip, but it is out of date. This change will make the `T-cargo` user group consistent with the members of Cargo listed here.

PR #812, opened 2022-08-26
Expose dismiss stale

rylev:
This change adds one additional configuration option to protected branches on a repos: dismissing stale reviews which default to `false`. 

Additionally, this PR adds some documentation about the options available for repo configuration.


docs/toml-schema.md
```diff
@@ -151,3 +151,36 @@ bors.some-repo.review = true
 # be set if `review` is also set.
 bors.some-repo.try = true
 ```
+
+## Repos
+
+Repos are configured by creating a file in the `repos` folder.
```
Mark-Simulacrum, line 157: Does the filename matter? I think either way we should suggest a pattern - maybe org-name if we don't want subdirectories.

rylev: File names don't matter at this point. `org-name` might get confusing since the org and the name can include `-`. Perhaps `org#name` or is that too odd?

rylev: @Mark-Simulacrum I implemented support for single-level directories.

PR #818, opened 2022-08-29
Add compiler team Zulip group

rylev:
This adds a Zulip user group for the compiler team. `T-compiler` already exists on Zulip, so the group will not be created, but there will be several changes in membership of that group:

The following people will no longer be on the `T-compiler` Zulip group:
* @arielb1
* @cramertj
* @varkor

Additionally, @cjgillot will be added to the user group.

Also, @michaelwoerister, @eddyb, @davidtwco, and @nagisa - please note I added your Zulip ID. 

r? @wesleywiser @pnkfelix

### DIFF
--- a/people/cjgillot.toml
+++ b/people/cjgillot.toml
@@ -2,3 +2,4 @@ name = 'Camille Gillot'
 github = 'cjgillot'
 github-id = 1822483
 email = "gillot.camille@gmail.com"
+zulip-id = 248906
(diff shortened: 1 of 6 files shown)


wesleywiser (approved):
Thanks @rylev! It's great to have this automated 🙂

Mark-Simulacrum (approved):

PR #819, opened 2022-08-29
improve readability, by adding pauses, and reducing overlong sentence

tshepang:


rylev (changes requested):
Thanks! The change looks good, but can you please format the lines so they are all roughly of the same length?

rylev (approved):

PR #820, opened 2022-08-29
Add thomcc to project-portable-simd

thomcc:
I've been pretty active in this group from the beginning, although never was on this list. Someone pointed out that I should be, so I ran it by @workingjubilee who agreed.


rylev (approved):
It would be good to get confirmation @workingjubilee has seen this but looks great!

workingjubilee:
This is absolutely correct, yes!

PR #823, opened 2022-09-02
Add Exploit Mitigations Project Group

rcvalle:
This PR adds the Exploit Mitigations Project Group (see rust-lang/compiler-team#545).

### DIFF
--- a/teams/project-exploit-mitigations.toml
+++ b/teams/project-exploit-mitigations.toml
@@ -1,0 +1,24 @@
+name = "project-exploit-mitigations"
+kind = "project-group"
+subteam-of = "compiler"
+
+[people]
+leads = [
+    "rcvalle",
+]
+members = [
+    "cuviper",
+    "rcvalle",
+]
+
+[website]
+name = "Exploit Mitigations Project Group"
+description = "Maintaining and improving the existing, implementing, and researching new exploit mitigations for the Rust compiler"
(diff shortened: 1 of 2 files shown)


wesleywiser (approved):
Thanks @rcvalle!

teams/project-exploit-mitigations.toml
```diff
@@ -0,0 +1,21 @@
...
+]
+
+[website]
+name = "Exploit Mitigations Project Group"
+description = "Maintaining and improving the existing, implementing, and researching new exploit mitigations for the Rust compiler"
+repo = "https://github.com/rust-lang/project-exploit-mitigations"
+zulip-stream = "project-exploit-mitigations"
+
+[[github]]
+orgs = ["rust-lang"]
```
rylev, line 21: Would it make sense to also add a `[[zulip-groups]]` annotation ([like this](https://github.com/rust-lang/team/blob/7a9db9bcd1eddfd911dc17095de40002d9532a1c/teams/infra.toml#L53-L54)) so that this group can be pinged on Zulip?

rylev (approved):

teams/project-exploit-mitigations.toml
rcvalle: Yes, thanks for pointing it out! Done.

rylev (approved):

PR #825, opened 2022-09-02
Zulip meetings groups for compiler and types teams

rylev:
Adds Zulip groups for T-compiler/meeting and T-types/meetings (note: the difference in the use of "meeting" with and without an 's' at the end is a reflection of the status quo).

This adds Zulip IDs for @Xanewok and @arielb1.

When this syncs, the Zulip groups will remain unchanged with the exception that @cjgillot will be added to T-compiler/meeting (since they are a member of the compiler team). 

The extra members listed are a reflection of the status quo. While it would be fine to keep the status quo as is, it might be nice to use this opportunity to do some housekeeping. 

r? @pnkfelix @wesleywiser @nikomatsakis @jackh726


wesleywiser (approved):

Mark-Simulacrum:
Hm, I sort of wonder if maintaining these via team repo is the right tool. I guess it seems okay - we should probably make that clear in the description, or maybe even have sync-team send a message (privately) when making edits here.

Mark-Simulacrum (approved):

rylev:
@Mark-Simulacrum it definitely feels like the right place for me. Could you talk more about why this wouldn't be the right place? Sending a private message when this information changes, seems like a really good idea!

Mark-Simulacrum:
I mostly mean that my impression is the meeting groups are intended for "random" users (e.g., if we look at the lists here they're not strictly expansions into the contributor group or similar) to add themselves, so keeping them in sync via team repo could represent fairly high overhead. That said, they don't change that much and seem to not include people who aren't already otherwise noted somewhere in rust-lang/team, so it seems okay.

rylev:
That's fair though I think the overhead of figuring out who has access to manually edit Zulip user groups can be much higher overhead for some folks than opening a PR (especially when we lower the threshold for merging these sorts of non-critical changes).

Mark-Simulacrum:
Yeah, I think it's anyone in the group? But that isn't a trivial overhead, I agree.

PR #830, opened 2022-09-05
Delay running the sync-team tool so that GitHub pages has time to fully deploy

rylev:
The SLA for how much time it takes for GitHub pages to deploy is 10 minutes. However, hopefully it will be faster than that in reality. 

Alternatively, we could try to modify the deploy.rs script to figure out if the deploy has succeeded (perhaps using [this API endpoint](https://docs.github.com/en/rest/pages#get-latest-pages-build)). However, this comes with its own set of issues including the need for a GitHub access key and the fact that it's simply more code that could break.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,6 +48,7 @@ jobs:
 
       - name: Start the synchronization tool
         run: |
+          sleep 5 # give GitHub pages enough time to fully deploy
           aws --region us-west-1 lambda invoke --function-name start-sync-team output.json
           cat output.json | python3 -m json.tool
         env:


Mark-Simulacrum:
I added a similar (longer) delay in https://github.com/rust-lang/team/pull/813/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R62, so in general this seems OK.

I think if we want to avoid this being an issue, the right solution is probably to move the static file hosting here to a dedicated s3 bucket + cloudfront distribution with short or zero cache TTLs. Might be a little more reliable, too; I know we've run into issues with caching on GH Pages in the past (with the blog, though, not this, but presumably the same problems can happen here).

pietroalbini:
Let's use the implementation in https://github.com/rust-lang/team/pull/813

PR #808, opened 2022-08-23
Clean up WG-CLI's members

epage:
- @spacekookie: through DM said they were ok to be removed
- @yoshuawuyts: through DM said they were ok to be removed
- @pksunkara: has been acting in the role of a member for a while
- @epage (myself): has been acting as the de facto leader for a while
- deg4uss3r: confirmed over email their continued involvement
- matthiasbeyer: confirmed over email their continued involvement
- Dylan-DPC: conmfirmed over discored their continued involvement

PR #831, opened 2022-09-05
Update My Profile Email and ZulipID

deg4uss3r:
Add ZulipID and update email in profile and WG-CLI

### DIFF
--- a/teams/wg-cli.toml
+++ b/teams/wg-cli.toml
@@ -29,6 +29,6 @@ extra-emails = [
     "mail@r0tty.org",
     "msarfati@protonmail.com",
     "mail+rust-wg-cli@pksunkara.com",
-    "Ricky@Hosfelt.io",
+    "Ricky@Hosfe.lt",
     "mail@beyermatthias.de",
 ]
--- a/people/deg4uss3r.toml
+++ b/people/deg4uss3r.toml
@@ -1,4 +1,5 @@
 name = 'Ricky'
 github = 'deg4uss3r'
 github-id = 15351059
-email = "ricky@hosfelt.io"
+email = "ricky@hosfe.lt"
+zulip-id = 473301


rylev (approved):

teams/wg-cli.toml
```diff
@@ -29,6 +29,6 @@ extra-emails = [
     "mail@r0tty.org",
     "msarfati@protonmail.com",
     "mail+rust-wg-cli@pksunkara.com",
-    "Ricky@Hosfelt.io",
```
rylev, line 32: @deg4uss3r I believe this can just be deleted. You are on the team and this address automatically goes to email associated with each of the members of the team.

rylev (changes requested):

teams/wg-cli.toml
deg4uss3r: Thanks @rylev! Can we leave it for posterity?

I am one of the active members and would like people to be able to contact me if necessary. For example if they are looking for something related to the Working Group and I'm not on Discord/Zulip active. 

Unless we're actively clearing all emails from all groups I'd like to make sure I'm there :)

rylev: @deg4uss3r I think there might be a miss understanding. Your email is in your personal toml file ([people/deg4uss3r.toml](https://github.com/rust-lang/team/pull/831/files#diff-80b065898e39ccfc76f247a358dabd6d942fe1da522e94d783a65189117e24e8)) so if someone wants to get your email they can easily find it there. As this currently stands your email is being repeated twice. I don't think this repetition is actually helpful.

deg4uss3r: There's two options here. 

1. Remove just my email now in this MR
1. Opening a new PR (or doing this in #808) that is more focused to clean out all of the extra emails. 

I prefer the latter.

rylev: Ok. I'm not sure I agree, but this is not worth blocking merging this. 😊

rylev (approved):

PR #832, opened 2022-09-07
Add Zulip config for async working group

rylev:
This adds the `WG-async` and `WG-async-sprint-planning` Zulip user groups so that they are automatically synced.

The `WG-async` is set to just the members of the working group. This would remove @hawkw, @csmoe, @gilescope, @Aaron1011, @cramertj, @betamos, @LucioFranco, @Nemo157, @nellshamrell, @Stupremee, and @withoutboats as these folks are not members of the working group. If the #WG-async user group is being used for more than just members, we can easily add folks back using the `extra-members` feature.

The `WG-async-sprint-planning` is set to the members of the working group plus @hawkw whose info has been added to the repo. This also adds @yoshuawuyts and @taiki-e to this user group as they had not yet been added.

r? @nikomatsakis @tmandry

### DIFF
--- a/teams/wg-async.toml
+++ b/teams/wg-async.toml
@@ -42,3 +42,12 @@ name = "Async working group"
 description = "Pursuing core language and library support for async-await"
 repo = "https://github.com/rust-lang/wg-async"
 zulip-stream = "wg-async"
+
+[[zulip-groups]]
+name = "WG-async"
+
+[[zulip-groups]]
+name = "WG-async-sprint-planning"
+extra-people = [
+  "hawkw"
+]
--- a/teams/wg-async.toml
+++ b/teams/wg-async.toml
@@ -48,6 +48,3 @@ name = "WG-async"
 
 [[zulip-groups]]
 name = "WG-async-sprint-planning"
-extra-people = [
-  "hawkw"
-]
--- a/people/hawkw.toml
+++ b/people/hawkw.toml
@@ -1,0 +1,5 @@
+name = "Eliza Weisman"
+email = 'eliza@buoyant.io'
+github = 'hawkw'
+github-id = 2796466
+zulip-id = 307286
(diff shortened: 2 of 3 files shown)


teams/wg-async.toml
```diff
@@ -42,3 +42,12 @@ name = "Async working group"
...
 repo = "https://github.com/rust-lang/wg-async"
 zulip-stream = "wg-async"
+
+[[zulip-groups]]
+name = "WG-async"
+
+[[zulip-groups]]
+name = "WG-async-sprint-planning"
+extra-people = [
+  "hawkw"
```
tmandry, line 52: I don't think I remember @hawkw coming to meetings, so I'm not sure she wants to be in this list

teams/wg-async.toml
```diff
@@ -42,3 +42,12 @@ name = "Async working group"
 description = "Pursuing core language and library support for async-await"
 repo = "https://github.com/rust-lang/wg-async"
 zulip-stream = "wg-async"
+
+[[zulip-groups]]
+name = "WG-async"
+
+[[zulip-groups]]
+name = "WG-async-sprint-planning"
```
tmandry, line 50: The original point of this alias was to be a subset of the members who lead a focus area or otherwise participate in sprint planning meetings. It can probably be revisited. I can follow up with a PR to make a sub-team.

tmandry (approved):
Oh, great – I've been wanting a way to sync these for awhile!

teams/wg-async.toml
rylev: Removed for now. We can always re-add later.

teams/wg-async.toml
rylev: We might be able to use https://github.com/rust-lang/team/pull/833 to accomplish this.

tmandry (approved):
Thanks!

Mark-Simulacrum (approved):

PR #836, opened 2022-09-09
Add a wg-prioritization/alerts Zulip user group

rylev:
This adds automatic syncing of the `wg-prioritization/alerts' Zulip user group with the membership of the wg-prioritization working group.

The user group is not currently very well synced with the working group. When this merges the following will happen:

These folks will join the user group: @camelid, @Dylan-DPC, @hkmatsumoto, @inquisitivecrystal, @lcnr, @pnkfelix, @TaKO8Ki.

The following folks will be removed (both of whom are not in the team repo, and so I don't know their GitHub handles): Xu Ran, Chris Dolby. 

r? @apiraino @wesleywiser

### DIFF
--- a/teams/wg-prioritization.toml
+++ b/teams/wg-prioritization.toml
@@ -46,3 +46,6 @@ orgs = ["rust-lang"]
 
 [[zulip-groups]]
 name = "WG-prioritization"
+
+[[zulip-groups]]
+name = "WG-prioritization/alerts"
--- a/teams/wg-prioritization.toml
+++ b/teams/wg-prioritization.toml
@@ -49,3 +49,12 @@ name = "WG-prioritization"
 
 [[zulip-groups]]
 name = "WG-prioritization/alerts"
+excluded-people = [
+    "lcnr",
+    "Dylan-DPC",
+    "camelid",
+    "pnkfelix",
+    "hkmatsumoto",
+    "inquisitivecrystal",
+    "TaKO8Ki",
+]


lcnr:
I do not want to be part of the `alerts` zulip group. We intentionally split `@*WG-prioritization*` and `@*WG-prioritization/alerts*` so that people can be part of `WG-prioritization` without getting pinged for new threads in the alerts channel.

rylev:
This is good context. We can easily add your name to the `excluded-people` field which removes you from that Zulip group. I'll wait for others to chime in on if they'd like to be excluded before committing that.

Dylan-DPC:
Would like to be excluded from the alerts as well for the time being

spastorino:
@rylev I'd also want to be left out of `@*WG-prioritization/alerts*` or well maybe even better I should just be removed from the whole working group as I'm not doing anything since a while.

Anyway, in general, I'd say that we can just copy what we already have on Zulip. There's some people that belongs to the Zulip `@*WG-prioritization*` group and a subset of that people belongs to `@*WG-prioritization/alerts*`. That allows to ping the team for general discussion and to ping for alert on every single issue that ask for notifications.

In any case, cc @apiraino, it may be also good to also take some time and cleanup the list of members.

apiraino:
Why do we want to map the Zulip `t-compiler/wg-prioritization/alerts` group in the team repository? What happens when this Zulip group is mapped in this file?

As far as I know, `t-compiler/wg-prioritization/alerts` was just an _opt-in_ subgroup to receive notifications about issues to be prioritized (the actual Zulip group is `t-compiler/wg-prioritization`).

Besides, @spastorino is right - we need to update the wg-prioritization roster to remove stale members (something I will take care of).

rylev:
@apiraino we want to be able to control all zulip user groups from this repo (or at the very least some other well defined place). The current process of "just know that you have to ask someone with permissions" is not sustainable.

camelid:
I would also like to be removed from the new `wg-prioritization/alerts`, though I would like to remain part of `wg-prioritization`.

rylev:
@Mark-Simulacrum this should be ready to go. I've added more folks to the excluded list such that it should not actually change who is in the user group.

Mark-Simulacrum (approved):

PR #842, opened 2022-09-10
Add myself to wg-ffi-unwind

nbdd0121:
I have been working on unwinding related stuff for quite a while now but not formally in the WG.

@BatmanAoD

Issue #826, opened 2022-09-02
Add remaining Zulip user groups


rylev:
The following are the remaining Zulip user groups that are not yet synced between the team repo and the Zulip instance. 

Once these are added, the Zulip user groups will be completely managed by the team repo. 

* [ ] 'WG-rls2.0'
* [ ] 'WG-llvm'
* [ ] 'WG-traits'
* [ ] 'project-custom-linter'
* [ ] 'WG-polonius'
* [ ] 'WG-rustc-dev-guide'
* [ ] 'WG-const-eval'
* [ ] 'WG-async-sprint-planning'
* [ ] 'WG-pipelining'
* [ ] 'WG-mir-opt'
* [ ] 'mods'
* [ ] 'WG-governance'
* [ ] 'WG-unsafe-code-guidelines'
* [ ] 'WG-ffi-unwind'
* [ ] 'WG-polymorphization'
* [ ] 'WG-self-profile'
* [ ] 'WG-async'
* [ ] 'WG-rfc-2229'
* [ ] 'WG-diagnostics'
* [ ] 'WG-prioritization/alerts'
* [ ] 'project-const-generics'
* [ ] 'T-compiler/WG-meta'
* [ ] 'WG-profile-guided-optimization'
* [ ] 'WG-nll'
* [ ] 'wg-incr-comp'
* [ ] 'T-community/l10n'
* [ ] 'foundation'

apiraino:
@rylev some thoughts while scrolling the list:

- I believe **WG-Traits** is not active anymore (note to self: I think it can be marked as such [in this page](https://rust-lang.github.io/compiler-team/working-groups/traits/)). Am I correct @jackh726 ?
- **WG-rls2.0** has been renamed to `t-compiler/rust-analyzer` (on [Zulip](https://rust-lang.zulipchat.com/#streams/185405/t-compiler.2Frust-analyzer)). I suspect they want to propagate this new name everywhere.
- **WG-rfc-2229**, **WG-nll**, **WG-pipelining**,  are all inactive by looking at [this list](https://rust-lang.github.io/compiler-team/working-groups/)

hth!

jackh726:
wg-traits is not active anymore

NLL working group needs to be archived, so not worth adding new zulip group

PR #845, opened 2022-09-12
`wg-ffi-unwind` zulip sync

BatmanAoD:
#826 for `wg-ffi-unwind`; also includes #842

### DIFF
--- a/teams/wg-ffi-unwind.toml
+++ b/teams/wg-ffi-unwind.toml
@@ -10,6 +10,7 @@ members = [
     "CAD97",
     "acfoltzer",
     "bjorn3",
+    "chorman0773",
     "cratelyn",
     "gnzlbg",
     "nbdd0121",
--- a/teams/wg-ffi-unwind.toml
+++ b/teams/wg-ffi-unwind.toml
@@ -10,6 +10,7 @@ members = [
     "CAD97",
     "acfoltzer",
     "bjorn3",
+    "cratelyn",
     "gnzlbg",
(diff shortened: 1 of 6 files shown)


teams/wg-ffi-unwind.toml
```diff
@@ -20,3 +26,9 @@ zulip-stream = "project-ffi-unwind"
 
 [[github]]
 orgs = ["rust-lang"]
+
+[[zulip-groups]]
+name = "WG-ffi-unwind"
+
+# Katelynn Martin has two Zulip accounts.
```
Mark-Simulacrum, line 33:
```suggestion
# Katelyn Martin has two Zulip accounts.
```

Interesting. Are both of these active on an ongoing basis?

BatmanAoD: I hadn't checked, actually, but it looks like one account was effectively unused, so I've removed that one.

rylev:
We should make everyone who is impacted by these changes aware that they are happening. 

@CAD97, @bjorn3, @chorman0773, @cratelyn, @joshtriplett, @gnzlbg, @Amanieu

rylev (approved):
This looks good to me. 

As a recap, this changes the Zulip user group by adding @chorman0773. Otherwise, the Zulip group stays the same. Of course, the official membership of the working group itself does expand considerably. 

It's fine to merge this without approval from everyone impacted as reverting these changes are low-cost and trivial.

PR #849, opened 2022-09-16
Add Nell to dev desktops

jdno:
Give @nellshamrell access to the dev desktops

cc @pietroalbini


pietroalbini (approved):

Mark-Simulacrum:
I think we probably ought to do this via https://github.com/rust-lang/team/blob/master/teams/cloud-compute.toml, just to keep it all in one place for the ad-hoc added folks.

PR #856, opened 2022-09-28
Archive wg-nll

rylev:
As a follow up to discussion in https://github.com/rust-lang/team/pull/736, this PR archives the nll working group. In that previous discussion, it was pointed out that this was an appropriate action, but we should wait until nll was fully stabilized [which it now is](https://blog.rust-lang.org/2022/08/05/nll-by-default.html). 

r? @nikomatsakis @pnkfelix


spastorino:
I think we should also archive wg-meta and wg-parallel-rustc, unsure about other wgs which I do not belong. Also, unsure what's the plan with traits now that we have types team cc @jackh726

jackh726:
I imagine we need to do here what we did for wg-traits in #760 (remove permissions but keep in place). I'm not sure if anything has changed since then. @pietroalbini 

This change looks good to me otherwise. Thanks for making this, have meant to but forgot.

rylev:
@jackh726 I believe the reason we didn't archive the types working group was because there were folks on that working group who were not on any other team, and the team repo can't handle that currently. CI is passing so I don't believe that's necessary in this case. Am I missing something?

jackh726:
Ah, I guess that's true

rylev:
> I think we should also archive wg-meta and wg-parallel-rustc, unsure about other wgs which I do not belong. Also, unsure what's the plan with traits now that we have types team cc @jackh726

I can do archive those other teams in a follow up PR.

Mark-Simulacrum (approved):

PR #858, opened 2022-09-29
Archive wg traits

rylev:
A follow up to #760. This finally archived wg-traits.

In order to do so, it allows people in the people directory if they are only referenced by archived teams. 

r? @jackh726


rylev:
@Mark-Simulacrum I don't believe there is an `alumni` key on a person's toml file. The comment you're referring to is simply saying add this person to the team's alumni list (or create it if one doesn't already exist), but that's not possible when the team itself is being archived. We could add some indication to the person's toml file that the person in question is only here for archival purposes, but I'm not sure that's actually better.

Mark-Simulacrum:
I think the main disadvantage is that `ls people | wc -l` becomes less accurate in some sense for 'current members of the project' -- that's a useful command from time to time, though it's obviously not really reflective of any concrete reality.

I think we would at minimum want this PR to automatically add the people into the alumni team (https://www.rust-lang.org/governance/teams/alumni) if they're only a member of the archived teams. But I'm OK leaving that to follow-up work, so I'm going to approve this PR for now.

Mark-Simulacrum (approved):

jackh726 (approved):

rylev:
> I think the main disadvantage is that ls people | wc -l becomes less accurate in some sense for 'current members of the project'

That's already the case though. Marker teams already include folks we don't consider "part of the project". If we have needs for useful stats about the project we should build that into the tool. Not quite as easy as using bash tools, but it's already necessary to get a truly accurate picture.

Mark-Simulacrum:
Yeah, for sure. It's not really a good metric in any case.

rylev:
Going to merge. I'll post a follow up with the desired improvements.

PR #859, opened 2022-10-04
More alumni

rylev:
A follow up to #858. This adds members of archived teams that are no longer active anywhere else to the alumni team. It also adds some more validation that the alumni team is correct (and allows us to remove some from the explicit list of alumni since they are reflected elsewhere). Finally, there is some light refactoring to remove unnecessary use of `Result`.

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -129,17 +129,12 @@ impl Data {
         self.people.values()
     }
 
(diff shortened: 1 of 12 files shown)


rylev:
@Mark-Simulacrum Just a friendly ping on this.

Mark-Simulacrum (approved):

PR #860, opened 2022-10-05
Allow comparing bots

rylev:
Needed for https://github.com/rust-lang/sync-team/pull/10


Mark-Simulacrum (approved):
Maybe worth Eq and Hash impls?

rylev:
@Mark-Simulacrum possibly but I think I'd rather wait until those are actually motivated (even if it might cause some annoying back and forth).

PR #862, opened 2022-10-08
Give cjgillot access to the cloud dev environment.

cjgillot:
My current contributions are done using a 10-yo personal laptop, which is a bit slow for running the rustc test suite.
Having access to a professional-grade environment would really help.

### DIFF
--- a/teams/cloud-compute.toml
+++ b/teams/cloud-compute.toml
@@ -15,6 +15,7 @@ members = [
     "fee1-dead",
     "Dajamante",
     "nellshamrell",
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
https://github.com/rust-lang/team/blob/master/teams/compiler.toml#L32

I think you should already have it?

cjgillot:
I must have missed the communication, I thought this was still in testing phase.
This is so great! My warmest thanks to the foundation for providing this!

PR #865, opened 2022-10-21
Move regex team next to the crate-maintainers team on the website.

m-ou-se:
This moves the regex team to appear right next to crate-maintainers on the website, to make it more obvious that crate-maintainers doesn't cover maintenance of `regex`.

### DIFF
--- a/teams/regex.toml
+++ b/teams/regex.toml
@@ -10,5 +10,6 @@ orgs = ["rust-lang"]
 
 [website]
 name = "Regex crate team"
+weight = -250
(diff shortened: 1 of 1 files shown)


rylev (approved):
Seems fine to me. @m-ou-se did you test that this indeed does what you intend it to do?

m-ou-se:
> Seems fine to me. @m-ou-se did you test that this indeed does what you intend it to do?

I tested the `weight =` feature when I added it. I didn't test this particular `weight = -250`, but it works for the other subteams on the lib page..

PR #867, opened 2022-10-21
add pnkfelix to cloud-compute team.

pnkfelix:


Mark-Simulacrum:
https://github.com/rust-lang/team/pull/862#issuecomment-1272409914

We should probably put a comment in the file or something - but this isn't necessary I would expect.

pnkfelix:
> 

Filed https://github.com/rust-lang/team/pull/868 to add such a comment

PR #868, opened 2022-10-21
ward off people who already have access

pnkfelix:
As suggested by @simulacrum in https://github.com/rust-lang/team/pull/867#issuecomment-1286909457, add comment warding off people who already have access.

### DIFF
--- a/teams/cloud-compute.toml
+++ b/teams/cloud-compute.toml
@@ -3,6 +3,9 @@ kind = "marker-team"
 
 [people]
 leads = []
+# Note: if you are a member of a Rust team that has permissions.dev-desktop set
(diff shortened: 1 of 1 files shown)


rylev (commented):
Perhaps we should create a validation step that checks for this? That way, this will at least get caught in CI.

pnkfelix:
I don’t think we should *prevent* people from joining the cloud-compute team just due to being on another team. Eg they might well plan *leave* that other team and would prefer to have continuous access to the cloud desktops during their transition off the team

rylev (approved):
Ok, we can always add such a check later if we change our minds.

PR #875, opened 2022-10-24
Add pkg-config repo

rylev:
This adds configuration for the [`pkg-config` repo](https://github.com/rust-lang/pkg-config-rs). This largely mirrors the current config on GitHub with addition of granting the rest of the crate-maintainers team maintainer rights.  

This will also remove @sdroege's maintainer rights which might not be desired. I'll have to add support for adding individuals to repos if that's a requirement here. 

r? @m-ou-se @joshtriplett

PR #861, opened 2022-10-05
Add new crate maintainers team.

m-ou-se:
Hey @Amanieu @JohnTitor @KodrAus @joshtriplett @the8472 @thomcc @workingjubilee, does this look right?

PR #879, opened 2022-10-26
More crates-maintainers repos

rylev:
r? @m-ou-se 

cc @rust-lang/crate-maintainers 

Note: some of these repos used to give access to `libs` and/or `libs-contributors` . If anyone is missing access, it's just a PR away from giving it to them.

### DIFF
--- a/repos/rust-lang/cc-rs.toml
+++ b/repos/rust-lang/cc-rs.toml
@@ -1,0 +1,10 @@
+org = 'rust-lang'
+name = 'cc-rs'
+description = 'Rust library for build scripts to compile C/C++ code into a Rust library'
+bots = []
+
+[access.teams]
+crate-maintainers = 'write'
+
+[[branch]]
+name = 'main'
--- a/repos/rust-lang/cc-rs.toml
+++ b/repos/rust-lang/cc-rs.toml
@@ -4,7 +4,7 @@ description = 'Rust library for build scripts to compile C/C++ code into a Rust 
 bots = []
 
 [access.teams]
-crate-maintainers = 'write'
+crate-maintainers = 'maintain'
 
 [[branch]]
 name = 'main'
--- a/repos/rust-lang/glob.toml
+++ b/repos/rust-lang/glob.toml
@@ -4,4 +4,4 @@ description = 'Support for matching file paths against Unix shell style patterns
 bots = []
 
 [access.teams]
-libs = 'write'
+crate-maintainers = 'maintain'
--- a/repos/rust-lang/glob.toml
+++ b/repos/rust-lang/glob.toml
@@ -1,0 +1,7 @@
+org = 'rust-lang'
+name = 'glob'
+description = 'Support for matching file paths against Unix shell style patterns.'
+bots = []
+
+[access.teams]
+libs = 'write'
--- a/src/main.rs
+++ b/src/main.rs
@@ -195,7 +195,7 @@ fn run() -> Result<(), Error> {
             let github = github::GitHubApi::new();
             let repo = match github.repo(&org, &name)? {
                 Some(r) => r,
-                None => failure::bail!("The repo '{org}/{name}' was not found on GitHub"),
+                None => failure::bail!("The repo '{}/{}' was not found on GitHub", org, name),
             };
             let mut teams = HashMap::new();
             let mut bots = Vec::new();
--- a/src/main.rs
+++ b/src/main.rs
@@ -167,21 +167,21 @@ fn run() -> Result<(), Error> {
             info!("written data to {}", file);
         }
         Cli::AddRepo { org, name } => {
-            #[derive(serde::Serialize)]
+            #[derive(serde::Serialize, Debug)]
             #[serde(rename_all = "kebab-case")]
             struct AccessToAdd {
                 teams: HashMap<String, String>,
             }
-            #[derive(serde::Serialize)]
+            #[derive(serde::Serialize, Debug)]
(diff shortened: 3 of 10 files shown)


src/main.rs
```diff
@@ -167,34 +167,35 @@ fn run() -> Result<(), Error> {
...
                 bots: Vec<String>,
                 access: AccessToAdd,
+                #[serde(skip_serializing_if = "Vec::is_empty")]
                 branch: Vec<BranchToAdd>,
             }
             let github = github::GitHubApi::new();
             let repo = match github.repo(&org, &name)? {
                 Some(r) => r,
-                None => failure::bail!("The repo '{org}/{name}' was not found on GitHub"),
+                None => failure::bail!("The repo '{}/{}' was not found on GitHub", org, name),
```
tshepang, line 198: curious why this changed... is an older version of Rust used

repos/rust-lang/cc-rs.toml
```diff
@@ -0,0 +1,10 @@
+org = 'rust-lang'
+name = 'cc-rs'
+description = 'Rust library for build scripts to compile C/C++ code into a Rust library'
+bots = []
+
+[access.teams]
+crate-maintainers = 'write'
```
thomcc, line 7: Is there a reason this one is `write` and the others are `maintain` (caveat: I don't know what the difference is between these).

repos/rust-lang/glob.toml
```diff
@@ -0,0 +1,7 @@
+org = 'rust-lang'
+name = 'glob'
+description = 'Support for matching file paths against Unix shell style patterns.'
+bots = []
+
+[access.teams]
+libs = 'write'
```
thomcc, line 7: Should this be crate-maintainers?

thomcc (commented):
Some comments -- perhaps these are they way they are deliberately though.

repos/rust-lang/cc-rs.toml
rylev: Nope - just a typo.

repos/rust-lang/glob.toml
rylev: Yes it should

thomcc:
Note that this is missing `pkg-config`, `flate2`, `socket2`, `getopts`, `backtrace-rs` from the list in https://github.com/rust-lang/team/pull/861#issuecomment-1286670330.

I don't have super strong opinions on these (but have already started helping out on `pkg-config-rs`).

rylev:
`pkg-config` was done in #875. The others require some added functionality I'd like to do in a follow up PR.

Mark-Simulacrum (approved):
Would like to see approval from Mara as well to validate this, but seems good to me.

thomcc:
Should this allow publishing to crates.io? I'd like to cut releases of some of these crates soon and looking at e.g. <https://crates.io/crates/cc> it still only lists rust-lang/libs and rust-lang-owner as having publish rights, not rust-lang/crates-maintainers.

Mark-Simulacrum:
I personally would like to avoid adding all crates-maintainers directly to the publishing rights, but rather allow publishing via some token that is owned by rust-lang-owner (or a similar bot account). We need to build that system still though. (This is mainly to reduce the burden of trust for adding someone to the team: pushing/merging in GitHub and a crates.io publish have pretty different blast radii).

In the meantime I'm happy to own the debt of publishing crates where necessary with relatively little latency.

thomcc:
> In the meantime I'm happy to own the debt of publishing crates where necessary with relatively little latency.

That sounds like a pretty tedious workflow if I'm being honest, especially if it's to last until some future system is built that doesn't sound like it's been designed yet.

Worth noting that one of the points of frustration people have about some of these crates has been too much time between releases.

thomcc:
FWIW, in the short term I only really want publishing rights for cc-rs and cmake-rs, since those are the crates most immediately I intend to try to improve (well, pkg-config too, but that just had a release cut). [flate2](https://github.com/rust-lang/flate2-rs/issues/321) also has users asking for a release, and I haven't had time to look over most of the others yet.

Amanieu:
In the case of compiler-builtins I often publish a release immediately after a PR is merge. This is sufficiently infrequent (less than once a week) that if I don't do it immediately I will forget to do it. In any it's not like we're at risk of running out of version numbers.

thomcc:
I don't know that I would do it that frequently, but I keep fairly odd hours a lot of the time, and do a lot of open source work pretty late at night for most folk (and over the weekend).

Given that the release has parts that happen before publish (bumping the version) and after publish (cutting a tag, compiling a changelog, publishing a github release), I would really rather not have an intermediate step there of "ask someone else to perform the publish for me" -- that sounds really annoying.

thomcc:
FWIW, https://github.com/rust-lang/team/pull/861#issuecomment-1286670330 does indicate that crates-maintainers should have publish access, if it makes any difference.

Mark-Simulacrum:
I think we should discuss more, but I would guess a simple scheme can be put together pretty quickly that automates the publish of an existing GitHub commit (for example).

It may be that we give the team access for now to most of these crates and then drop it in the future.

thomcc:
Yeah, that scheme sounds fine; I would only ever want to publish commits that are already in the repository of course.

My objection was based mainly around having to wait on someone to get things done -- longer term I don't mind having automation be responsible for it at all (especially if it's integrated with tagging, which allow some assurance what got published matches what was tagged).

(And as you might expect, as a result I'm in favor of giving access now and dropping it once the scheme is in place, but I can wait for more discussion)

PR #885, opened 2022-11-09
Remove master branch protection on compiler-builtins

Amanieu:
This crate is fairly low volume and only has a single maintainer (me). Currently I have no way of publishing a release because PRs also require an approval from someone other than the author.


Mark-Simulacrum (approved):
Approving for now to get back to previous state.

PR #887, opened 2022-11-09
add Rustc Trait System Rewrite Initiative

lcnr:
cc https://github.com/rust-lang/types-team/issues/58

I've added every one who is already a part of an existing team and mentioned that they're interested in https://github.com/rust-lang/types-team/issues/58 to the initiative for now: @BoxyUwU @cjgillot @compiler-errors @spastorino.

Not sure if opening a new repository for the initiative is too useful as it would only 2-4 documents and in my experience initiative repos are somewhat hard to discover. I don't know, maybe we discover that a repo is really useful '^^ can still make a new one at that point

@rust-lang/types for review/approval

### DIFF
--- a/teams/project-trait-system-rewrite.toml
+++ b/teams/project-trait-system-rewrite.toml
@@ -1,0 +1,24 @@
+name = "project-trait-system-rewrite"
+kind = "project-group"
+subteam-of = "types"
+
+[people]
+leads = [
+    "lcnr",
+]
+members = [
+    "BoxyUwU",
+    "cjgillot",
+    "compiler-errors",
+    "spastorino",
+    "lcnr",
+]
+
+[website]
+name = "Rustc Trait System Rewrite Initiative"
+description = "Rewriting the trait system of rustc"
+repo = "https://github.com/rust-lang/types-team"
+
+[[github]]
+orgs = ["rust-lang"]
+team-name = "initiative-trait-system-rewrite"


lcnr:
changed it to be a subteam of compiler as types is already a subteam of it and having subteams of a subteam isn't allowed.

jackh726 (dismissed):
I think rewrite -> refactor. Other than that, LGTM.

lcnr:
:+1: changed to refactor.

jackh726 (approved):

jackh726:
Ah wait, do we want a zulip stream? Do you think it'd be useful?

rylev:
Zulip streams aren't yet configured through this repo. You'll have to ask #infra on Zulip for that for now.

teams/project-trait-system-refactor.toml
```diff
@@ -0,0 +1,26 @@
+name = "project-trait-system-refactor"
+kind = "project-group"
+# Really should be types, but types is already a subteam,
+# so that's not allowed for now.
```
rylev, line 4: @pietroalbini do you know why this restriction was put in place?

rylev (approved):

teams/project-trait-system-refactor.toml
lcnr: maybe because we of the governance webpage as there's currently a 1 to 1 correspondence between the parent team and the governance page: https://www.rust-lang.org/governance/teams/compiler :thinking:

lcnr:
> Ah wait, do we want a zulip stream? Do you think it'd be useful?

Will try and experiment with project boards for this: https://github.com/orgs/rust-lang/projects/26/views/1. For that each area should then have a project board entry and a https://github.com/rust-lang/rust issue.

We could have a zulip stream with a thread for each of these issues, but I don't think that's too helpful. If we just need a place to discuss the impl of issues or general questions, staying in the #t-types stream is probably better as its easier to discover. Don't expect that we're going to have too many zulip topics for this initiative so using the #t-types stream should not drown out other work.

PR #893, opened 2022-11-18
crates.io: Switch from Discord to Zulip

Turbo87:
We're already using `[[zulip-groups]]` in the file and have moved our private channels over. This PR should update the website to also link to Zulip instead of Discord.


rylev (approved):
❤️

PR #897, opened 2022-11-29
Move CraftSpider to rustdoc team alumnis

GuillaumeGomez:
@CraftSpider hasn't been very active in the last months. After discussing it with them and the team, they agreed to being moved to alumnis for the time being. Thanks **a lot** for all your work. We'd love to have you back on board in the future!

cc @rust-lang/rustdoc


rylev (approved):
Thanks ♥

PR #898, opened 2022-12-01
Let t-miri use perf

saethlin:
@jyn514 keeps telling me to submit this
cc @RalfJung @oli-obk 

The Miri team primarily works on a separate repo, but parts of the interpreter live in rust-lang/rust. So it seems like anyone on the team has reason to do perf ~and crater~ runs, and should be trusted to do so responsibly. Of course, this change only affects me currently.

### DIFF
--- a/teams/miri.toml
+++ b/teams/miri.toml
@@ -8,6 +8,8 @@ alumni = ["solson"]
 
 [permissions]
 bors.miri.review = true
+bors.rust.try = true
+perf = true
 
 [website]
 name = "Miri"


jyn514 (dismissed):

RalfJung:
No objection, though I don't think we ever needed crater runs for Miri.

teams/miri.toml
```diff
@@ -8,6 +8,9 @@ alumni = ["solson"]
 
 [permissions]
 bors.miri.review = true
+bors.rust.try = true
+crater = true
```
rylev, line 12: If crater runs have never been needed for miri, and they're relatively rare as is, I think it makes sense to remove this for now.

rylev (changes requested):

rylev (approved):

PR #907, opened 2022-12-24
demote ZMD to alumni contributor

zackmdavis:
Unfortunately, due to competing life demands, I only have [one commit](https://github.com/rust-lang/rust/commit/14eb94fe7a3ca3a92d5d7e42744d105c0fc6d527) in the compiler in the last three years. :cry: :cold_sweat:  According to the text of [RFC 2689](https://github.com/rust-lang/rfcs/blob/master/text/2689-compiler-team-contributors.md), I should have been demoted to alumni status some time ago. I'm assuming this PR is the correct way to do that?

### DIFF
--- a/teams/compiler-contributors.toml
+++ b/teams/compiler-contributors.toml
@@ -31,11 +31,11 @@ members = [
   "varkor",
   "WaffleLapkin",
   "Xanewok",
-  "zackmdavis",
 ]
 alumni = [
     "Centril",
     "LeSeulArtichaut",
+    "zackmdavis",
 ]
 
 [permissions]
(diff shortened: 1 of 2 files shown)


rylev (approved):
Thanks so much! I don't believe there's any reason to not have an alumni list for wg-rustfix as well if you want to add that.

zackmdavis:
@rylev Thanks, but I don't think workgroups need alumni (as contrasted to compiler team alumni being specified as a thing in RFC 2689). (Plus, I only have [three commits](https://github.com/rust-lang/rustfix/commits?author=zackmdavis) in rustfix.)

PR #911, opened 2023-01-02
Create `infra-bors` team

Kobzol:
I'm not sure if I should fill any permissions or github orgs.


Mark-Simulacrum:
I think we should update the (new) bors repo with this team having write (or maintain) access.

This should probably also be a subteam of the infra team.

pietroalbini (dismissed):

pietroalbini:
Changes look good, we should also manage the new bors repo and the legacy homu repo within this team (using the new tooling to configure repos with the team repo).

Kobzol:
Sorry for the force push after approval, I created the PR on top of a stale main branch, so I had to update it.

teams/infra-bors.toml
```diff
@@ -0,0 +1,10 @@
+name = "infra-bors"
```
rylev, line 1: Nit: is "bors-infra" a more appropriate name? As it is, it reads like we're trying to indicate this is a subteam of infra (which is not the highest priority information) while "bors-infra" seems to indicate what the team is (i.e., the team in charge of the infrastructure related to bors).

Kobzol: There was already `infra-admins`, so I took inspiration in that. I think that the naming `higher level -> lower level` makes sense, but I'm not opposed to changing it.

Mark-Simulacrum: Most similar teams are named with parent prefix (e.g., libs-contributors, compiler-contributors, etc.)

Renaming should be cheap down the line in any case.

Mark-Simulacrum:
> Changes look good, we should also manage the new bors repo and the legacy homu repo within this team (using the new tooling to configure repos with the team repo).

I think this part is missing from the PR?

Mark-Simulacrum (approved):

PR #918, opened 2023-01-12
Add ids for `WaffleLapkin`

WaffleLapkin:
And also fix the "schema". BTW, is there any docs that describe how `zulip-id` and especially `discord-id` are used?


rylev (approved):
There are no docs in this repo (I suppose because this information is provided as is without knowledge of what it might be used for). The primary use for Zulip IDs is for [`sync-team`](https://github.com/rust-lang/sync-team) which uses those ids to sync Zulip user groups. I'm not aware of a use for the discord ids, but Discord hasn't really been used by most of the project since I started maintaining this code base.

PR #919, opened 2023-01-15
Create a bootstrap sub-team of infra and add ozkanonur to it

jyn514:
Infra has quite broad permissions. Create a subteam so that people can be recognized as a contributors without also having permission to bypass bors checks, etc.

Welcome, @ozkanonur! Thank you for your contributions and I look forward to continuing to work with you :)


rylev (approved):
🎉

PR #926, opened 2023-01-26
Add opsem team

JakobDegen:
This PR creates the opsem team, as specified in rust-lang/rfcs#3346 . Welcome everyone, I'm exciting to be getting this team off the ground!

The PR also archives @rust-lang/wg-unsafe-code-guidelines - that work is of course not forgotten, but instead being inherited by the new team. Thanks to everyone who has contributed!

cc @CAD97 , @digama0 , @saethlin , @RalfJung as members and @nikomatsakis and @joshtriplett as lang leads


JakobDegen:
CI is still failing, as Mario does not have an email in the the teams repo - @digama0 , could I ask you to submit a separate PR adding one? I believe github will give you a dummy/anonymous one if you prefer.

Besides that, the PR also refers to a couple things that don't exist yet, and there are a couple other specific infra items that we could use support on. I'm happy to repost these requests elsewhere or implement them myself via follow ups to this repo to the extent that I can (I just don't know enough about what needs to be done here). Specifically:

 - The rust-lang/opsem-team repo does not exist yet. Not sure if it'll be auto-created
 - I also expect that we'll want to keep the rust-lang/unsafe-code-guidelines repo around. Can the permissions on that be set up so that T-opsem has triage and merge perms?
 - For the T-opsem stream on Zulip, I think we should probably rename t-lang/unsafe-code-guidelines . Happy to create a new one if there are objections though.
 - Neither of the Zulip groups exist yet.

rylev:
> The rust-lang/opsem-team repo does not exist yet. 

We do not auto create repos for teams. You can add repos through this repository by following [these instructions](https://github.com/rust-lang/team/blob/master/docs/toml-schema.md#repos). You can see how this is done for other repos [here](https://github.com/rust-lang/team/tree/master/repos/rust-lang).

> I also expect that we'll want to keep the rust-lang/unsafe-code-guidelines repo around. Can the permissions on that be set up so that T-opsem has triage and merge perms?

If you can, I would prefer that we create a repo config file (just like the one described above) for the `unsafe-code-guidelines` repo. This will move configuration away from manual manipulation on GitHub (which is how we used to do things) to the new more transparent way of configuring through the team repo.

> For the T-opsem stream on Zulip, I think we should probably rename t-lang/unsafe-code-guidelines . Happy to create a new one if there are objections though.

I can do this once this PR is merged. (unfortunately Zulip streams are still manually managed by admins)

> Neither of the Zulip groups exist yet.

These will be automatically created once this PR is merged.

Happy to help with any of this! Thanks so much!

RalfJung:
I am not sure archiving the ucg repo is the right call. That will just lead to tons of duplication of issues. We did this for the old memory model repo and it was quite annoying, in the end i did the work of closing or moving the old issues basically alone and it took years.

IMO we should rename the repo, not create a new one.

RalfJung:
> this PR archives the wg, not the repo

Oh, I misunderstood then.

In that case, fine for me. Figuring out what to do with the repo (e.g. rename it to attach it more directly to the team) should probably be somewhere early on the team's agenda.

JakobDegen:
@rylev what more is needed to get this PR merged?

rylev (approved):

RalfJung:
Shouldn't merging this PR have created @rust-lang/t-opsem ? No such team seems to exist yet.

rylev:
There's a bug in sync-team that has been fixed and is waiting for a review: https://github.com/rust-lang/sync-team/pull/34

PR #929, opened 2023-01-29
add perf permission for parallel-rustc wg

SparrowLii:
The work of the parallel compiler is very dependent on the perf tool, I hope to obtain the permission to use the tool

### DIFF
--- a/teams/wg-parallel-rustc.toml
+++ b/teams/wg-parallel-rustc.toml
@@ -20,6 +20,7 @@ alumni = [
 ]
 
 [permissions]
+bors.rust.try = true
 perf = true
 
 [website]
--- a/teams/wg-parallel-rustc.toml
+++ b/teams/wg-parallel-rustc.toml
@@ -19,6 +19,9 @@ alumni = [
     "spastorino",
 ]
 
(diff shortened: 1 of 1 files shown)


teams/wg-parallel-rustc.toml
```diff
@@ -19,6 +19,9 @@ alumni = [
     "spastorino",
 ]
 
+[permissions]
+perf = true
+
```
lqd, line 24: Unless the `perf` permission also implies the right to trigger `try` builds (I don't believe it's the case), I think the following permission is also going to be needed.

```toml
bors.rust.try = true
```

lqd:
(IIRC the team's lead approval will also be needed, therefore cc-ing wg-parallel-rustc's lead @cjgillot)

cjgillot:
Ok on my end.

teams/wg-parallel-rustc.toml
SparrowLii: Thanks for pointing out! I have added the line

Mark-Simulacrum (approved):

PR #930, opened 2023-02-02
rename diannasoreil to shadows-withal

shadows-withal:

PR #932, opened 2023-02-03
Remove mikerite

ehuss:
The user https://github.com/mikerite has been deleted as far as I can tell (does not appear to be renamed). This is causing the check script to fail.

This will also require #930 to pass the checks.


Mark-Simulacrum:
Hm, could you cherry pick the other commit as well? That'll let us merge without bypassing branch protection.

ehuss:
Done.

Mark-Simulacrum (approved):

PR #933, opened 2023-02-03
Display a more helpful error message for a removed user

ehuss:
The current error message when a user is deleted is somewhat difficult to deal with:

```
[ERROR rust_team::validate] validation error: couldn't verify GitHub usernames: graphql error: Could not resolve to a node with the global id of 'MDQ6VXNlcjMzOTgzMzMy'.
```

The solution here is to go through the batch of 100 user IDs and query each one until it fails to determine which GitHub user was deleted.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -136,14 +136,49 @@ impl GitHubApi {
             }
         ";
 
+        let cant_resolve = |e: &Error| e.to_string().contains("Could not resolve to a node");
+
         let mut result = HashMap::new();
         for chunk in ids.chunks(100) {
-            let res: GraphNodes<Usernames> = self.graphql(
+            let res: GraphNodes<Usernames> = match self.graphql(
                 QUERY,
                 Params {
                     ids: chunk.iter().map(|id| user_node_id(*id)).collect(),
                 },
-            )?;
+            ) {
+                Ok(res) => res,
+                Err(e) => {
+                    if cant_resolve(&e) {
+                        // This error happens when a user is deleted. Provide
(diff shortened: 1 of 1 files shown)


rylev (approved):

src/github.rs
```diff
@@ -136,14 +136,49 @@ impl GitHubApi {
...
                     ids: chunk.iter().map(|id| user_node_id(*id)).collect(),
                 },
-            )?;
+            ) {
+                Ok(res) => res,
+                Err(e) => {
+                    if cant_resolve(&e) {
+                        // This error happens when a user is deleted. Provide
+                        // a more helpful error message to pinpoint it:
+                        for id in chunk {
```
rylev, line 154: Is doing a binary search worth the potential added complexity? Doing potentially 100 requests for the error message sounds like quite a lot...

rylev:
A rebase should fix the CI issue.

rylev (approved):

src/github.rs
rylev: Yes exactly. Binary search would definitely help reduce the number of requests. I agree we're probably ok just making the possible ~100 requests, but we *might* run into issues in the future. CI runs every hour, and I've also considered making it every 10 minutes (to make sure things are always in sync).

ehuss: Ah, I didn't know about the scheduled jobs.

Now that I think about it, I'm not actually sure what the limit is here. I think the rust-lang org is an Enterprise account, which I think has a limit of 15,000 requests per hour per repo using the GitHub Actions auto-generated token.

PR #937, opened 2023-02-07
Manage the cargo repo via teams

ehuss:
This should set up the rust-lang/cargo repo to be managed by the team sync. The main motivation here is to demote the Cargo team's access from Maintain to Write to remove the "Merge" button on PRs. I do not think any of us need any of the extra permissions afforded by the Maintain level. IIUC, this configuration should create a branch protection that only allows bors to push to master. 

I also believe this should still allow creating branches. If not, then this might interfere with the release process which creates a tag and branch.

### DIFF
--- a/repos/rust-lang/cargo.toml
+++ b/repos/rust-lang/cargo.toml
@@ -1,0 +1,10 @@
+org = "rust-lang"
+name = "cargo"
+description = "The Rust package manager"
+bots = ["bors", "rustbot"]
+
+[access.teams]
+cargo = "write"
+
+[[branch]]
+name = "master"


rylev (approved):
> I also believe this should still allow creating branches. 

Yes, this does not limit creating branches - it just protects the branches named in the config (in this case the master branch).

rylev (changes requested):
@ehuss I just checked, and cargo also has branch protections for rust-1.* branches. If we merge this, that branch protection will be deleted (as you [can see here](https://github.com/rust-lang/sync-team/blob/fa5cd3f77b2f324dd5965e1e48188451d605ff30/src/github/mod.rs#L321-L328)). I assume we want to keep that? The repo configuration does not yet have support for matching a group of branches. We would need to add that first.

ehuss:
Yea, we should probably keep that branch restriction. I can probably help add support for patterns. Can you say why the current version doesn't support wildcards? I would expect the `*` character to be the same as other characters. I did read somewhere that `*` branch protections don't work with the REST API and only via GraphQL, but I'm uncertain if that's the issue here.

rylev:
> Can you say why the current version doesn't support wildcards?

The current version deals in protected branches rather than branch protection rules. This means that if the repo config configures a protected branch name `rust-1.*`, the sync tool will look for a branch called `rust-1.*` and attempt to create or update the rule that protects that branch.

This approach is obviously wrong as we want to think in terms of protection rules that can potentially span past a single branch. One small reason I originally went down this route is that the REST API does not support querying for or creating branch protection rules (only getting and manipulating branch protections rules *for a given branch*). 

However, the [graphql API does have support for this functionality](https://docs.github.com/en/graphql/reference/mutations#createbranchprotectionrule) so we should be able to make the switch. 

The unfortunate thing is that the team API calls these "branches" instead of "branch protections". This makes sense in the status quo where we really configuring single branches. If we move to a path of creating branch protections, we might want to consider changing what the team API refers to these as. This would be a breaking change however...

rylev (approved):
With https://github.com/rust-lang/sync-team/pull/36 landing this can now be merged

ehuss:
@rylev Presumably we also want to add this?

```toml
[[branch]]
name = "rust-1.*"
```

rylev:
@ehuss yes - this was added in #945

PR #940, opened 2023-02-14
Rename wg-rls-2 to rust-analyzer and make it a team

Veykril:
Renames the wg-rls-2 and wg-rls-2-triage teams to wg-rust-analyzer and wg-rust-analyzer-contributors. Also gives the wg-rust-analyzer-contributors team bors permissions, as the core team is running out of reviewing hands.

### DIFF
--- a/teams/rust-analyzer.toml
+++ b/teams/rust-analyzer.toml
@@ -1,6 +1,5 @@
-name = "wg-rust-analyzer"
+name = "rust-analyzer"
 subteam-of = "compiler"
-kind = "working-group"
 
 [people]
 leads = ["Veykril"]
@@ -8,9 +7,8 @@ members = ["matklad", "lnicola", "Veykril", "flodiebold"]
 alumni = ["jonas-schievink"]
 
 [website]
-name = "rust-analyzer working group"
+name = "rust-analyzer team"
 description = "Experimenting with a new compiler architecture tailored for IDEs"
-repo = "https://rust-lang.github.io/compiler-team/working-groups/rust-analyzer/"
 zulip-stream = "t-compiler/rust-analyzer"
 
 [permissions]
--- a/teams/rust-analyzer.toml
+++ b/teams/rust-analyzer.toml
@@ -8,7 +8,7 @@ alumni = ["jonas-schievink"]
 
 [website]
 name = "rust-analyzer team"
-description = "Experimenting with a new compiler architecture tailored for IDEs"
+description = "Rust IDE support and error-resilient compiler front-end"
 zulip-stream = "t-compiler/rust-analyzer"
 
 [permissions]
--- a/teams/wg-rust-analyzer-contributors.toml
+++ b/teams/wg-rust-analyzer-contributors.toml
@@ -1,4 +1,4 @@
-name = "wg-rls-2-triage"
+name = "wg-rust-analyzer-contributors"
 subteam-of = "compiler"
 kind = "working-group"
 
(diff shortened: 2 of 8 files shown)


teams/wg-rust-analyzer-contributors.toml
```diff
@@ -1,10 +1,13 @@
-name = "wg-rls-2-triage"
+name = "wg-rust-analyzer-contributors"
```
rylev, line 1: Would it make sense to just call the team rust-analyzer (and make it a team instead of a working group)? cc @pnkfelix @wesleywiser

teams/rust-analyzer.toml
```diff
@@ -1,16 +1,14 @@
-name = "wg-rls-2"
+name = "wg-rust-analyzer"
```
rylev, line 1: We should probably rename this as well, no?

Veykril: Oh I missed that

rylev:
CI is failing now because there are some members of the rust-analyzer team who don't have email addresses but with r-a becoming a team they now join the all@ mailing list. 

I'd like to get approval of the change of r-a from a working group to a subteam from @pnkfelix and/or @wesleywiser before we merge.

Veykril:
As rylev pointed out, turning the r-a working groups into teams requires all active members to supply an e-mail:
Mails are missing for @edwin0cheng, @lowr and @lnicola. Are you fine with adding mails to your people files (or handing them to me so I can add them in this PR)?

Veykril:
On another note, do we support subteams of subteams? (i.e. making the contributors team a subteam of the r-a team which is a subteam of compiler?). Didn't make that change here because I assume that doesn't work and it not really mattering.

rylev:
> On another note, do we support subteams of subteams? (i.e. making the contributors team a subteam of the r-a team which is a subteam of compiler?). Didn't make that change here because I assume that doesn't work and it not really mattering.

We prevent subteams of subteams though I'm not 100% sure why. Let's keep it as a subteam of compiler and we can think about improving this later.

lowr:
Filed #943, but you can cherry-pick and close it if that's faster.

Veykril:
Probably easier if I pick it over 👍

teams/rust-analyzer.toml
```diff
@@ -1,16 +1,14 @@
...
 
 [people]
 leads = ["Veykril"]
 members = ["matklad", "lnicola", "Veykril", "flodiebold"]
 alumni = ["jonas-schievink"]
 
 [website]
-name = "RLS 2.0 working group"
+name = "rust-analyzer team"
 description = "Experimenting with a new compiler architecture tailored for IDEs"
```
wesleywiser, line 11: Nit: maybe update the description too since we're renaming it?

wesleywiser (approved):
@pnkfelix and I discussed this and we're both 👍 on the idea!

teams/rust-analyzer.toml
rylev: @Veykril Does this description still fit or shall we update it?

lnicola: Maybe "Rust IDE support and error-resilient compiler front-end" or something like? @matklad?

I think Veykril is less around these days, not sure if I can push to this branch.

Veykril: Updated it to lnicola's suggestion

lnicola (approved):

rylev (approved):

PR #946, opened 2023-02-22
Remove the branches field from repos API

rylev:
**BREAKING CHANGE**

With https://github.com/rust-lang/sync-team/pull/38 merged no one is using these fields any more. Since these are duplicates of other fields, it would be much cleaner to remove these before some other user starts to depend on them.


Mark-Simulacrum (approved):
Deploy when ready.

PR #954, opened 2023-03-07
Remove kw-fn

ehuss:
The GitHub user `kw-fn` has deleted their account.  This should get CI working again.

cc @yaahc


rylev (approved):

rylev:
If this user has a new handle, it should be easy to add them back.

PR #960, opened 2023-03-13
Remove camsteffen from clippy team

camsteffen:
I haven't been contributing for a while now so I think it's time. It's been fun!


Manishearth (approved):
Thanks for all your work!

r? @jntrnr

jntrnr (approved):

PR #963, opened 2023-03-20
Give types team members access to perf, crater and dev-desktops.

oli-obk:
Not all our members are compiler-contributors, compiler or other similar team members.

In https://github.com/rust-lang/rust/pull/109388#issuecomment-1476049002 @aliemjay was unable to run perf.

### DIFF
--- a/teams/types.toml
+++ b/teams/types.toml
@@ -40,7 +40,6 @@ extra-people = ["eholk"]
 [permissions]
 bors.chalk.review = true
 bors.rust.review = true
-bors.rust.try = true
 perf = true
 crater = true
 dev-desktop = true
--- a/teams/types.toml
+++ b/teams/types.toml
@@ -39,3 +39,7 @@ extra-people = ["eholk"]
 
 [permissions]
 bors.chalk.review = true
+bors.rust.try = true
+perf = true
+crater = true
+dev-desktop = true
--- a/teams/types.toml
+++ b/teams/types.toml
@@ -39,6 +39,7 @@ extra-people = ["eholk"]
 
 [permissions]
 bors.chalk.review = true
+bors.rust.review = true
 bors.rust.try = true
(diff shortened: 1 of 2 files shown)


teams/types.toml
```diff
@@ -39,3 +39,6 @@ extra-people = ["eholk"]
 
 [permissions]
 bors.chalk.review = true
+perf = true
```
compiler-errors, line 42:
```suggestion
bors.rust.try = true
perf = true
```

Also need to give try permissions.

teams/types.toml
```diff
@@ -39,3 +39,7 @@ extra-people = ["eholk"]
 
 [permissions]
 bors.chalk.review = true
+bors.rust.try = true
+perf = true
+crater = true
```
rylev, line 44: Crater runs don't happen super often. Given that they take a large amount of expensive machinery to run it seems reasonable to limit its use. Is there any reason the types team members can't reach out to a compiler team member when this is needed?

compiler-errors: > Crater runs don't happen super often

Types team members run crater far more frequently than most other compiler contributors.

teams/types.toml
```diff
@@ -39,3 +39,7 @@ extra-people = ["eholk"]
 
 [permissions]
 bors.chalk.review = true
+bors.rust.try = true
```
aliemjay, line 42:
```suggestion
bors.rust.review = true
bors.rust.try = true
```
needs review permission as well.

teams/types.toml
rylev: Sounds find to me. cc @Mark-Simulacrum and @pietroalbini

Mark-Simulacrum: I'm fine with types having crater access.

rylev (approved):

PR #973, opened 2023-04-08
Add thomcc to `T-opsem/meetings` ping group

thomcc:


rylev (approved):
cc @JakobDegen and @RalfJung for awareness

PR #974, opened 2023-04-08
Add myself to people and cloud-compute

Sp00ph:
@jyn514 [recommended](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Cloud.20compute.20questions/near/347918804) I just open this PR and state my reasons for wanted access here:

I often go weeks at a time where I only have access to a slow laptop, which routinely takes minutes for e.g. a stage 0 recompile of the stdlib after a 1 line change, and which also chokes when trying to use rust-analyzer on the stdlib or the compiler source. I believe access to the dev desktops would boost my productivity massively in those time periods. I would like to use them to work on the stdlib and maybe also move into working on the compiler itself. I would also like to (if this is deemed an appropriate use case) work on improvements of the LLVM codegen, since this would also speed up programs produced by the Rust compiler.

### DIFF
--- a/people/Sp00ph.toml
+++ b/people/Sp00ph.toml
@@ -1,0 +1,5 @@
+name = 'Markus Everling'
+github = 'Sp00ph'
+github-id = 61327188
+zulip-id = 563417
+email = "markuseverling@gmail.com"
--- a/teams/cloud-compute.toml
+++ b/teams/cloud-compute.toml
@@ -13,6 +13,7 @@ members = [
     "JoelMarcey",
     "nellshamrell",
     "ouz-a",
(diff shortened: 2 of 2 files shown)


jyn514 (approved):

rylev (approved):
cc @jdno for awareness

jdno:
Thanks for the ping, @rylev!

Welcome to the dev desktops, @Sp00ph. If you have any questions or need help, feel free to ping me in the [t-infra](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra) stream on Zulip.

PR #975, opened 2023-04-11
add repo for the trait system refactor initiative

lcnr:
cc @rust-lang/initiative-trait-system-refactor

this should supersede both https://github.com/lcnr/solver-woes/issues and https://github.com/compiler-errors/next-solver-hir-issues/issues and I also want a place to store all the design decisions made before stabilization.

### DIFF
--- a/repos/rust-lang/trait-solver-refactor-initiaitive.toml
+++ b/repos/rust-lang/trait-solver-refactor-initiaitive.toml
@@ -1,8 +1,0 @@
-org = "rust-lang"
-name = "trait-solver-refactor-initiative"
-description = "The Rustc Trait Solver Refactor Initiative"
-bots = []
-
-[access.teams]
-project-trait-system-refactor = "maintain"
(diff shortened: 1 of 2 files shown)


lcnr:
why is initiative-trait-system-refactor not a team? :thinking: 

the name in the teams repo is project-trait-system-refactor but that also didn't work :thinking: what do i actually have to use there?

jyn514:
maybe it doesn't like that this is a "project-group" and not a team? not sure if it's okay to relax that restriction, I'm not familiar with the distinction

https://github.com/rust-lang/team/blob/1a232e520c076e84eb2c30ca45663575e89132c1/teams/project-trait-system-refactor.toml#L2

Mark-Simulacrum:
I think this is a bug in the validation logic, I'll see if I can push a fix shortly.

Mark-Simulacrum (approved):

PR #983, opened 2023-04-19
crates-io: Remove obsolete `crates-io-ops-bot.staging-crates-io` permission

Turbo87:
We are no longer using neither the bot nor Discord.

This change appears to be required before we can add new team members that don't have a Discord ID.

/cc @rust-lang/core @rust-lang/infra


Turbo87:
Note that this likely requires further cleanup, since the discord bot appears to be mentioned quite a few more times throughout the repository, but I don't know enough about the infrastructure here to confidently rip it all out 😅

rylev (approved):

PR #992, opened 2023-05-02
Update WG-prioritization member list

apiraino:
Removing inactive members and moving to `alumni` section past members.

cc: @wesleywiser

r? @Mark-Simulacrum

### DIFF
--- a/teams/wg-prioritization.toml
+++ b/teams/wg-prioritization.toml
@@ -5,34 +5,33 @@ kind = "working-group"
 [people]
 leads = ["apiraino", "wesleywiser"]
 members = [
-    "am-1t",
-    "apiraino",
-    "bawerd",
-    "camelid",
-    "djcarpe",
     "Dylan-DPC",
-    "frxstrem",
+    "apiraino",
+    "camelid",
     "hameerabbasi",
     "hkmatsumoto",
     "inquisitivecrystal",
-    "JamesPatrickGill",
-    "jechasteen",
-    "JohnTitor",
     "lcnr",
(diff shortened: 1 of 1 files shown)


apiraino:
From the Ci fail:
> [ERROR rust_team::validate] validation error: 'TaKO8Ki' was specifically excluded from the Zulip group 'WG-prioritization/alerts' but they were already not included

I honestly don't understand what this sentence is trying to tell me :)

wesleywiser:
I think it might be saying that the list here needs to be updated as well:

https://github.com/rust-lang/team/blob/4e723863c59ebaa01095f5e2641cde17b0e440ea/teams/wg-prioritization.toml#L59

wesleywiser (approved):
lgtm with the excluded-people list updated

wesleywiser (approved):

rylev:
@apiraino it looks like CI is failing because some folks were removed from the team without being moved to alumni. Validation requires that if someone is in the team repo that they must be there for a reason (even if the reason is to be mentioned in an alumni section). To fix this, we'll need to either add those folks to the alumni section or remove them entirely from the repo. I'll leave it to you to decide which is more appropriate.

pnkfelix:
@apiraino why not move them to alumni then?

apiraino:
ok, I've put them all into the `alumni` list, though this does not completely reflect reality.

@rylev can you merge when you have one sec? thanks

wesleywiser (dismissed):

Mark-Simulacrum (approved):

PR #1000, opened 2023-05-24
move myself from compiler to compiler contributors

nikomatsakis:
I won't lie, it makes me a bit sad to step back from @rust-lang/compiler, but the reality is that I haven't been actively involved in compiler development for some time, so I think contributors is a better fit. It's so exciting to see how vibrant and active the team has become! 💜 you all.

r? @pnkfelix @wesleywiser


wesleywiser (approved):
@nikomatsakis thanks for all of your contributions, advice, mentoring and leadership to the compiler and the compiler team over the years!

PR #1005, opened 2023-05-31
Update WG memberships

joshtriplett:
I was the liaison when these groups got started, and participated early
on, but have not been an active member for a while.


nikomatsakis (approved):

nikomatsakis:
We probably should clean-up these working group definitions, but +1 to this removal.

Mark-Simulacrum (approved):

PR #1013, opened 2023-06-16
Rename maegamist to basil-cow

jackh726:
To fix broken CI


rylev:
@jackh726 you'll have to change the reference `wg-traits.toml` as well.

PR #1017, opened 2023-06-19
add rustfmt zulip group

calebcartwright:


calebcartwright:
I recognize the website section still has the discord info, but I don't want to change that at this time (we still kinda use both) unless it's necessary to do so in order for us to be able to `@t-rustfmt` in zulip.

pietroalbini (approved):

PR #1026, opened 2023-06-25
Update the branch name on libc repo

JohnTitor:
The default branch is now `main`: https://github.com/rust-lang/libc


JohnTitor:
And removed `highfive` from the bots entry as [rustbot now takes care of assignment](https://github.com/rust-lang/libc/pull/3283#issuecomment-1606064279).

rylev (approved):

PR #1030, opened 2023-07-14
Fix naming of rustc_apfloat repo

wesleywiser:
I'm not sure if it will cause any problems to do this renaming but currently both repos exist

image


Mark-Simulacrum (approved):
This should do the right thing in terms of modifying the repo we imported, I think we'll need manual cleanup for the spuriously created one though.

PR #1032, opened 2023-07-19
Allow extending teams from other teams

rylev:
This PR allows extending team membership by the membership of another team. Concretely this PR changes the following:

* `Team::members`, `Data::active_members`, and `Permissions::allowed_people` now all return a `Result` since getting all members can fail if an included team does not exist.
* Added `Team::contains_person` to test whether a team contains a person as a member
* Made `wg-leads` a marker team instead of a working group
* Created a `project-group-leads` marker team
* Removed the use of the various special "include-*" options from all teams except for the special marker teams that those options exist for. All other teams that also want to include those members, can use `included-teams` instead.

And lastly, the entire point of this PR:
* Made the inside-rust-reviewers include the leadership council

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -55,6 +55,9 @@ members = [
 alumni = [
     "buildbot",
 ]
+# Optional, name of other teams whose members will be included as members of this team.
+# Defaults to empty.
+included-teams = []
 # Optional, include all members of all other teams.
 # Defaults to false.
 include-all-team-members = false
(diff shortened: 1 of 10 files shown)


Mark-Simulacrum:
Hm, this seems good (looks like CI is failing). For the specific use case of council it probably makes sense to just grant write permissions in general on the repo rather than just inside rust.

But happy to approve once CI is fixed.

rylev:
@Mark-Simulacrum done

Mark-Simulacrum (approved):

PR #1040, opened 2023-08-09
Create the rustdoc-contributors team

GuillaumeGomez:
Creating this subteam will allow to finally give recognition to @fmease's work and also give them access to our tools.

Please don't merge until the rustdoc team gave its approval. The discussion for this PR is [here](https://rust-lang.zulipchat.com/#narrow/stream/266220-rustdoc/topic/rustdoc.20subteam).

cc @rust-lang/rustdoc


teams/rustdoc-contributors.toml
```diff
@@ -0,0 +1,23 @@
...
+subteam-of = "rustdoc"
+
+[people]
+leads = []
+members = [
+    "fmease"
+]
+
+[permissions]
+crater = true
```
rylev, line 11: Are we sure this is necessary?

teams/rustdoc-contributors.toml
```diff
@@ -0,0 +1,23 @@
...
+bors.rust.review = true
+dev-desktop = true
```
rylev, line 13: cc @jdno for awareness.

rylev (dismissed):
Looks good. I'm unsure about the need for crater though. It would be nice to limit who has access to crater runs (since they are not cheap to run).

teams/rustdoc-contributors.toml
GuillaumeGomez: We can remove `crater` permissions for the time being. Can always be added later on or rustdoc team members can run it if it seems necessary.

fmease:
It's unfortunate that we can't have subsubteams (rustdoc-contributors &lt; rustdoc &lt; devtools) but I guess that goes overboard. I wonder what can be done about it. If we made it a subteam of devtools instead as GuillaumeGomez suggested to me, would that work? Not sure if the ping groups are set up correctly in such case (pinging rustdoc would ideally ping rustdoc-contributors, too, right?).

rylev:
I think we want the ability for subteams to have subteams. cc @Mark-Simulacrum and @pietroalbini for thoughts on why we can't just remove this limitation.

pietroalbini:
That'd break the website right now I think.

teams/rustdoc-contributors.toml
GuillaumeGomez: Done.

fmease:
I guess subteams of subteams should get laid out flat below their parent team, that looks like the most straightforward solution.
This way, *Rustdoc team contributors* gets placed below *Rustdoc team* over at https://www.rust-lang.org/governance/teams/dev-tools.

fmease:
I've now opened #1046 (hard blocker) and rust-lang/www.rust-lang.org#1854 (soft blocker).

The creation of *rust-contributors* has been pretty much approved (6 explicit approvals, 1 pending approval).

fmease:
Alright, 7 out of 7 rustdoc team members approve (see linked Zulip topic) :tada: Thanks everyone!

teams/rustdoc-contributors.toml
```diff
@@ -0,0 +1,22 @@
...
+perf = true
+
+[[github]]
+orgs = ["rust-lang", "rust-lang-nursery"]
+extra-teams = ["rustdoc"]
+
+[website]
+name = "Rustdoc team contributors"
+description = "Contributing to rustdoc on a regular basis"
```
fmease, line 21:
```suggestion
description = "Contributing to Rustdoc on a regular basis"
```

It's also capitalized in `rustdoc.toml`

https://github.com/rust-lang/team/blob/b1b37232eb3a91d75de4168c53d882653c4ba7a6/teams/rustdoc.toml#L38

GuillaumeGomez: :+1:

fmease:
Since #1046 (*Permit subteams to have subteams*) is now merged, could somebody kick off a CI run? Thanks!
Once it's green, this PR should be ready to be merged, right? As far as I can tell, it's sufficiently approved :)

rylev:
@GuillaumeGomez can you rebase this against the new version of master?

GuillaumeGomez:
Rebased.

PR #1058, opened 2023-08-29
Remove myself (khionu) from mods and council

khionu:
[Not a farewell, just a bye-for-now.](https://khionu.net/blog/2023-stepping-down-from-rust/)


rylev (approved):
❤️

PR #1062, opened 2023-08-30
remove wg-parselib

matklad:
It didn't really fully existed anyway. Still, we should share the darn parser someday, this dragon still awaits its knight!


rylev (approved):
cc @davidtwco @wesleywiser for awareness

PR #1046, opened 2023-08-17
Permit subteams to have subteams

fmease:
This would allow us to create the team *rustdoc-contributors* (#1040).

Let me know if I should add a test anywhere.
Should we enforce a depth limit?

<details><summary>Outdated information</summary>

According to https://github.com/rust-lang/team/pull/1040#issuecomment-1672989957, nothing should be negatively affected except for the website. I've built its server locally with a patched `rust_team_data` package (part of this repo) and a patched `rust_team_data` static API (that contains the new team) and the only consequence is that *rustdoc-contributors* doesn't show up anywhere on the website (it doesn't crash the server or anything like that). Therefore, updating the website is not a hard blocker.

I've opened rust-lang/www.rust-lang.org#1854 proposing to add proper website support for subsubteams (etc.). I plan on implementing it (or GuillaumeGomez will).

</details>

PR #1070, opened 2023-09-08
move wg-polonius to t-types

lqd:
As described in the [initial t-types announcement/plans](https://blog.rust-lang.org/2023/01/20/types-announcement.html#the-types-team):

> Another relevant working group is the [polonius working group](https://rust-lang.github.io/compiler-team/working-groups/polonius/), which primarily works on the design and implementation of the [Polonius](https://github.com/rust-lang/polonius) borrow-checking library. While the working group itself will remain, it is now also under the purview of the types team.

This PR moves the WG under t-types rather than t-compiler. Note that the move is limited in what the team repo and website support (much like t-types' structure itself): [subteams currently can't have working groups](https://github.com/rust-lang/team/actions/runs/6122552119/job/16618573188#step:5:11). 

It's not a big deal:
1. we can still have the WG as a child of t-compiler in the team repository -- this is still ultimately true transitively :) -- and document that it's a t-types WG. This is what this PR does.
2. or, have the wg be a team. @fmease has recently made changes to allow subteams to have their own subteams. 

Let me know what you think @nikomatsakis and @jackh726. It feels the WG structure has merit in this case, compared to making it a t-types subteam. It seems fine to me to document that this is a t-types WG, but still attach it to t-compiler in the existing data structure of the team repo and website UI.

(This ultimately feels fixable, but I'm not sure it's absolutely necessary right now)

### DIFF
--- a/teams/wg-polonius.toml
+++ b/teams/wg-polonius.toml
@@ -1,4 +1,5 @@
 name = "wg-polonius"
+# Really should be "types", but the team repo doesn't handle that right now
 subteam-of = "compiler"
 kind = "working-group"
 
@@ -10,7 +11,7 @@ members = ["lqd", "nikomatsakis", "amandasystems", "matthewjasper", "ecstatic-mo
 name = "Polonius working group"
 description = "Working on an experimental new borrow-checker implementation"
 repo = "https://rust-lang.github.io/compiler-team/working-groups/polonius/"
-zulip-stream = "t-compiler/wg-polonius"
+zulip-stream = "t-types/polonius"
 
 [[github]]
 orgs = ["rust-lang"]


fmease:
Ah, in #1046 I only allowed subteams (like *types*) to have subteams but not working groups (like *wg-polonius*) since there isn't a nice way to render the latter on the website. If you find a solution to nicely render WGs of subteams on the website (see rust-lang/www.rust-lang.org#1857), you can just lift this restriction, I'd say.

You dropping `kind = "working-group"` makes the subsubteam check succeed but of course the prefix check fail.

lqd:
Thanks, I was looking at the PR to remove/update that limitation.

jackh726 (approved):
Would love to eventually get proper support of this (and types as subteam of lang+compiler), but this is good for now :)

rylev (approved):

PR #1090, opened 2023-10-04
Add rustup repository

jdno:
The permissions for the rust-lang/rust repository are out-of-date, and instead of manually updating them we want to manage the repository automatically.

Fixes https://github.com/rust-lang/infra-team/issues/80

### DIFF
--- a/repos/rust-lang/rustup.toml
+++ b/repos/rust-lang/rustup.toml
@@ -1,0 +1,11 @@
+org = "rust-lang"
+name = "rustup"
+description = "The Rust toolchain installer"
+bots = ["rustbot"]
+
+[access.teams]
+rustup = "maintain"
+mods = "maintain"
+
+[[branch-protections]]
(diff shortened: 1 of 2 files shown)


repos/rust-lang/rustup.toml
```diff
@@ -0,0 +1,11 @@
+org = "rust-lang"
+name = "rustup"
+description = "The Rust toolchain installer"
+bots = ["rustbot"]
+
+[access.teams]
+rustup = "maintain"
+mods = "maintain"
```
rylev, line 8: I don't think we need this. The mods have org wide permissions to do what they need to do.

rylev (approved):

repos/rust-lang/rustup.toml
jdno: Perfect!

rylev (approved):

PR #1091, opened 2023-10-04
create the *lang-ops* team

nikomatsakis:
This team is chartered to prepare meeting agendas, help with issue triage, schedule and document meetings, and generally perform the other "operations" that make the lang team function.

Currently its sole membership is TC but the hope is that it will grow!

cc @rust-lang/lang

### DIFF
--- a/teams/lang-ops.toml
+++ b/teams/lang-ops.toml
@@ -1,0 +1,18 @@
+name = "lang-ops"
+subteam-of = "lang"
+
+[people]
+leads = ["TC"]
+members = [
+    "TC",
+]
+alumni = [
+]
+
+[[github]]
+orgs = ["rust-lang"]
+
+[website]
+name = "lang-ops team"
+description = "Operations for the lang team: preparing agenda, issue triage, scheduling and documenting meetings"
+zulip-stream = "t-lang/ops"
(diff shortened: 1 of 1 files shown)


nikomatsakis:
I'm confused by the errors; TC.toml is definitely present in people...

teams/lang-ops.toml
```diff
@@ -0,0 +1,18 @@
+name = "lang-ops"
+subteam-of = "lang"
+
+[people]
+leads = ["TC"]
+members = [
+    "TC",
+]
```
compiler-errors, line 8:
```suggestion
leads = ["traviscross"]
members = [
    "traviscross",
]
```

like wg-async:
https://github.com/rust-lang/team/blob/765876e23d928b3934a6b07855ed741c4dd0b837/teams/wg-async.toml#L17

nikomatsakis: Huh, OK, we use github names I guess, and not the name of the "people" file?

nikomatsakis:
This isn't really a *proposal* -- it's been discussed already and we decided to do it. So, renaming to make it clear this is ready to merge.

Mark-Simulacrum (approved):

PR #1092, opened 2023-10-10
Automatically add compiler-contributors to the private decision making stream

oli-obk:
unblocks https://github.com/rust-lang/compiler-team/issues/649

Also

* does the same for @rust-lang/types 
* also removes extra-people that are already compiler-contributors

### DIFF
--- a/teams/compiler-contributors.toml
+++ b/teams/compiler-contributors.toml
@@ -57,3 +57,12 @@ extra-teams = ["compiler"]
 name = "Compiler team contributors"
 repo = "http://github.com/rust-lang/compiler-team"
 description = "Contributing to the Rust compiler on a regular basis"
+
+[[zulip-groups]]
+name = "T-compiler"
+
+[[zulip-groups]]
+name = "T-compiler/meeting"
+
+[[zulip-groups]]
+name = "T-compiler/contrib-private"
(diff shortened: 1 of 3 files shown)


teams/compiler-contributors.toml
```diff
@@ -57,3 +57,12 @@ extra-teams = ["compiler"]
...
 description = "Contributing to the Rust compiler on a regular basis"
+
+[[zulip-groups]]
+name = "T-compiler"
+
+[[zulip-groups]]
+name = "T-compiler/meeting"
+
+[[zulip-groups]]
+name = "T-compiler/contrib-private"
```
oli-obk, line 68: not sure this is needed, as contributors are a subteam of compiler, but I hope it is, otherwise types members automatically get access to the private stream, too

compiler-errors, line 68: How does group membership translate to stream membership? I thought those were different concepts in Zulip.

oli-obk: oh 🤦 I just saw the matching names and conveniently overlooked the `group` part of `zulip-groups`.

oof. zulip will just add individual users from a team to a stream if you add a team. So we'll need to manage this completely on our end.

PR #1094, opened 2023-10-13
Update and split "arm" team.

jacobbramley:
This adds an "arm-maintainers" team, which we can refer to in rust-lang/rust#116004.

`cargo run -- dump-team arm` is unaffected.

### DIFF
--- a/people/jacobbramley.toml
+++ b/people/jacobbramley.toml
@@ -1,3 +1,4 @@
 name = 'Jacob Bramley'
 github = 'jacobbramley'
 github-id = 5206553
+zulip-id = 399135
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum (approved):

RalfJung:
FWIW, the newly introduced "arm-maintainers" team here was not actually pingable. Triagebot only allows pinging teams that are also explicitly listed in `triagebot.toml`. https://github.com/rust-lang/rust/pull/161307 finally fixes this, but I wonder why we need such a 2-stage process where the 2nd stage is inevitably going to be forgotten.

PR #1097, opened 2023-10-14
Move @japaric to rust-embedded alumni

adamgreig:


rylev (approved):

teams/wg-embedded-cortex-r.toml
```diff
@@ -5,7 +5,6 @@ kind = "working-group"
 [people]
```
rylev, line 5: Should these team be removed since it has no members?

adamgreig: For the wg, the team still exists but is empty (so that it can continue to own the repos etc), but I'm happy to remove it in this PR and re-add it when a new member joins, if that's better for you.

PR #1101, opened 2023-10-19
Add forge repo permissions.

ehuss:
I would like to set up the forge so that a broader set of people have permission to help merge PRs. My thinking is to set up auto-assignment via rustbot so that individual teams will be responsible for their respective areas. Additionally, I can help with overall management of the repo itself (like dealing with mdbook or blacksmith). I think giving a broad set of people permission to help and encourage participation, avoiding things getting stuck, and to avoid putting the review responsibility on a small number of people.

I don't know if this would need sign off from the council or not. I see this as mainly being under the purview of the infrastructure team (and we can formalize that if needed).

cc https://github.com/rust-lang/leadership-council/issues/47

### DIFF
--- a/repos/rust-lang/rust-forge.toml
+++ b/repos/rust-lang/rust-forge.toml
@@ -1,0 +1,20 @@
+org = 'rust-lang'
+name = 'rust-forge'
+description = 'Information useful to people contributing to Rust'
+bots = ["rustbot"]
+
+[access.teams]
+community = "maintain"
+compiler = "maintain"
+compiler-contributors = "maintain"
+crates-io = "maintain"
+docs-rs = "maintain"
+infra = "maintain"
+lang = "maintain"
+leadership-council = "maintain"
+libs = "maintain"
+release = "maintain"
+triagebot = "maintain"
+
+[[branch-protections]]
+pattern = "master"


repos/rust-lang/rust-forge.toml
```diff
@@ -0,0 +1,19 @@
+org = 'rust-lang'
+name = 'rust-forge'
+description = 'Information useful to people contributing to Rust'
+bots = ["rustbot"]
+
+[access.teams]
+community = "maintain"
+compiler = "maintain"
```
lqd, line 8: Does that include subteams? If not, maybe adding `compiler-contributors` could be helpful, as there are twice as many people?

nikomatsakis:
Big +1, I think we are too conservative with our permissions in general. I see no reason that people on a team should not be able to update forge. Giving access to specific areas seems fine, but I also think it's...a bit overkill, personally.

repos/rust-lang/rust-forge.toml
ehuss: It does not include subteams. I went ahead and added compiler-contributors.

apiraino:
I occasionally need to work on docs on forge.r-l.o and I am so grateful for this initiative!

Mark-Simulacrum (approved):

Mark-Simulacrum:
IMO, expanding permissions to forge is a clearly reversible action so I'm going to merge this. We can go back on that if people have concerns, but I don't expect that to be the case.

PR #1115, opened 2023-11-11
Remove @Disasm from the EWG RISC-V team and the risc-v marker team

Disasm:
Embedded work group PR: https://github.com/rust-embedded/wg/pull/717


rylev (approved):
This is fine, but you may wish to take advantage of the "alumni" key for the working groups to show former participation.

Disasm:
I added myself to the alumni list of `wg-embedded`. I don't really see a point of doing so for a separate team inside it.

PR #1118, opened 2023-11-16
Add rust-lang/spec repository.

m-ou-se:
cc @rust-lang/spec

### DIFF
--- a/repos/rust-lang/specification.toml
+++ b/repos/rust-lang/specification.toml
@@ -1,0 +1,7 @@
+org = "rust-lang"
+name = "spec"
+description = "The Rust specification"
+bots = ["rustbot", "rfcbot"]
+
+[access.teams]
+spec = 'maintain'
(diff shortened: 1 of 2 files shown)


repos/rust-lang/specification.toml
```diff
@@ -0,0 +1,7 @@
+org = "rust-lang"
+name = "spec"
+description = "The Rust specification"
+bots = ["rustbot", "rfcbot"]
+
+[access.teams]
+spec = 'maintain'
```
Urgau, line 7: May I suggest adding branch protection for the `main` branch.
```suggestion
spec = 'maintain'

[[branch-protections]]
pattern = 'main'
```

rylev (approved):
Looks good - waiting on whether @m-ou-se wants to add the branch protection or not.

m-ou-se:
I think we'll just enable that at a later point. While we're setting everything up, it might be easier to just be able to push directly. ^^

repos/rust-lang/specification.toml
m-ou-se: I think we'll just enable that at a later point. While we're setting everything up, it might be easier to just be able to push directly. ^^

PR #1110, opened 2023-11-07
Give stable-mir r+ access for their code

wesleywiser:
We want to give the stable-mir project the ability to r+ stable mir PRs in the compiler since that area is under their purview. 

r? @davidtwco

PR #1119, opened 2023-11-16
Allow assigning PRs to members of the project stable MIR

celinval:
These members were granted r+ permission here: https://github.com/rust-lang/team/pull/1110

However, trying to assign a PR to them will trigger a rustbot error. I think the problem is that we also need to add the team as repo maintainers.

### DIFF
--- a/repos/rust-lang/rust-forge.toml
+++ b/repos/rust-lang/rust-forge.toml
@@ -15,6 +15,7 @@ leadership-council = "maintain"
 libs = "maintain"
 release = "maintain"
 triagebot = "maintain"
+project-stable-mir = "maintain"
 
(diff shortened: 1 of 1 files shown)


ehuss:
Hi! This isn't the correct file to edit to give permissions on the `rust-lang/rust` repo. This file controls access to the https://github.com/rust-lang/rust-forge/ repo, which controls the forge website.

To be assignable, they need to be given permissions on the `rust-lang/rust` repo, which unfortunately is not yet managed by this automated infrastructure. Someone in the infra team will need to manually add them. I suggest following up in the `#t-infra` stream on Zulip.

celinval:
Gotcha! Let me do that. Thanks

PR #1120, opened 2023-11-16
Add burakemir to opsem extra people

burakemir:
Hello!

I am interested in following the discussions around Rust operational semantics, including the unsafe fragment, more closely.

- Burak

### DIFF
--- a/people/burakemir.toml
+++ b/people/burakemir.toml
@@ -1,0 +1,5 @@
+name = 'Burak Emir'
+github = 'burakemir'
+github-id = 10248140
+zulip-id = 638751
+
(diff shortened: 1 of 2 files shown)


rylev:
@burakemir we currently don't allow people who are not a member of any team to be added to the list of extra people, but perhaps we should. 

Thoughts? @Mark-Simulacrum @JakobDegen @RalfJung

JakobDegen:
I don't see why we wouldn't, I suppose. Theres nothing that that role gets you other than pinged when you want to be

RalfJung:
Yes we should allow that. I do assume that being in those extra-group does not grant any of the team permissions (for bors, crater, etc).

burakemir:
Thanks for confirming, I inferred from the CI failure. I don't want to cause anyone extra work and would be happy to keep track by other means than "extra people"

jackh726:
CI should be fixed by #1124, can you rebase?

burakemir:
Done.

rylev (approved):

PR #1122, opened 2023-11-22
teams/opsem.toml: Add joshlf to zulip-groups

joshlf:
Not sure if I've done this right since it doesn't actually look like Zulip uses usernames, but other folks seem to have used their GitHub usernames, so I'm doing the same 🤷‍♂️

### DIFF
--- a/teams/opsem.toml
+++ b/teams/opsem.toml
@@ -31,4 +31,4 @@ name = "T-opsem"
 
 [[zulip-groups]]
 name = "T-opsem/meetings"
-extra-people = ["chorman0773", "thomcc", "pnkfelix"]
+extra-people = ["chorman0773", "thomcc", "pnkfelix", "joshlf"]


rylev (changes requested):
@joshlf I don't believe this will work. The field expects GitHub names because the team repo uses GitHub names as the unique ID to identify people. This then gets resolved to a Zulip ID. However, you don't have a Zulip ID in joshlf.toml. You'll want to add that for this to work. 

cc @JakobDegen and @RalfJung for awareness.

RalfJung:
> other folks seem to have used their GitHub usernames, so I'm doing the same 

AFAIK this refers to the filenames in https://github.com/rust-lang/team/tree/master/people. They do not have to be the same as the github usernames.

joshlf:
Okay, added my zulip-id as well; thanks for the pointer!

rylev (approved):

PR #1134, opened 2023-12-01
Backfill all currently listed alumni's teams

dtolnay:
Based on the code from #1125 to extract former team membership from commit history.

This PR fills in team membership for all contributors currently listed in the deprecated alumni.toml, per Ryan's recommendation in https://github.com/rust-lang/team/pull/1126#pullrequestreview-1754881281 to bring back deleted teams in archived form.

The new archived teams added in this PR need to be filled in with other former members beyond this, but I will handle that separately. For now this PR is enough to bring the deprecated alumni team down to 0 members, and begin removing code related to it.

The only change to the website will be for birkenfeld, who will begin to appear under Clippy alumni at https://www.rust-lang.org/governance/teams/dev-tools#clippy. The other teams are archived teams, which do not get reflected on the website. All archived team members and birkenfeld will continue to appear in the overall alumni list at https://www.rust-lang.org/governance/teams/alumni.

PR #1136, opened 2023-12-04
Simplify validation of alumni team

dtolnay:
The alumni team used to allow explicit members as long as they were neither an active nor alumni of any other team.

Following #1134, explicit members of the alumni team are no longer allowed at all.

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -232,39 +232,14 @@ fn validate_team_members(data: &Data, errors: &mut Vec<String>) {
     });
 }
 
-/// Ensure alumni are not active
+/// Alumni team must consist only of automatically populated alumni from the other teams
 fn validate_alumni(data: &Data, errors: &mut Vec<String>) {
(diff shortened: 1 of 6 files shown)


src/validate.rs
```diff
@@ -232,39 +232,14 @@ fn validate_team_members(data: &Data, errors: &mut Vec<String>) {
...
+/// Alumni team must consist only of automatically populated alumni from the other teams
 fn validate_alumni(data: &Data, errors: &mut Vec<String>) {
-    let active_members = match data.active_members() {
-        Ok(ms) => ms,
-        Err(e) => {
-            errors.push(e.to_string());
-            return;
-        }
+    let Some(alumni_team) = data.team("alumni") else {
+        return;
```
rylev, line 238: I wonder if we should error here since if the alumni team is removed, presumably this whole check should be removed?

dtolnay: Good call. Fixed.

PR #1147, opened 2023-12-15
Add Rust blog repository to automation

Kobzol:
I'm not sure who should have merge rights (CC @Mark-Simulacrum), so the team list is empty currently.

### DIFF
--- a/repos/rust-lang/blog.rust-lang.org.toml
+++ b/repos/rust-lang/blog.rust-lang.org.toml
@@ -1,0 +1,10 @@
+org = "rust-lang"
+name = "blog.rust-lang.org"
+description = "Home of the Rust and Inside Rust blogs"
+bots = []
+
+[access.teams]
+inside-rust-reviewers = "write"
+
+[[branch-protections]]
+pattern = "master"


repos/rust-lang/blog.rust-lang.org.toml
```diff
@@ -0,0 +1,9 @@
+org = "rust-lang"
+name = "blog.rust-lang.org"
+description = "Home of the Rust and Inside Rust blogs"
+bots = []
+
+[access.teams]
```
rylev, line 6: Those with access currently are:
* core
* highfive
* inside-rust-reviewers
* leadership-council
* leads
* mods
* website

I don't think mods, website, and core are necessary any more, and I don't believe highfive is being used.

Kobzol: Thanks! Since `inside-rust-reviewers` contains both `leadership-council` and `leads`, maybe just `inside-rust-reviewers` is enough?

rylev: Sounds good to me! 

I know this isn't part of this PR, but I think `inside-rust-reviewers` was originally meant to be those that can merge anything in the Inside Rust blog, but we're using it as the the rights for who can merge any blog PR. That's something we should probably change ASAP

Kobzol: Since both blogs are in the same repository, can we even distinguish that? AFAIK we can't distinguish individual paths in a repository.

WaffleLapkin: There is the "code owners" thing that requires a review from a specific team depending on edited files... but I don't think it works with bors

rylev: > Since both blogs are in the same repository, can we even distinguish that? AFAIK we can't distinguish individual paths in a repository.

The hope is to add this, but you're correct that there is no support currently. In the meantime it might help to either change `inside-rust-reviewers` to `blog-reviewers` in acknowledgment of the status quo or to add a `main-blog-reviewers` marker team as a nod to this aspirational future.

rylev (approved):

PR #1148, opened 2023-12-15
Add maintainers for impl-trait-utils

tmandry:
Add @compiler-errors and @traviscross as maintainers.

### DIFF
--- a/repos/rust-lang/impl-trait-utils.toml
+++ b/repos/rust-lang/impl-trait-utils.toml
@@ -7,6 +7,8 @@ bots = []
 wg-async = "write"
 
 [access.individuals]
+compiler-errors = "maintain"
+TC = "maintain"
 tmandry = "maintain"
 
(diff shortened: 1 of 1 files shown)


repos/rust-lang/impl-trait-utils.toml
```diff
@@ -7,6 +7,8 @@ bots = []
 wg-async = "write"
 
 [access.individuals]
+compiler-errors = "maintain"
+TC = "maintain"
```
traviscross, line 11:
```suggestion
traviscross = "maintain"
```

The tools here seem to work based off of the GH username.

jackh726:
Any reason to not just bump wg-async as maintain? I think generally we've been trying to limit individual access like this.

tmandry:
Okay, updated.

jackh726 (approved):

PR #1151, opened 2023-12-17
Make alumni field non-optional

dtolnay:
```rust
// Teams must contain an `alumni = […]` field (even if empty) so that there
// is an obvious place to move contributors within the same file when
// removing from `members`.
//
// Marker teams are exempt from this, as well as teams which comprise only
// members of other teams via `include-team-leads` or similar; they do not
// need `alumni = […]`. For these teams, the correct place to put alumni is
// in the same team they're being included from.
```

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -52,6 +52,10 @@ members = [
 ]
 # Past members of the team. They will not be considered as part of the team,
 # but they will be recognized on the website.
+#
+# Most teams are required to have this alumni key, even if its value is an empty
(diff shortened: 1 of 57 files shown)


rylev (changes requested):
Seems fine to me. Can we add a note about this requirement to https://github.com/rust-lang/team/blob/master/docs/toml-schema.md

dtolnay:
Sure thing, good call. Done.

rylev (approved):

PR #1154, opened 2023-12-22
Add support for custom roles in website, beyond "Team leader"

dtolnay:
I would like to render "Editor" below @JoelMarcey's entry in https://www.rust-lang.org/governance/teams/lang#spec, in the same location that others have "Team leader".

Source: https://blog.rust-lang.org/inside-rust/2023/11/15/spec-vision.html#specification-team

Output of `cargo run -- dump-website` for translations:

```console
...
governance-team-wg-triage-name = Triage working group
governance-team-wg-triage-description = Triaging repositories under the rust-lang organisation

governance-team-wg-wasm-name = WebAssembly (WASM) working group
governance-team-wg-wasm-description = Improving on the end-to-end experience of embedding Rust code in JS libraries and apps via WebAssembly

governance-role-spec-editor = Editor
```

Output of `cargo run -- static-api` under v1/teams/spec.json:

```json
{
  "name": "spec",
  "kind": "team",
  "subteam_of": "lang",
  "members": [
    {
      "name": "Mara Bos",
      "github": "m-ou-se",
      "github_id": 783247,
      "is_lead": true
    },
    {
      "name": "Felix Klock",
      "github": "pnkfelix",
      "github_id": 173127,
      "is_lead": true
    },
    {
      "name": "Eric Huss",
      "github": "ehuss",
      "github_id": 43198,
      "is_lead": false
    },
    {
      "name": "Joel Marcey",
      "github": "JoelMarcey",
      "github_id": 3757713,
      "is_lead": false,
      "roles": [
        "spec-editor"
      ]
    }
  ],
```

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -49,6 +49,12 @@ members = [
     "rfcbot",
     "craterbot",
     "rust-timer",
+    # Any subset of members may hold custom roles, beyond "Team leader" which is
+    # controlled by the `leads` array above. See [[website.roles]]. Members with
+    # roles are written using an inline table as follows. A simple string member
+    # like "bors" is equivalent to {name = "bors", roles = []}.
+    { name = "Crab01", roles = ["cohost"] },
+    { name = "Crab02", roles = ["cohost"] },
 ]
 # Past members of the team. They will not be considered as part of the team,
 # but they will be recognized on the website.
@@ -140,10 +146,6 @@ weight = -100
 id = "cohost"
 # Text to appear on the website beneath the team member's name and GitHub handle.
 description = "Co-host"
-# Subset of the team's members holding this role. A member may hold arbitrarily
-# many roles. Each team member's roles will appear comma-separated in the same
-# order as roles are listed in the website.roles array.
-members = ["Crab01", "Crab03"]
 
 # Define the mailing lists used by the team
 # It's optional, and there can be more than one
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -133,6 +133,18 @@ zulip-stream = "t-lang"
 # Default is 0.
 weight = -100
 
+# Customized roles held by a subset of the team's members, beyond "Team leader"
+# which is rendered automatically for members of the `leads` array.
+[[website.roles]]
+# Kebab-case id for the role. This serves as a key for translations.
+id = "cohost"
+# Text to appear on the website beneath the team member's name and GitHub handle.
+description = "Co-host"
+# Subset of the team's members holding this role. A member may hold arbitrarily
+# many roles. Each team member's roles will appear comma-separated in the same
+# order as roles are listed in the website.roles array.
+members = ["Crab01", "Crab03"]
+
 # Define the mailing lists used by the team
 # It's optional, and there can be more than one
 [[lists]]
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -50,11 +50,12 @@ members = [
     "craterbot",
     "rust-timer",
     # Any subset of members may hold custom roles, beyond "Team leader" which is
-    # controlled by the `leads` array above. See [[website.roles]]. Members with
-    # roles are written using an inline table as follows. A simple string member
-    # like "bors" is equivalent to {name = "bors", roles = []}.
-    { name = "Crab01", roles = ["cohost"] },
-    { name = "Crab02", roles = ["cohost"] },
+    # controlled by the `leads` array above. Members with roles are written
+    # using an inline table as follows. A simple string member like "bors" is
+    # equivalent to {github = "bors", roles = []}. The strings in `roles` must
+    # be present as the `id` of some role in [[roles]] section below.
+    { github = "Crab01", roles = ["cohost"] },
+    { github = "Crab02", roles = ["cohost"] },
 ]
 # Past members of the team. They will not be considered as part of the team,
 # but they will be recognized on the website.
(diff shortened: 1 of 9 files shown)


dtolnay:
Another possible application: team affiliation in https://www.rust-lang.org/governance/teams/leadership-council#leadership-council. "Lang team rep"

src/static_api.rs
```diff
@@ -121,6 +135,9 @@ impl<'a> Generator<'a> {
                         github: (*github_name).into(),
                         github_id: person.github_id(),
                         is_lead: leads.contains(github_name),
+                        roles: website_roles
+                            .get(*github_name)
+                            .map_or_else(Vec::new, Vec::clone),
```
rylev, line 140: Nit: this might be slightly more clear:

```suggestion
                            .cloned().unwrap_or_default()
```

src/validate.rs
```diff
@@ -774,6 +776,64 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
...
+/// (because the role id becomes the Fluent id used for translation).
+fn validate_website_roles(data: &Data, errors: &mut Vec<String>) {
+    let mut role_descriptions = HashMap::new();
+
+    wrapper(
+        data.teams().chain(data.archived_teams()),
+        errors,
+        |team, errors| {
+            let team_name = team.name();
+            if let Some(website_data) = team.website_data() {
```
rylev, line 790: Nit: might help if we try to reduce the rightward drift a bit:

```suggestion
           let Some(website_data) = team.website_data() else { return Ok(()); };
```

rylev (approved):
Thanks for working on this! Overall it's a great change, and I'd be happy to merge except for one thing.

The current schema requires duplicating the member's name both in the team's `members` array as well as part of role's `members` array. Would it make sense to see if we can change the team `members` array schema to allow for specifying roles inline to the member declaration?

Perhaps something like:

```toml
members = [
  "rylev",
  { name =  "dtolnay", role = "foo" }
]
```

According to the [toml spec](https://github.com/toml-lang/toml/blob/1.0.0/toml.md#array) this should be fine, but I think maybe some toml parsers struggle with this syntax?

src/static_api.rs
dtolnay: Fixed.

src/validate.rs
dtolnay: Done. Good call.

dtolnay:
Updated to set roles inline in the team's existing `members` array, as you suggested.

src/schema.rs
```diff
@@ -521,6 +528,34 @@ pub(crate) struct TeamPeople {
     pub include_all_alumni: bool,
 }
 
+#[derive(serde::Deserialize, Clone, Debug)]
+#[serde(remote = "Self")]
+pub(crate) struct TeamMember {
+    #[serde(rename = "name")]
+    pub github: String,
```
rylev, line 535: I know I suggested calling this "name", but I do wonder if using "github" as the key would be better. This is what it's called in the people toml config whereas "name" is used for the person's preferred name.

docs/toml-schema.md
```diff
@@ -133,6 +139,14 @@ zulip-stream = "t-lang"
 # Default is 0.
 weight = -100
 
+# Customized roles held by a subset of the team's members, beyond "Team leader"
+# which is rendered automatically for members of the `leads` array.
+[[website.roles]]
```
rylev, line 144: Namespacing these roles under website feels a bit strange. I know the reason for this PR is to display this information on the website, but is this information only relevant to the website? Perhaps this should just be a top level `roles` array?

docs/toml-schema.md
```diff
@@ -49,6 +49,12 @@ members = [
     "rfcbot",
     "craterbot",
     "rust-timer",
+    # Any subset of members may hold custom roles, beyond "Team leader" which is
+    # controlled by the `leads` array above. See [[website.roles]]. Members with
+    # roles are written using an inline table as follows. A simple string member
+    # like "bors" is equivalent to {name = "bors", roles = []}.
```
rylev, line 55: Can we mention that the roles listed with each member must also be present in the roles array?

rylev (commented):
Awesome! I think we can merge as is as the static API looks good (and that's the only thing that is somewhat tricky to change after the fact). However, I'd like to get your opinion on my comments on the schema.

docs/toml-schema.md
dtolnay: Fixed.

docs/toml-schema.md
dtolnay: Done.

src/schema.rs
```diff
@@ -601,6 +641,13 @@ impl WebsiteData {
     }
 }
 
+#[derive(serde_derive::Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct MemberRole {
```
dtolnay, line 646: `MemberRole` in contrast with `DiscordRole`. Not sure if this is a great name. `Role` did not seem adequate.

https://github.com/rust-lang/team/blob/57e50ecf446550dc0ef48f8e48bcc3b8abada10e/src/schema.rs#L465-L468

rylev (approved):
I think this is a great start! Let's merge this and we can adjust from here.

@dtolnay seems like there might be conflicts that need to be resolved first though.

dtolnay:
Rebased to fix superficial conflict with #1164 in src/validate.rs.

rylev (approved):

PR #1156, opened 2023-12-29
Manage rust-lang/docker-rust

sfackler:
Managing the https://github.com/rust-lang/docker-rust repo, and adding @Muscraft and me as the initial maintainers.


jackh726 (approved):
I think this is fine for now, since it's just the status quo - I guess in the future we should think about maybe moving this under infra or something

sfackler:
Sure, no strong feelings on my end about the proper setup.

PR #1158, opened 2024-01-03
Add `crates-io-admins` team

Turbo87:
This is a marker team that defines the group of people that are granted admin access on crates.io. Once the PR is merged we can implement a regular background job on the crates.io side that syncs the team repo into the crates.io user database.

Related:

- https://github.com/rust-lang/crates.io/pull/7852
- https://github.com/rust-lang/crates.io/pull/7858

### DIFF
--- a/teams/crates-io-admins.toml
+++ b/teams/crates-io-admins.toml
@@ -1,0 +1,10 @@
+# This team is regularly synced to the crates.io database and defines who has
+# access to admin functionality on crates.io.
+
+name = "crates-io-admins"
+kind = "marker-team"
+
+[people]
+leads = []
+members = []
+included-teams = ["crates-io"]


teams/crates-io-admins.toml
```diff
@@ -0,0 +1,7 @@
+name = "crates-io-admins"
```
rylev, line 1: It might be helpful to write a comment on what this team is supposed to be the admins of. It took me a second to realize that this was supposed to be for administration of crates.io itself even if in retrospect that is pretty obvious.

Turbo87: I've added a small comment. Does that match what you had in mind?

rylev: Perfect!

rylev (approved):

PR #1161, opened 2024-01-04
Archive project-edition-2021

ehuss:
This group has been essentially replaced by project-edition-2024.

### DIFF
--- a/teams/archive/project-edition-2021.toml
+++ b/teams/archive/project-edition-2021.toml
@@ -3,16 +3,13 @@ subteam-of = "core"
 kind = "project-group"
 
 [people]
-leads = [
-    "nikomatsakis",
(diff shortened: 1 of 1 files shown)


apiraino:
Just a passing comment. Could the `project-edition` team lose the suffix `-year` and be recycled at every edition, just updating the members? :slightly_smiling_face: maybe that could reduce some organizational chores (like creating places for discussions)

(maybe now it's too late or this was already discussed and discarded)

ehuss:
I don't remember if there was a reason to have the year in the name. It would probably be fine to drop it. I don't know if sync-team can support renaming, though.

m-ou-se (approved):

rylev (approved):

PR #1164, opened 2024-01-06
Add a branch protection option to choose required approval count

Kobzol:
Before this PR, the default value was implicitly 0 if bors is used or 1 if bors isn't used. However, for some repositories it's enough to require PRs, but approvals are not needed. Setting `required-approvals = 0` will allow that (once it's implemented in `sync-team`).

r? @Mark-Simulacrum :)

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -292,4 +292,8 @@ ci-checks = ["CI"]
 # merging into this branch require another review. 
 # (optional - default `false`)
 dismiss-stale-review = false
+# How many approvals are required for a PR to be merged.
+# This option is only relevant if bors is not used.
+# (optional - default `1`)
+required-approvals = 1
 ```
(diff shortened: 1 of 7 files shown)


docs/toml-schema.md
```diff
@@ -292,4 +292,8 @@ ci-checks = ["CI"]
 # merging into this branch require another review. 
 # (optional - default `false`)
 dismiss-stale-review = false
+# How many approvals are required for a PR to be merged.
+# This option is only relevant if bors is not used.
```
Mark-Simulacrum, line 296: Can we enforce/validate that it is not set when bors is used?

Kobzol: Sure, I'll add validation. Using bors is already quite funky though, since it's set globally per repo and we don't distinguish which branch (protection) is managed by it.

Kobzol: Done.

Mark-Simulacrum (approved):

PR #1165, opened 2024-01-06
Add support for repo archiving

Kobzol:
Archived repos are in `repos/archive`. Currently there is no validation that the repo contains no team access, I can add it if you want.

### DIFF
--- a/src/static_api.rs
+++ b/src/static_api.rs
@@ -36,7 +36,13 @@ impl<'a> Generator<'a> {
 
     fn generate_repos(&self) -> Result<(), Error> {
         let mut repos: IndexMap<String, Vec<v1::Repo>> = IndexMap::new();
-        for r in self.data.repos() {
+        let repo_iter = self
+            .data
+            .repos()
+            .map(|repo| (repo, false))
+            .chain(self.data.archived_repos().map(|repo| (repo, true)));
+
+        for (r, archived) in repo_iter {
             let branch_protections: Vec<_> = r
                 .branch_protections
                 .iter()
@@ -98,6 +104,7 @@ impl<'a> Generator<'a> {
                     })
                     .collect(),
                 branch_protections,
+                archived,
             };
 
             self.add(&format!("v1/repos/{}.json", r.name), &repo)?;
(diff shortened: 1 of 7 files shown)


src/static_api.rs
```diff
@@ -98,6 +104,7 @@ impl<'a> Generator<'a> {
                     })
                     .collect(),
                 branch_protections,
+                archived,
```
rylev, line 107: Hmmm... Do we want to include archived repos in the repos endpoint or continue to only show current repos and have a dedicared `archived_repos` endpoint?

I think I would prefer the dedicated endpoint approach.

rylev (changes requested):

src/static_api.rs
Kobzol: Hmm, we don't even have an endpoint for archived teams. So maybe we should not even expose the archived repos as an endpoint?

ehuss: Not including the archived status may cause issues for sync-team, since it won't know the difference between a normal repo, an archived repo, and a repo that is not yet tracked.

(IMHO, I think a flag is sufficient since this only really matters for sync-team, and it is also similar to the way GitHub's API works.)

Kobzol: Uhh, of course, we need to expose that information, that's the whole point of this PR, silly me :) The flag will definitely make both `team` and `sync-team` much simpler.

rylev (approved):

src/static_api.rs
Kobzol: I guess that I don't consider it to be that fundamental :thinking: For example, we might also want to manage private repos in the future, should these also have a separate endpoint in your opinion?

Kobzol: That raises a good point. Should archived repos be interactable and have any properties set at all? E.g., what happens if you change a branch protection for an archived repo? Should that do something? (Will that unarchive the repo automatically?).

rylev: I was definitely working under the assumption that archived repos are in a read-only state (more or less by definition), and in order to change them, you would need to unarchive them.

ehuss: I think it is fine if sync-team makes changes to archived repos. It won't unarchive them, and the API seems to be fine with making changes.

ehuss: To clarify: I don't think it matters much. There might be small things, for example the description. Someone may want to change that, and I think it would be kinda fussy to have to unarchive, change, and then re-archive. But I don't expect that to happen much, so I think it probably doesn't matter.

PR #1171, opened 2024-01-06
Add the possibility to mark a repository as private

Kobzol:
This PR adds a `private-non-synced` field to the TOML schema. It should serve for documentation purposes to mark private repositories, but it actually won't be managed by automation (`sync-team`), because the action of making a repository private is destructive.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -259,6 +259,10 @@ description = "A repo for awesome things!"
 homepage = "https://www.rust-lang.org/"
 # The bots that this repo requires (required)
 bots = ["bors", "rustbot", "rust-timer"]
+# Should the repository be private? (optional - default `false`)
+# Note that this only serves for documentation purposes, it is
+# not synchronized by automation.
+private-non-synced = false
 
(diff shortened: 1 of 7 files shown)


Kobzol:
On second thought, maybe this isn't such a good idea, because making a repo private is an incredibly destructive action, and it would be bad if it has happened by accident. Maybe we should leave this one only for manual setting?

rylev (dismissed):
I have avoided making any changes to repos that are not 100% reversible. I think we should at the very least wait until we're much more confident in the repository syncing infrastructure before making this change.

Kobzol:
I think that making a repo private/public is *very rare*, and shouldn probably just be done manually. I would be quite scared to have this managed by automation. I will change my implementation to automate archiving instead, that should be less destructive, and we need it more.

jackh726:
I think we should let @Mark-Simulacrum take the review for this one, it's definitely more infra-y.

Mark-Simulacrum (approved):
I think having this documented doesn't hurt.

PR #1173, opened 2024-01-08
Add project group for const traits initiative

compiler-errors:
This PR adds a project group for the "const traits" initiative.

@fee1-dead and a few others have gathered around an effort to enable const traits in the language (rust-lang/rfcs#2632). The RFC never landed, but work has been under experimentation (https://github.com/rust-lang/rfcs/pull/2632#issuecomment-874924659), and has gained speed especially recently since all of the work was ripped out and reimplemented under the new `#![feature(effects)]`.

Adding this PG helps to acknowledge some of the people who work on this feature, and allows us to have a single ping group for important PRs, etc.

This PG is related to the keyword generics initiative (cc @rust-lang/initiative-keyword-generics) in that it's implementing a "const" effect, but I'd like for this to be under the stewardship of the compiler team (or ideally, types team, but that's not possible), since it's primarily focused around practical implementation work and implementation concerns (allowing users to implement and call `const` traits), rather than theoretical work of designing an over-arching feature of the language like the kw-generics initiative aims to do.

### DIFF
--- a/teams/project-const-traits.toml
+++ b/teams/project-const-traits.toml
@@ -1,0 +1,16 @@
+name = "project-const-traits"
+kind = "project-group"
+subteam-of = "compiler"
+
+[people]
+leads = [
+    "fee1-dead",
+]
+members = [
+    "compiler-errors",
+    "fee1-dead",
+    "fmease",
+    "oli-obk",
+]
+alumni = []
+


compiler-errors:
IDK if this needs an MCP or anything

r? @wesleywiser @davidtwco

Also this is less "starting up a new project group" and more "reflecting the existing set of people working on a specific initiative". If this PG didn't exist, the work would still get done, but of course I'd be a bit sadder.

oli-obk:
> it's primarily focused around practical implementation work and implementation concerns (allowing users to implement and call `const` traits), rather than theoretical work of designing an over-arching feature of the language like the kw-generics initiative aims to do.

I agree with this. the effects initiative is about the language level design, while this project group is for the implementation work in the type system. I don't believe there is much of an issue if the language design changes from what the impl does, and we need to adjust the impl. That part is mostly at the syntax and lowering level, and not what made implementing effects a hard thing

yoshuawuyts:
Fair enough; Oli and I also talked offline a bit and we should be good for now with Oli serving as a bridge between both groups. Withdrawing my concern; excited for const traits!

davidtwco (approved):
This makes sense to me!

fee1-dead:
> experimentation with `~const Trait` bounds was approved as part of the work on _const_ generics

I don't think const traits is related to const generics. They are two separate initiatives. The approval for experimentation is [here](https://github.com/rust-lang/rfcs/pull/2632#issuecomment-567699174)

fee1-dead:
> And relegates the Effects Initiative to only being about a theoretical future unifying design; not what's currently being implemented. To me this shows that we may already be experiencing communication issues, which wouldn't be helped by formalizing a further separation.

I'm happy to chat about this on Zulip, and I agree that communication needs to be better. AFAIK, the implementation has been for const traits since the beginning, with the implementation design allowing it to be easier to implement other effects as well. Given the narrow focus of our current progress, and that no concrete proposals for how other syntaxes would work (like `~async`), it appears to me that we're not just implementing the entirety of keyword-generics, and "implementing const traits with the design of effects/keyword generics in mind" seems closer to what we are actually doing.

Anyways, it would be nice if we continued this discussion elsewhere, since it is irrelevant to the formation of this project group for now.

rylev (approved):
It seems there was some discussion on whether this group should exist at all. It would be probably be better in the future for such discussion to happen elsewhere so that this repo can be used for more mundane particularities.

That being said, it seems that @davidtwco has approved this PR so it can be merged. I'll wait a bit to make sure I didn't misunderstand where this discussion landed.

wesleywiser (approved):

fmease:
Could somebody with the right set of permissions create a Zulip stream, **#t-compiler/project-const-traits** I guess? Thanks in advance :)

PR #1175, opened 2024-01-08
Add  `infra = "triage"` permissions to `team`

Kobzol:
Members of the `infra` team (like me :) ) do work on this repository, so it would be nice to be able to ask for reviews.

### DIFF
--- a/repos/rust-lang/team.toml
+++ b/repos/rust-lang/team.toml
@@ -7,6 +7,7 @@ bots = []
 core = "admin"
 mods = "maintain"
 team-repo-admins = "write"
+infra = "triage"
 
 [[branch-protections]]
 pattern = "master"


rylev (approved):

jackh726:
This should probably go a step further and be maintain, since per policy infra team is in charge of the "infra" code here, whereas team-repo-admins is just the people/teams/repos

Kobzol:
The problem is that we can't distinguish permissions based on subdirectories. And all of infra probably shouldn't have maintain here (?), since changes to this repo can have a lot of security implications for the whole of rust-lang.

jackh726:
I mean, you could say the same for team-repo-admins :)

Kobzol:
That team has seemingly been created exactly to oversee the permissions of the whole rust-lang org. If it was deemeed ok to give all of infra these permissions, I suppose that a separate team wouldn't be created, but I don't know the history :)

@Mark-Simulacrum @rylev thoughts?

Mark-Simulacrum:
I don't think we should give all of infra permissions in excess of hard need in this repo, similar to how not all of infra has full access to AWS or other sensitive accounts.

PR #1177, opened 2024-01-08
add calendar repositories

davidtwco:
The compiler team currently use Google Calendar to manage our team calendar, and that's a bit annoying because we can't manage the access controls to that in the same way we do everything else - it requires manually adding team member's Google account (assuming they have one) to the calendar. 

As an alternative to this, I've implemented a small utility for generating calendar files from TOML and hosting them in GitHub Pages. This makes it easy to get a link that can be subscribed to in a user's calendar application of choice, and it can be managed using the same workflows and tooling we have for everything else. I've written this in such a way that other teams could easily add their own calendars, so that this becomes a standard mechanism within the project, and that the general calendar generation can be re-used easily by other projects too.

- `calendar-generation` contains a utility that will generate an iCalendar document from TOML files containing the events. This repository already exists at [davidtwco/toml-to-ical](https://github.com/davidtwco/toml-to-ical) and if we are to use it, it should be moved so that I can continue to maintain it as the project uses it, and can be maintained by the rest of the project if necessary.
- `calendar` contains TOML files and CI configuration to use `calendar-generation` to generate and then host the calendar documents on GitHub Pages. A draft of this repository exists at [davidtwco/rust-calendars-draft](https://github.com/davidtwco/rust-calendars-draft), replicating the current events in the compiler team's calendar.


repos/rust-lang/calendar-generation.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[access.teams]
+compiler = "write"
+infra = "write"
+
+[[branch-protections]]
+pattern = "main"
+ci-checks = ['continuous integration']
+dismiss-stale-review = false
+pr-required = true
```
Kobzol, line 14:
```
pr-required = true
review-required = true
```
are not actually valid TOML entries for `toml`, it's just an artifact of my extraction script.

`dismiss-stale-review = false` is the default option, so it doesn't have to be used here.

davidtwco: Thanks, fixed.

repos/rust-lang/calendar.toml
```diff
@@ -0,0 +1,20 @@
...
+compiler = "maintain"
+compiler-contributors = "maintain"
+crates-io = "maintain"
+docs-rs = "maintain"
+infra = "maintain"
+lang = "maintain"
+leadership-council = "maintain"
+libs = "maintain"
+release = "maintain"
+triagebot = "maintain"
```
rylev, line 17: This shouldn't be necessary with `bots = ["rustbot"]`

repos/rust-lang/calendar.toml
```diff
@@ -0,0 +1,20 @@
+org = "rust-lang"
+name = "calendar"
+description = "Calendars for Rust project teams"
+bots = ["rustbot"]
+
+[access.teams]
+community = "maintain"
```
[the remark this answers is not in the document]
davidtwco: Fixed - had just copied the list from the forge repo.

repos/rust-lang/calendar.toml
davidtwco: Fixed - had just copied the list from the forge repo.

Mark-Simulacrum (approved):
Not sure if a single calendar repo vs. per team makes sense, but I think we can start here.

davidtwco:
Links to created repositories for those following along at home:

- https://github.com/rust-lang/calendar-generation
- https://github.com/rust-lang/calendar

PR #1178, opened 2024-01-08
Resolve clippy lints

dtolnay:
CI does not run `cargo clippy` because https://github.com/rust-lang/team/issues/560. But it's helpful for contributors to be able to run `cargo clippy` on their change without seeing pre-existing warnings.

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -751,7 +751,7 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
                 repo.org
             );
         }
-        for (team_name, _) in &repo.access.teams {
+        for team_name in repo.access.teams.keys() {
             if !github_teams.contains(&(repo.org.clone(), team_name.clone())) {
                 bail!(
                         "access for {}/{} is invalid: '{}' is not configured as a GitHub team for the '{}' org",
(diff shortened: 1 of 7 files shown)


src/validate.rs
```diff
@@ -634,16 +635,14 @@ fn validate_discord_team_members_have_discord_ids(data: &Data, errors: &mut Vec<
         if team.discord_roles().is_some() && team.name() != "all" {
             let team_members = team.members(data)?;
             if team_members.len() != team.discord_ids(data)?.len() {
-                let members: String = team_members
+                let mut missing_discord_id = team_members
                     .into_iter()
-                    .filter(|name| data.person(name).map(|p| p.discord_id()) == Some(None))
-                    .map(|name| format!("{}, ", name))
-                    .collect();
+                    .filter(|name| data.person(name).map(|p| p.discord_id()) == Some(None));
```
Kobzol, line 640: We can just add `.collect::<Vec<_>>();` here to avoid including `itertools` (a new dependency), I think.

dtolnay: Done.

Mark-Simulacrum (approved):

PR #1180, opened 2024-01-08
Add T-lang-ops and WG-async to calendar repo

traviscross:
T-lang uses the calendar managed by the `calendar` repository, and the lang team itself already has the correct permissions here.  However, the T-lang-ops team mostly handles the scheduling for the T-lang meetings, so T-lang-ops needs perms on this repos also.  Let's add those.

Separately, WG-async uses the calendar managed by the `calendar` repository and so should also have permissions.

### DIFF
--- a/repos/rust-lang/calendar.toml
+++ b/repos/rust-lang/calendar.toml
@@ -10,9 +10,11 @@ crates-io = "maintain"
 docs-rs = "maintain"
 infra = "maintain"
 lang = "maintain"
+lang-ops = "maintain"
 leadership-council = "maintain"
 libs = "maintain"
 release = "maintain"
+wg-async = "maintain"
 
 [[branch-protections]]
 pattern = "main"


davidtwco (approved):

repos/rust-lang/calendar.toml
```diff
@@ -10,9 +10,11 @@ crates-io = "maintain"
 docs-rs = "maintain"
 infra = "maintain"
 lang = "maintain"
+lang-ops = "maintain"
 leadership-council = "maintain"
 libs = "maintain"
 release = "maintain"
+wg-async = "maintain"
```
rylev, line 17: @davidtwco is there a reason all of these teams need maintain permissions? It might make sense to bump everyone down to write. The difference between maintain and write is that maintainers can do some configuration of the repo like manage topics and configure a publishing source for GitHub pages (you can find [more info here](https://docs.github.com/en/github-ae@latest/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role)).

I'll merge this to unblock, but perhaps we should follow up with a PR bumping the permissions down?

rylev (approved):

PR #1196, opened 2024-01-11
Introduce v2 permissions API

Turbo87:
The v1 permissions API can be misleading in that the order of the `github_users` does not match the order of the `github_ids` (see https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/team.20repo.20integration/near/412320079).

This change introduces a v2 permissions API which uses a list of objects instead of an object with multiple single-field lists. This allows us to easily show both GitHub login and ID of a person in the logs.

## Example

```json
{
  "people": [
    {
      "name": "Aaron Hill",
      "github": "Aaron1011",
      "github_id": 1408859,
      "discord_id": null
    },
    {
      "name": "Ed Page",
      "github": "epage",
      "github_id": 60961,
      "discord_id": null
    },
    {
      "name": "Jamie Hill-Daniel",
      "github": "clubby789",
      "github_id": 13556931,
      "discord_id": 85400548534145024
    }
  ]
}
```


rust_team_data/src/v2.rs
```diff
@@ -0,0 +1,16 @@
+use serde::{Deserialize, Serialize};
+
+pub static BASE_URL: &str = "https://team-api.infra.rust-lang.org/v2";
```
apiraino, line 3: I'm about to ask something perhaps out of scope in this PR.

Can we please move `BASE_URL` out of the code and into an external env var?

The reason is: when I work on the triagebot and I use my own [test repository](https://github.com/apiraino/test-triagebot), I need to live patch URL to include other users so I can test f.e. PR review assignment.

Turbo87: seems like that env var should live in triagebot though, no? 🤔

rust_team_data/src/v2.rs
```diff
@@ -0,0 +1,16 @@
...
+pub struct Permission {
+    pub people: Vec<PermissionPerson>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub struct PermissionPerson {
+    pub github_id: usize,
+    pub github: String,
+    pub name: String,
+    pub discord_id: Option<usize>,
```
pietroalbini, line 15: Can we remove the discord ID until some tool actually needs it?

rust_team_data/src/v2.rs
pietroalbini: Let's discuss that in a separate PR.

rust_team_data/src/v2.rs
Turbo87: are you sure that no tool is using it? why was it added to the v1 API if nothing is using it? 😅

Turbo87:
@rylev the alternative is to add a `people` key to the existing v1 API response that duplicates all the data, but that feels kinda dirty to me. is there an actual cost to having a v2 API? the users of v1 don't necessarily need to migrate, though with two systems already having implemented this in a buggy way it feels better to make it a v2 without the old cruft alongside the new output.

rylev:
First of all, this isn't a blocking concern per say, but I'd just be remiss not to say that it doesn't sit right with me. 

The perceived down side of a v2 from my perspective is simply the overhead of having to know about and maintain two endpoints. Another alternative would be to return back only the new data if a certain query param is sent. This makes the responses a bit cleaner with the only "hack" being that a query param controls the shape of the response body.

rylev:
🤦 You're right... Disregard that suggestion. My point still stands, but I'll let others weigh in. If I'm the only one with a concern, feel free to merge.

Kobzol:
It seems way simpler to just add the new field though. Like, probably one day we'll have to do a BC break and switch to v2, but this seems like something that can be resolved easily by adding new data instead.

PR #1197, opened 2024-01-11
Add `crates-io-admin` permission

Turbo87:
see https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/team.20repo.20integration/near/412316069

/cc @pietroalbini


jdno (commented):
Should the new permission be documented in https://github.com/rust-lang/team/blob/master/docs/toml-schema.md?

rylev (approved):

PR #1198, opened 2024-01-11
Add book repository under automation

Kobzol:
Repo: https://github.com/rust-lang/book

The current GitHub permissions (see below) don't really contain any reasonable team, so I'm not sure who to add here, since the docs team is archived. I saw @carols10cents being active, so I added access for her.

Extracted from GH:
```
org = "rust-lang"
name = "book"
description = "The Rust Programming Language"
bots = []

[access.teams]
security = "pull"
highfive = "write"

[access.individuals]
carols10cents = "admin"
rust-lang-owner = "admin"
rylev = "admin"
badboy = "admin"
jdno = "admin"
Mark-Simulacrum = "admin"
pietroalbini = "admin"
rust-highfive = "write"
frewsxcv = "write"

[[branch-protections]]
pattern = "main"
ci-checks = []
dismiss-stale-review = false
pr-required = true
review-required = true
```

### DIFF
--- a/repos/rust-lang/book.toml
+++ b/repos/rust-lang/book.toml
@@ -1,0 +1,13 @@
+org = "rust-lang"
+name = "book"
+description = "The Rust Programming Language"
+bots = []
+
+[access.teams]
+
+[access.individuals]
+carols10cents = "maintain"
+
+[[branch-protections]]
+pattern = "main"
+ci-checks = []


carols10cents:
Yes, I am the only maintainer of the book right now.

rylev:
Don't want to block this PR, but it might make sense for us to make a book team and make @carols10cents the only member of that team.

rylev (approved):

Kobzol:
> Don't want to block this PR, but it might make sense for us to make a book team and make @carols10cents the only member of that team.

What should be the parent team of a `book` team?

rylev:
That's a good question - it feels a bit awkward but landing pad is specifically set up for teams that don't have good homes otherwise. If the choice is not making a team at all even though there should be one or putting that team under landing pad, I'd personally prefer seeing it under landing pad.

PR #1200, opened 2024-01-13
Create a repos entry for bindgen

tgross35:
This sets up https://github.com/rust-lang/rust-bindgen to be able to use rustbot. This was okayed at https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Bindgen.20triage.

cc @emilio

### DIFF
--- a/repos/rust-lang/rust-bindgen.toml
+++ b/repos/rust-lang/rust-bindgen.toml
@@ -1,0 +1,19 @@
+org = "rust-lang"
+name = "rust-bindgen"
+description = "Automatically generates Rust FFI bindings to C (and some C++) libraries."
+bots = ["rustbot"]
+
+[access.teams]
+wg-bindgen = "write"
+
+[access.individuals]
+emilio = "admin"
+fitzgen = "admin"
+pvdrz = "write"
+
+[[branch-protections]]
+pattern = "main"
+required-approvals = 0
+ci-checks = [
+    "test",
+]


Kobzol:
I would perhaps

Extracted from GH:
```
org = "rust-lang"
name = "rust-bindgen"
description = "Automatically generates Rust FFI bindings to C (and some C++) libraries."
bots = []

[access.teams]
security = "pull"

[access.individuals]
highfive = "write"
fitzgen = "admin"
Mark-Simulacrum = "admin"
badboy = "admin"
kulp = "write"
rylev = "admin"
Manishearth = "admin"
scoopr = "write"
jdno = "admin"
pvdrz = "write"
rust-lang-owner = "admin"
amanjeev = "write"
emilio = "admin"
pietroalbini = "admin"

[[branch-protections]]
pattern = "main"
ci-checks = []
dismiss-stale-review = false
pr-required = false
review-required = false
```

The branch protection should be set to the `main` branch, and I would also use `required-approvals = 0`, to keep the behavior as it was before automation (PRs are required, but PR approvals are not required).

Kobzol:
The CI check name seems to be `text` (https://github.com/rust-lang/rust-bindgen/blob/main/.github/workflows/bindgen.yml#L136C5-L136C5).

Kobzol:
Ah, sorry, I have probably confused you. The TOML that I have posted here is not completely valid and shouldn't be reproduced 1:1. It's just a helper output that I use to decide what teams/permissions we should use for newly managed repos. Your previous version was actually OK, just the branch name and CI check name was wrong.

tgross35:
Oh, whoops. I changed it back. I thought that this file would remove owners if not listed, but sounds like that is not the case?

Kobzol:
I'm not sure what do you mean by "owners", but the preferred way to manage repos is to only allow access to teams, not individuals (unless there is some edge case).

tgross35:
I meant admin, not owner. I left three individual entries

```toml
emilio = "admin"
fitzgen = "admin"
pvdrz = "write"
```

Since emilio and fitzgen are the primary maintainers and admins, and pvdrz hasn't yet gotten added to the wg for write access (I requested they submit a PR).

I assume the other admins get added implicitly. Happy to change as needed

Kobzol:
I see. I *think* that we don't want to keep explicit admin accesses, and lower everything to write/maintain, but I'm not sure. CC @Mark-Simulacrum

Mark-Simulacrum:
If possible, yes. Admin actions should be rare and my hope is that either infra-admins or team repo PRs are the way to handle those.

Kobzol:
Implemented in https://github.com/rust-lang/team/pull/1259.

PR #1204, opened 2024-01-13
chore: Add T-cargo to calendar repo

Muscraft:
`T-cargo` uses the [calendar managed by the calendar repository](https://github.com/rust-lang/calendar/blob/62949e2748910a998d4bc15cd42fe3a3ad692bc0/cargo.toml), but does not have permission to edit the calendar. This adds `write` permissions for `T-cargo` as [suggested by this comment](https://github.com/rust-lang/team/pull/1180#discussion_r1446057760).

### DIFF
--- a/repos/rust-lang/calendar.toml
+++ b/repos/rust-lang/calendar.toml
@@ -4,6 +4,7 @@ description = "Calendars for Rust project teams"
 bots = ["rustbot"]
 
 [access.teams]
+cargo = "maintain"
 compiler = "maintain"
 compiler-contributors = "maintain"
 crates-io = "maintain"


repos/rust-lang/calendar.toml
```diff
@@ -4,6 +4,7 @@ description = "Calendars for Rust project teams"
 bots = ["rustbot"]
 
 [access.teams]
+cargo = "write"
```
compiler-errors, line 7:
```suggestion
cargo = "maintain"
```

consistent with all the other teams with hosted calendars

Muscraft: I did this because it was suggested [in this comment](https://github.com/rust-lang/team/pull/1180#discussion_r1446057760). I don't think `T-cargo` will use anything in `maintain`, we really only need `write`.

rylev: It's fine to keep things consistent and give cargo maintain privileges. We can then open a PR bumping everyone down to write.

rylev (approved):

PR #1206, opened 2024-01-14
Add launching-pad team.

ehuss:
This adds the launching-pad team, and connects all of its subteams into it. This should help with tracking the team membership and their status and connections.

cc @jonathanpallant

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -185,15 +185,6 @@ fn validate_subteam_of(data: &Data, errors: &mut Vec<String>) {
                 );
             };
 
-            if !matches!(team.kind(), TeamKind::Team) && parent.subteam_of().is_some() {
-                bail!(
-                    "{} `{}` can't be a subteam of a subteam (`{}`)",
-                    team.kind(),
-                    team.name(),
-                    parent.name(),
-                );
-            }
-
             team = parent;
         }
         Ok(())
--- a/teams/community.toml
+++ b/teams/community.toml
@@ -1,4 +1,5 @@
 name = "community"
+subteam-of = "launching-pad"
 
 [people]
 leads = []
--- a/teams/docker.toml
+++ b/teams/docker.toml
@@ -1,4 +1,5 @@
 name = "docker"
+subteam-of = "launching-pad"
 
 [people]
 leads = []
--- a/teams/launching-pad.toml
+++ b/teams/launching-pad.toml
@@ -1,0 +1,7 @@
+name = "launching-pad"
+
+[people]
+# This is an "umbrella" team, and thus does not have any members.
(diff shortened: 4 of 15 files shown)


ehuss:
Opening as a draft because this will break the website until https://github.com/rust-lang/www.rust-lang.org/pull/1914 is merged and deployed.

In the future, I would like to make it so that subteam-of will be required, but we aren't quite there yet. One wrinkle is keeping track of the top-level teams in a way that makes sense. If they are subteams of the leadership-council, that will likely also cause problems for the website (and I'm not sure what else).

src/validate.rs
```diff
@@ -185,15 +185,6 @@ fn validate_subteam_of(data: &Data, errors: &mut Vec<String>) {
...
 
-            if !matches!(team.kind(), TeamKind::Team) && parent.subteam_of().is_some() {
-                bail!(
-                    "{} `{}` can't be a subteam of a subteam (`{}`)",
-                    team.kind(),
-                    team.name(),
-                    parent.name(),
-                );
-            }
-
```
ehuss, line 196: I don't 100% understand why this clause was left in #1046. @fmease Can you say why this is here?

rylev:
@ehuss with https://github.com/rust-lang/www.rust-lang.org/pull/1914 merged, is this ready for review?

src/validate.rs
fmease: Sorry, I didn't get notified by the ping, maybe because this PR is marked as a draft?

fmease: I kept this check because it's not clear where to place WGs and PGs that are children of subteams on the website. WGs and PGs are placed at the bottom of the page inside new distinct sections. I doesn't feel right to place WGs/PGs that aren't actually direct children of the team whose page we are looking at at the same level as other WGs/PGs especially since the user cannot draw any connection between the parent team and the groups.

fmease: For example:

```
Team A
+-- Team A-A
|   +-- Working Group A-A-A
+-- Team A-B
```

The webpage of team A would look like:

```
Team A
--- Subteams
Team A-A
Team A-B
--- WGs
Working Group A-A-A
--- PGs
```

Now it's unclear if WG A-A-A is a direct child of A, A-A or A-B.

ehuss:
Review is fine, but it cannot be merged as the deploy is currently not working, see: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/www.2Erust-lang.2Eorg.20deploy.20is.20failing/near/412879305

ehuss:
Now that the website has been deployed, this should be ready.

rylev (approved):

PR #1211, opened 2024-01-15
Add regex repo under automation

Kobzol:
Repo: https://github.com/rust-lang/regex

CC @BurntSushi

Extracted from GH:
```
org = "rust-lang"
name = "regex"
description = "An implementation of regular expressions for Rust. This implementation uses finite automata and guarantees linear time matching on all inputs."
bots = []

[access.teams]
regex = "maintain"

[access.individuals]
badboy = "admin"
bors = "write"
jdno = "admin"
BurntSushi = "maintain"
pietroalbini = "admin"
Mark-Simulacrum = "admin"
rust-lang-owner = "admin"
rylev = "admin"
```

### DIFF
--- a/repos/rust-lang/regex.toml
+++ b/repos/rust-lang/regex.toml
@@ -1,0 +1,12 @@
+org = "rust-lang"
+name = "regex"
+description = "An implementation of regular expressions for Rust. This implementation uses finite automata and guarantees linear time matching on all inputs."
+bots = []
+
+[access.teams]
+regex = "maintain"
+
+[[branch-protections]]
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
I think this LGTM.

Note that the regex repo hasn't used bors for years. (And likely never will again.)

Kobzol:
> Note that the regex repo hasn't used bors for years. (And likely never will again.)

That is fine, this configuration doesn't enable bors permissions :+1:

rylev (approved):

PR #1218, opened 2024-01-16
Add this-week-in-rust repo under automation

Kobzol:
Repo: https://github.com/rust-lang/this-week-in-rust

CC @nellshamrell

Extracted from GH:
```
org = "rust-lang"
name = "this-week-in-rust"
description = "Data for this-week-in-rust.org"
bots = []

[access.teams]
twir-reviewers = "maintain"
security = "pull"

[access.individuals]
mariannegoldin = "maintain"
ericseppanen = "maintain"
rust-lang-owner = "admin"
rylev = "admin"
U007D = "maintain"
pietroalbini = "admin"
nellshamrell = "maintain"
extrawurst = "maintain"
Mark-Simulacrum = "admin"
bennyvasquez = "maintain"
cdmistman = "maintain"
badboy = "admin"
jdno = "admin"
andrewpollack = "maintain"

[[branch-protections]]
pattern = "master"
ci-checks = []
dismiss-stale-review = false
pr-required = false
review-required = false
```

### DIFF
--- a/repos/rust-lang/this-week-in-rust.toml
+++ b/repos/rust-lang/this-week-in-rust.toml
@@ -1,0 +1,12 @@
+org = "rust-lang"
+name = "this-week-in-rust"
+description = "Data for this-week-in-rust.org"
+bots = []
+
+[access.teams]
+twir-reviewers = "write"
+twir = "write"
+
+[[branch-protections]]
+pattern = "master"
+required-approvals = 0


ehuss:
Seems reasonable to me to just add a GitHub team whenever needed for these repo PRs.

Kobzol:
Ok, created https://github.com/rust-lang/team/pull/1226.

repos/rust-lang/this-week-in-rust.toml
```diff
@@ -0,0 +1,12 @@
+org = "rust-lang"
+name = "this-week-in-rust"
+description = "Data for this-week-in-rust.org"
+bots = []
+
+[access.teams]
+twir-reviewers = "maintain"
+twir = "maintain"
```
rylev, line 8: Same point as the other PR about the need for maintain - `admin` or `write` should be enough

Kobzol: Changed. I think that `maintain` can be used to push directly to protected branches, I wonder if that is used on some of these repos where CI is not set up and/or PRs are not used so commonly.

rylev: You're right. I also realized that I was switching `maintain` and `admin` in my head. So `maintain` actually likely is the appropriate permission in cases where we want to allow the maintainers to more-or-less do their own thing while `admin` is likely never appropriate.

rylev (approved):

repos/rust-lang/this-week-in-rust.toml
Kobzol: Yeah, I think that `maintain` is/was used in the cases where the teams/contributors had a large degree of autonomy over the repo. That being said, I think that it's not a bad idea to proactively lower the permissions to `write` and only allow `maintain` if someone really requires it.

PR #1182, opened 2024-01-08
Add team affiliation for leadership council representatives

dtolnay:
Screenshot from 2024-01-08 13-50-37

PR #1228, opened 2024-01-18
Implement RFC#3533 Combining Infra and Release

rylev:
This implements [RFC#3533](https://github.com/rust-lang/rfcs/pull/3533).

Concretely this means:
* Make Release a subteam of Infra
* Remove @rylev from Leadership Council
  * While @Mark-Simulacrum was the Release representative on the Council, it was agreed that as part of this move @rylev would step down as Infra representative and @Mark-Simulacrum would take over as the Infra rep

There are a few follow ups that need to happen as a result of implementing RFC 3533 that cannot be addressed in this PR:
* The launching pad needs to be created so wg-triage can be placed underneath it. This is being worked on in https://github.com/rust-lang/team/pull/1206 cc @ehuss 
* https://github.com/rust-lang/team/pull/1182 needs to be updated to reflect the change in roles (i.e., the @rylev is no longer a Council member, that the Release team no longer has direct representation, and that @Mark-Simulacrum is the new representative for Infra) cc @dtolnay

### DIFF
--- a/teams/leadership-council.toml
+++ b/teams/leadership-council.toml
@@ -10,10 +10,9 @@ members = [
     "jonathanpallant",
     "Mark-Simulacrum",
     "m-ou-se",
-    "rylev",
     "technetos",
 ]
-alumni = ["khionu"]
+alumni = ["khionu", "rylev"]
 
 [[github]]
 orgs = ["rust-lang", "rust-lang-nursery"]
--- a/teams/release.toml
+++ b/teams/release.toml
@@ -1,4 +1,5 @@
 name = "release"
+subteam-of = "infra"
 
 [people]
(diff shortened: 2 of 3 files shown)


jackh726 (approved):
Thank you @rylev for your work on the council!

dtolnay:
> https://github.com/rust-lang/team/pull/1182 needs to be updated to reflect the change in roles (i.e., that **<span>@</span>rylev** is no longer a Council member, that the Release team no longer has direct representation, and that **<span>@</span>Mark-Simulacrum** is the new representative for Infra) cc **<span>@</span>dtolnay**

Thanks for the ping. I updated #1182 to reflect these changes.

PR #1233, opened 2024-01-19
Make the infra-admins team a marker team

rylev:
This team is only used for permissions purposes and thus is more appropriate as a marker team.


ehuss:
I had a few other teams that I wanted to switch to markers which I opened in #1234.

rylev:
Superseded by https://github.com/rust-lang/team/pull/1234

PR #1254, opened 2024-01-22
Change leads of the community survey team to Kobzol and apiraino

Kobzol:
I have discussed the leadership change with rylev and apiraino, and I think that there's no issue with having two leads. The more, the merrier :grin:

CC @rylev @apiraino


rylev (approved):
Amazing! 🎉

PR #1279, opened 2024-01-22
Make sure that repository name is the same as its TOML filename

Kobzol:
For example, the `spec` repo is located in a `specification.toml` file, which is not very intuitive. Upholding this property also makes sure that the repo names in a given organization are unique (until we implement archiving, which might put archives repos in a different directory).


Kobzol:
(Force-pushing to rerun CI).

PR #1282, opened 2024-01-23
Allow specifying push allowances for branch protection

Kobzol:
This enables to allow only some teams to push/merge to specific branches, to make branch protection more fine-grained.

`sync-team` part of this change: https://github.com/rust-lang/sync-team/pull/66

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -306,4 +306,9 @@ dismiss-stale-review = false
 # This option is only relevant if bors is not used.
 # (optional - default `1`)
 required-approvals = 1
+# Which GitHub teams have access to push/merge to this branch.
+# If unspecified, all teams/contributors with write or higher access
+# can push/merge to the branch.
+# (optional)
+allowed-merge-teams = ["awesome-team"]
 ```
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -778,9 +778,24 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
 
 /// Validate that branch protections make sense in combination with used bots.
 fn validate_branch_protections(data: &Data, errors: &mut Vec<String>) {
+    let github_teams = data.github_teams();
+
     wrapper(data.repos(), errors, |repo, _| {
         let bors_used = repo.bots.iter().any(|b| matches!(b, Bot::Bors));
         for protection in &repo.branch_protections {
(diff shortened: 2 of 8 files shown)


src/validate.rs
```diff
@@ -778,16 +778,40 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
...
+                }
+                if !protection.allowed_merge_teams.is_empty() {
+                    bail!(
+                        r#"repo '{}' uses bors and its branch protection for {} uses the `allowed-merge-teams` attribute;
+please remove the attribute when using bors"#,
+                        repo.name,
+                        protection.pattern,
+                    );
+                }
+            }
```
rylev, line 803: Note: it was hard to tell that the check below this was not in the `if bors_used` block. Not sure if moving things out to separate functions would actually make things better so it might just be something we need to stick with, but I thought I'd mention it.

docs/toml-schema.md
```diff
@@ -306,4 +306,9 @@ dismiss-stale-review = false
 # This option is only relevant if bors is not used.
 # (optional - default `1`)
 required-approvals = 1
+# Which teams have access to push/merge to this branch.
```
rylev, line 309: It's unclear from these instructions that these are *GitHub* teams and not the names of teams in the team repo. I think I'd prefer for it to be team repo teams and for us to do the mapping on behalf of the user, but in any case, we should document which is meant.

rylev (changes requested):

src/validate.rs
Kobzol: I moved the check above the bors check, that should make it clearer.

docs/toml-schema.md
Kobzol: Teams and GitHub teams are not 1:1, although we could probably put all GH teams of the team in there. But we already use GitHub teams for `[access.teams]`, so I think that we should keep it as that to be consistent. I modified the documentation.

rylev (approved):

jackh726:
Seems like an alternative API here would be to allow the teams under `access.teams` to be able to specify allowing merging.

Kobzol:
The access is per branch though, so since we already have a per-branch setting (branch protections), I think that it's better to specify it on the branch. It's more natural to identify teams by name than to identify branch protections by their pattern, IMO.

PR #1283, opened 2024-01-23
Add command for dumping users with individual access

Kobzol:
Proposed by @rylev [here](https://github.com/rust-lang/team/pull/1271#issuecomment-1905699241).

Allows us to see how many individual accesses there are (we want to eventually get rid of these).

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,17 @@ cargo run dump-website
 
 The website will automatically load new teams added here, however they cannot be translated unless `tools.ftl` is also updated.
 
+You can also print a list of users with individual access to repositories
+
+```
+# Group the accesses by repository
+cargo run dump-individual-access --group-mode repo
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -102,6 +102,13 @@ cargo run dump-website
 
 The website will automatically load new teams added here, however they cannot be translated unless `tools.ftl` is also updated.
 
+You can also print a list of users with individual access to repositories
+
+```
+cargo run dump-individual-access
```
rylev, line 108: I think this is fine for now, but I have some other ideas for views on repos that we might want to do that makes me think we might want to think about a repo subcommand:

```
cargo run repo dump-individual-access
```

Kobzol:
Added a flag to choose how to grouop the accesses (`repo`/`person`).

rylev (approved):

PR #1290, opened 2024-01-24
Add `people` field to the `permissions` API

Turbo87:
This PR is similar to https://github.com/rust-lang/team/pull/1196, but adds the new fields to the v1 API instead of introducing a new API version.

---

The v1 permissions API can be misleading in that the order of the `github_users` does not match the order of the `github_ids` (see https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/team.20repo.20integration/near/412320079).

This change introduces a new `people` field on the `permissions` API which uses a list of objects instead of an object with multiple single-field lists. This allows us to easily show both GitHub login and ID of a person in the logs.

## Example

```json
{
  "people": [
    {
      "name": "Aaron Hill",
      "github": "Aaron1011",
      "github_id": 1408859,
      "discord_id": null
    },
    {
      "name": "Ed Page",
      "github": "epage",
      "github_id": 60961,
      "discord_id": null
    },
    {
      "name": "Jamie Hill-Daniel",
      "github": "clubby789",
      "github_id": 13556931,
      "discord_id": 85400548534145024
    }
  ]
}
```

Closes https://github.com/rust-lang/team/pull/1196

r? @rust-lang/infra

### DIFF
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -125,9 +125,17 @@ pub struct ZulipGroups {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Permission {
+    pub people: Vec<PermissionPerson>,
     pub github_users: Vec<String>,
     pub github_ids: Vec<usize>,
     pub discord_ids: Vec<usize>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub struct PermissionPerson {
+    pub github_id: usize,
+    pub github: String,
+    pub name: String,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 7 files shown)


rust_team_data/src/v1.rs
```diff
@@ -125,11 +125,20 @@ pub struct ZulipGroups {
...
     pub github_ids: Vec<usize>,
     pub discord_ids: Vec<usize>,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub struct PermissionPerson {
+    pub github_id: usize,
+    pub github: String,
+    pub name: String,
+    pub discord_id: Option<usize>,
```
rylev, line 139: I wonder if we could leave this out unless we're sure we need it. At some point I'd like to review whether we actually need discord support and that would be made easier by not adding more references to discord.

Turbo87: https://github.com/rust-lang/team/pull/1196#discussion_r1448944393 🤷‍♂️ 

I'm fine with removing it if we're sure that nothing is actually using it

Kobzol: At the very least, no one is *definitely* using it through the `people` field :laughing:

Turbo87: I've removed the field now

rylev (approved):

PR #1293, opened 2024-01-24
Re-introduce rust-analyzer-contributors

Veykril:
We kind of have a use for this again now

@rust-lang/rust-analyzer

### DIFF
--- a/teams/rust-analyzer.toml
+++ b/teams/rust-analyzer.toml
@@ -7,7 +7,6 @@ members = [
     "lnicola",
     "Veykril",
     "flodiebold",
-    "bjorn3",
     "lowr",
     "HKalbasi",
 ]
--- a/people/alibektas.toml
+++ b/people/alibektas.toml
@@ -1,0 +1,5 @@
+name = "Ali Bektas"
+github = "alibektas"
+github-id = 20956650
+email = "bektasali@protonmail.com"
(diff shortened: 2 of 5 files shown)


Veykril:
cc @alibektas, @davidbarsky, @Nadrieril

teams/rust-analyzer.toml
```diff
@@ -7,7 +7,6 @@ members = [
     "lnicola",
     "Veykril",
     "flodiebold",
-    "bjorn3",
```
rylev, line 10: Should this be removed to `alumni` or was the inclusion a mistake?

rylev (changes requested):

teams/rust-analyzer.toml
Veykril: That was intentional, `bjorn3` was only in the team for issue triaging permissions when we removed the contributors team. I've moved them to the contributors back now with this.

Nadrieril:
I notice that my `people/Nadrieril.toml` file doesn't have a `zulip-id`. If this is important for r-a, my id is 245339

Veykril:
I don't think it is

teams/rust-analyzer.toml
bjorn3: LGTM

PR #1299, opened 2024-01-25
remove branch protections from regex repo

BurntSushi:
Commit 55158aeb89463e1fc2d8e5b5321810cb70757fc3 introduced automation
for the regex repo but also enabled branch protection for `master`. The
regex repo has never used branch protection and it is low traffic enough
to be annoying.

While most changes go in through a PR, I do for example push directly to
master when cutting a release (which is done through my own automation
locally).

### DIFF
--- a/repos/rust-lang/regex.toml
+++ b/repos/rust-lang/regex.toml
@@ -5,8 +5,3 @@ bots = []
 
 [access.teams]
 regex = "maintain"
-
-[[branch-protections]]
-pattern = "master"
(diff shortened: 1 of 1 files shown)


Kobzol:
Sorry for that. I'm trying to proactively add branch protections where they were missing, but it's probably too annoying sometimes, even without any CI checks or required approvals.

Mark-Simulacrum (approved):

PR #1306, opened 2024-01-28
Add crates.io-index repository under automation

Kobzol:
Repo: https://github.com/rust-lang/crates.io-index

Extracted from GH:
```toml
org = "rust-lang"
name = "crates.io-index"
description = "Registry index for crates.io"
bots = []

[access.teams]
security = "pull"
crates-io = "write"
crates-io-on-call = "write"

[access.individuals]
jdno = "admin"
pietroalbini = "admin"
Xylakant = "write"
Mark-Simulacrum = "admin"
tshepang = "write"
MarcoIeni = "admin"
justahero = "write"
Turbo87 = "write"
rust-lang-owner = "admin"
Veykril = "write"
jtgeibel = "write"
mdtro = "write"
LawnGnome = "write"
skade = "write"
eth3lbert = "write"
listochkin = "write"
Rustin170506 = "write"
carols10cents = "write"

[[branch-protections]]
pattern = "master"
required-approvals = 0
pr-required = false
```

### DIFF
--- a/repos/rust-lang/crates.io-index.toml
+++ b/repos/rust-lang/crates.io-index.toml
@@ -1,0 +1,12 @@
+org = "rust-lang"
+name = "crates.io-index"
+description = "Registry index for crates.io"
+bots = []
+
+[access.teams]
+crates-io = "write"
+crates-io-on-call = "write"
+
+[[branch-protections]]
+pattern = "master"
+pr-required = false


repos/rust-lang/crates.io-index.toml
```diff
@@ -0,0 +1,12 @@
+org = "rust-lang"
+name = "crates.io-index"
+description = "Registry index for crates.io"
```
ehuss, line 3: To accommodate #1353.

```suggestion
description = "Registry index for crates.io"
homepage = "https://crates.io"
```

Turbo87:
> Should I add bors to it? It's not managed by `homu`, but the bors GH account pushes to the repo regularly (even though it seems like it didn't have access?!), this is quite weird.

I totally agree that it is weird, and IMHO we should change it eventually. The crates.io backend is pushing to the repository with a deploy SSH key (AFAIK) and uses git author information (i.e. email) to associate the commits to the bors GH account. those are not signed commits though, so anyone basically can create commits like that, just not necessarily in this repo.

to answer the initial question: no, bors does not need access to this repository.

repos/rust-lang/crates.io-index.toml
```diff
@@ -0,0 +1,13 @@
...
+homepage = "https://crates.io"
+bots = []
+
+[access.teams]
+crates-io = "write"
+crates-io-on-call = "write"
+
+[[branch-protections]]
+pattern = "master"
+required-approvals = 0
```
Turbo87, line 13: unless I'm missing something, the `pr-required = false` from the PR text seems quite important here, since the crates.io backend is pushing directly to the master branch

Kobzol: Indeed. The issue is that `pr-required` is not implemented yet :laughing:

Kobzol:
@Turbo87 Just to clarify the access patterns, who exactly pushes to this repository? It seems like it's the bors bot, and sometimes you manually force push a squashed version?

Do we want to change this so that the commits are created by someone else than the bors account?

Turbo87:
> Just to clarify the access patterns, who exactly pushes to this repository? It seems like it's the bors bot, and sometimes you manually force push a squashed version?

see [above](https://github.com/rust-lang/team/pull/1306#issuecomment-1967522650). I'm not sure how else to answer this 😅 



> Do we want to change this so that the commits are created by someone else than the bors account?

the commit are not created by the bors "account". they just use the same name/email combo as the bors account, which makes GitHub associate the commits with the account.

Kobzol:
Right, I understand that, I'm just not sure what's the association between the committer and the account that does the committing. You need to use some account/token for the pushes, and we need to provide access to that account to the protected branch on this repo. Essentially I'm asking what account is that (I know it's not bors) xD

Turbo87:
> You need to use some account/token for the pushes

no, it is using a deploy key, which is associated with a repository, not a specific account AFAIU.

Kobzol:
Ah, a deploy key, I see. In that case, if I understand the [documentation](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys#deploy-keys) correctly, a deploy key essentially has the same rights as an admin (on that repo). Therefore, we don't need to care about permisisons, I suppose.

If we specify `pr-required = false`, then the deploy key should be able to push to the branch normally. The problem is with force pushes, because we don't enable these in `sync-team`. Do you also force push with the deploy key, or with some other account? There is `bypassForcePushActorIds` which can be selectively choose accounts that can force push, but I'm not sure if we can scope that to a deploy key.

Turbo87:
> Do you also force push with the deploy key

yeah, the whole index squashing process is automated and runs on the server in a background job

Kobzol:
I realized that we don't need to model force pushes in `team`, we can just configure them manually, as they are not needed for almost any other repository. So this should be now ready for a review.

Turbo87 (approved):
LGTM AFAICT :)

rylev (approved):

Kobzol:
Looks like new commits are appearing, so hopefully all is well.

PR #1246, opened 2024-01-21
Add mdbook repo under automation

Kobzol:
Repo: https://github.com/rust-lang/mdBook

The `rust-mdbook-deploy` user looked.. intriguing, so I kept it there.

CC @ehuss

Extracted from GH:
```
org = "rust-lang"
name = "mdBook"
description = "Create book from markdown files. Like Gitbook but implemented in Rust"
bots = []

[access.teams]
infra = "pull"
security = "pull"

[access.individuals]
rust-mdbook-deploy = "write"
Mark-Simulacrum = "admin"
badboy = "admin"
ehuss = "write"
jdno = "admin"
rust-lang-owner = "admin"
Dylan-DPC = "write"
rustbot = "write"
rylev = "admin"
pietroalbini = "admin"

[[branch-protections]]
pattern = "master"
ci-checks = ["Success gate"]
dismiss-stale-review = false
pr-required = true
review-required = false
```

PR #1325, opened 2024-01-29
Fix mdbook permissions

ehuss:
https://github.com/rust-lang/team/pull/1246 added mdbook to the team repo, but I did not notice that it removed @Dylan-DPC's permissions. This restores those permissions.

I was intending to create a team for this, but I will need to write an RFC first.

### DIFF
--- a/repos/rust-lang/mdBook.toml
+++ b/repos/rust-lang/mdBook.toml
@@ -6,6 +6,11 @@ bots = ["rustbot"]
 [access.teams]
 devtools = "write"
 
+[access.individuals]
+# Dylan-DPC is maintaining mdbook along with ehuss. Planning to add a team
+# soon, and then this can be removed.
+Dylan-DPC = "write"
+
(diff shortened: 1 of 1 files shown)


repos/rust-lang/mdBook.toml
```diff
@@ -6,6 +6,9 @@ bots = ["rustbot"]
 [access.teams]
 devtools = "write"
 
+[access.individuals]
+Dylan-DPC = "write"
```
rylev, line 10: I'd love for us to get in the habit of documenting why individuals have individual access so we can continually evaluate whether those conditions still exist.

rylev (changes requested):

repos/rust-lang/mdBook.toml
ehuss: Good idea! I added a comment. I do hope this is temporary.

rylev (approved):

PR #1336, opened 2024-01-30
Don't use `usize`s for user ids

WaffleLapkin:
I don't think github (zulip, discord) IDs are platform dependant.

### DIFF
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -31,7 +31,7 @@ pub struct Team {
 pub struct TeamMember {
     pub name: String,
     pub github: String,
-    pub github_id: usize,
+    pub github_id: u64,
     pub is_lead: bool,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
     pub roles: Vec<String>,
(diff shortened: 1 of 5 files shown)


apiraino:
@WaffleLapkin interesting. Do you think it could also make sense propagating this change [to the triagebot](https://github.com/rust-lang/triagebot/blob/df574c1df483f9adc6926db284c82f3ddbfa6bad/src/github.rs#L20) (currently GH user ids are stored as `i64`) just for the sake of using the same type everywhere? FWIW the crate `octocrab` used by the triagebot [agrees with you](https://docs.rs/octocrab/latest/octocrab/models/struct.UserId.html).

to be clear: not suggesting to submit a patch, simply asking for an opinion :slightly_smiling_face:

WaffleLapkin:
@apiraino I actually started with the triagebot change and accidentally changed this cargo dep. so decided to submit this first and then pull changes into triagebot

rylev (approved):
Seems like a reasonable change.

We should probably change sync-team as well. cc @Kobzol

Kobzol:
Implemented [here](https://github.com/rust-lang/sync-team/pull/67). But it should be merged after this one, since it will remove some unneeded casts once `v1` switches to `u64`.

rylev (approved):

PR #1338, opened 2024-02-02
Add taplo configuration file to format TOML files

dtolnay:
Taplo is the TOML formatter used by VS Code's TOML extension. https://taplo.tamasfe.dev

I have not added a GitHub Actions job to enforce formatting because it isn't necessarily valuable enough to be something every contributor to this repo needs to install, but I am open to it. Even without a CI job, someone manually running the formatter when things seem to be getting out of shape (a few years apart) would still be positive.

~~I have not touched the `repos` directory because I know there is a lot of work in flight by @Kobzol there. It can be formatted afterward.~~ *Edit: the formatter-generated changes in that directory are minor.*


rylev (approved):
I'm personally in favor of this because my editor already does this.

weihanglo:
Not really a blocker. Just wonder what would happen when `cargo fmt` starts formatting `Cargo.toml`?

https://github.com/rust-lang/style-team/issues/188

dtolnay:
If the formatting clashes, we would add `exclude = [..., "**/Cargo.toml"]` in .taplo.toml.

jackh726:
I think this is a good chance. Semi-minor nit though @dtolnay can you squash to two commits? One with the configuration, and one with the actual formatting changes? (Or, just one commit is fine really)

dtolnay:
Squashed to 2 commits. The first commit is handwritten (only touching `.taplo.toml`) and the second commit is 100% generated by running `taplo format` in the repo root.

PR #1208, opened 2024-01-15
Add llvm-project repository under automation

Kobzol:
Repo: https://github.com/rust-lang/llvm-project

I kept the branch protections, although I'm not sure if they are even needed on this repo.

CC @nikic

Extracted from GH:
```
org = "rust-lang"
name = "llvm-project"
description = "Rust-specific fork of LLVM."
bots = []

[access.teams]
wg-llvm = "maintain"
core = "admin"
security = "pull"

[access.individuals]
cuviper = "maintain"
badboy = "admin"
jdno = "admin"
pietroalbini = "admin"
nikic = "maintain"
nagisa = "maintain"
rust-lang-owner = "admin"
Mark-Simulacrum = "admin"
rylev = "admin"

[[branch-protections]]
pattern = "rustc/*"
ci-checks = []
dismiss-stale-review = false
pr-required = false
review-required = false

[[branch-protections]]
pattern = "master"
ci-checks = []
dismiss-stale-review = false
pr-required = false
review-required = false
```

PR #1346, opened 2024-02-12
Drop branch protections on llvm-project

nikic:
To restore to something that more closely resembles the state before https://github.com/rust-lang/team/pull/1208. Merging new changes from upstream release branches into our fork doesn't really benefit from going through a PR or review, especially for work-in-progress upgrade branches.

I would have preferred to keep the "no force push and branch deletion" part of the branch protection, but it looks like disabling the PR requirement is not supported.

### DIFF
--- a/repos/rust-lang/llvm-project.toml
+++ b/repos/rust-lang/llvm-project.toml
@@ -5,9 +5,3 @@ bots = []
 
 [access.teams]
 wg-llvm = "maintain"
-
-[[branch-protections]]
-pattern = "rustc/*"
-
-[[branch-protections]]
-pattern = "master"


Kobzol:
Indeed, currently in team, a branch protection means that a PR will be required.

rylev:
Should we try to add a `pr-required = false` option so that we don't need to get rid of basic branch protection?

Kobzol:
We can, but we'd be slowly starting to converge to the messy system that GitHub currently has :)

rylev:
Well it feels wrong that in order to allow the release branch pattern, we are asking folks to give up on all protections.

Kobzol:
It would be nice to have some better syntax for this than 5 true/false booleans though, like
```toml
[[branch-protections]]
pattern = "..."
pr = false
# pr = { required = true, min-approvals = 0 }
```
but we'd have to change the team data format for that.

Anyway, I'm not opposed to adding `pr-required`.

Mark-Simulacrum (approved):
I think we should merge this to unblock/simplify LLVM work in the meantime -- @nikic does that make sense from your perspective?

jackh726 (approved):
I'm going to merge this for now - we can revisit with better support later.

PR #1347, opened 2024-02-18
Add calebzulawski to libs-contributors

calebzulawski:
From this comment, I'm wondering if I should have bors privileges: https://github.com/rust-lang/rust/pull/121269#issuecomment-1951374615

I don't think it makes sense to add bors permissions to the project-portable-simd team so adding myself to @rust-lang/libs-contributors might be correct?  Feel free to suggest any alternatives, and I'm not offended if this doesn't go anywhere.  Thanks!

### DIFF
--- a/teams/libs-contributors.toml
+++ b/teams/libs-contributors.toml
@@ -18,6 +18,7 @@ members = [
     "Nilstrieb",
     "jhpratt",
     "joboet",
+    "calebzulawski",
 ]
 alumni = [
     "shepmaster",
--- a/teams/libs-contributors.toml
+++ b/teams/libs-contributors.toml
@@ -18,7 +18,7 @@ members = [
     "Nilstrieb",
     "jhpratt",
     "joboet",
-    "calebzulawski",
+    "calebzulawski", # std::simd
 ]
 alumni = [
     "shepmaster",


workingjubilee:
Yes, probably. I got them shortly before libs-contributors was a thing, if memory serves, so I wasn't sure on what grounds I would recommend extending them, but given it exists now, that should in fact be done.

calebzulawski:
cc @thomcc @Amanieu if this is libs related

m-ou-se:
If this only for r+ permissions for updating the portable simd subtree, it's better to add it to people/calibzulawski.toml with a comment what it's for:

```toml
[permissions]
bors.rust.review = true # For the portable SIMD subtree.
```

calebzulawski:
I'd also like to review PRs to rust-lang/rust that change std::simd and the compiler intrinsics (not really T-libs I guess, but I've contributed a lot more to std than to rustc).  I'm not quite ready yet to be added to the rotation but my goal is to review elsewhere in std too.

calebzulawski:
Learned my lesson to never try to resolve the conflicts through github... Anyway, bumping this since I'd like to be able to review a new PR and contributors expect to be able to r? me (rust-lang/rust#122905).  It still sounds like lib-contributors is appropriate but I added a comment to clarify.

jackh726:
This needs approval from one of @m-ou-se or @Amanieu

Kobzol:
CC @m-ou-se @Amanieu (re-pinging after a few weeks)

Amanieu (approved):
There was some disagreement within the team about the exact scope, but we're happy to add @calebzulawski as a reviewer for things related to `std::simd`.

jackh726 (approved):

PR #1359, opened 2024-02-28
Add the option to not require a PR for protected branches

Kobzol:
This PR adds a `pr-required = <true/false>` option to branch protections. There are some repositories that use branch protections, but that do not require a PR.

I also refactored the branch protection mode to make the data representation clearer/make it impossible to represent invalid states (such as PR not required, but a required approval count is set). This is a compatibility break for the v1 data format, but branch protections are only [read](https://github.com/search?type=code&auto_enroll=true&q=org%3Arust-lang+required_approvals) by `sync-team` anyway, so it shouldn't be a big deal.

An interesting thing is that it does not actually make sense to require any CI status checks if a PR is not required, because we do not enable anyone to bypass the branch protections. Because if you set a CI check, and try to push to the branch without a PR, it will be rejected.

Without a required PR, the branch protection still disallows force pushes and deletion, and allows us to set push actor allowances.

Relevant repos:
- https://github.com/rust-lang/team/pull/1346
- https://github.com/rust-lang/team/pull/1306#discussion_r1504924219
- https://github.com/rust-lang/team/pull/1360

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -333,13 +333,18 @@ Admins cannot override these branch protections. If an admin needs to do that, t
 # The pattern matching the branches to be protected (required)
 pattern = "master"
 # Which CI checks to are required for merging (optional)
+# Cannot be set if `pr-required` is `false`.
 ci-checks = ["CI"]
 # Whether new commits after a reviewer's approval of a PR 
 # merging into this branch require another review. 
 # (optional - default `false`)
 dismiss-stale-review = false
+# Is a PR required when making changes to this branch?
+# (optional - default `true`)
+pr-required = true
 # How many approvals are required for a PR to be merged.
 # This option is only relevant if bors is not used.
+# Cannot be set if `pr-required` is `false`.
 # (optional - default `1`)
 required-approvals = 1
 # Which GitHub teams have access to push/merge to this branch.
--- a/tests/static-api/_expected/v1/repos.json
+++ b/tests/static-api/_expected/v1/repos.json
@@ -16,11 +16,15 @@
       "branch_protections": [
         {
           "pattern": "master",
-          "ci_checks": [
-            "CI"
-          ],
           "dismiss_stale_review": false,
-          "required_approvals": 1,
+          "mode": {
+            "pr_required": {
+              "ci_checks": [
+                "CI"
+              ],
+              "required_approvals": 1
+            }
+          },
           "allowed_merge_teams": []
         }
       ],
@@ -44,11 +48,15 @@
       "branch_protections": [
         {
           "pattern": "master",
-          "ci_checks": [
-            "CI"
-          ],
           "dismiss_stale_review": false,
-          "required_approvals": 1,
+          "mode": {
+            "pr_required": {
+              "ci_checks": [
+                "CI"
+              ],
+              "required_approvals": 1
+            }
+          },
           "allowed_merge_teams": [
             "foo"
           ]
--- a/tests/static-api/_expected/v1/repos/archived_repo.json
+++ b/tests/static-api/_expected/v1/repos/archived_repo.json
@@ -14,11 +14,15 @@
   "branch_protections": [
     {
       "pattern": "master",
-      "ci_checks": [
-        "CI"
-      ],
       "dismiss_stale_review": false,
-      "required_approvals": 1,
+      "mode": {
+        "pr_required": {
+          "ci_checks": [
+            "CI"
+          ],
+          "required_approvals": 1
+        }
+      },
       "allowed_merge_teams": []
     }
   ],
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -210,11 +210,20 @@ pub enum RepoPermission {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum BranchProtectionMode {
+    PrRequired {
+        ci_checks: Vec<String>,
+        required_approvals: u32,
+    },
+    PrNotRequired,
+}
+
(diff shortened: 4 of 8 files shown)


tests/static-api/_expected/v1/repos.json
```diff
@@ -16,11 +16,15 @@
...
           "dismiss_stale_review": false,
-          "required_approvals": 1,
+          "mode": {
+            "PrRequired": {
+              "ci_checks": [
+                "CI"
+              ],
+              "required_approvals": 1
+            }
+          },
```
Turbo87, line 27: this seems like a breaking change to the API. idk if anything is using the existing keys, but since some of them are removed I guess it should treated as potentially breaking?

docs/toml-schema.md
```diff
@@ -301,13 +301,18 @@ octocat = "write"
 # The pattern matching the branches to be protected (required)
 pattern = "master"
 # Which CI checks to are required for merging (optional)
+# Cannot be set if `pr-required` is `false`.
 ci-checks = ["CI"]
 # Whether new commits after a reviewer's approval of a PR 
 # merging into this branch require another review. 
 # (optional - default `false`)
 dismiss-stale-review = false
+# Is a PR required when making changes to this branch?
```
rylev, line 310: Should we consider consolidating all PR related options into a `pr-options` (or similar name)? The presence of the options key would be equivalent to `pr-required = true`.

docs/toml-schema.md
```diff
@@ -301,13 +301,18 @@ octocat = "write"
 # The pattern matching the branches to be protected (required)
 pattern = "master"
 # Which CI checks to are required for merging (optional)
+# Cannot be set if `pr-required` is `false`.
```
rylev, line 304: Would it ever make sense for PRs to be optional but to require these checks _if_ a PR is opened?

Kobzol: I was also thinking of this, but it's not actually possible using the GH model, AFAIK. You can enable or disable checks. Once checks are enabled, PRs are no longer optional, because without a PR you cannot do a merge.

docs/toml-schema.md
Kobzol: I wanted to do that originally, but it seemed a bit disruptive, since it would also need to contain `ci-checks`. Actually, pretty much all of the current fields except for `pattern` and `allowed-merge-teams` would need to be in the `pr-options` field. At that point it seemed easier to me to just keep it top level.

tests/static-api/_expected/v1/repos/archived_repo.json
```diff
@@ -14,11 +14,15 @@
   "branch_protections": [
     {
       "pattern": "master",
-      "ci_checks": [
-        "CI"
-      ],
       "dismiss_stale_review": false,
-      "required_approvals": 1,
+      "mode": {
+        "PrRequired": {
```
rylev, line 19: nit: should this be serialized with snake case?

Kobzol: Indeed it should! Fixed, thanks.

tests/static-api/_expected/v1/repos.json
Kobzol: https://github.com/rust-lang/sync-team/pull/75 will also be needed to be merged first, since it adds support for `RenovateBot` (`sync-team` now doesn't compile after updating `team` since the bot was added).

tests/static-api/_expected/v1/repos.json
```diff
@@ -43,11 +47,15 @@
       "branch_protections": [
         {
           "pattern": "master",
-          "ci_checks": [
```
jackh726, line 46: This change breaks anything (namely sync-team) that relies on the current schema?

Don't think we can do this (at least not immediately).

Kobzol: I don't think that there's anything else than `sync-team` that read the branch protections, and the `sync-team` change is [ready](https://github.com/rust-lang/sync-team/pull/76). I already did some breaking changes recently that broke sync-team, it doesn't seem like a big deal.

Although, it would be bad if this broke other tools that just reuse the schema, even if they don't read the branch protections at all. Not sure if there are any of these though. Is there even any other tool apart from `sync-team` that reads the `repos` endpoint?

Kobzol:
@Turbo87 @jackh726 @rylev Apart from the breaking change to `sync-team` (which we expect and which will be resolved soon after merging this), do you have any other concerns about this change?

Kobzol:
Yes, the index repo will need some additional special treatment I think, that is tangential (although it will also need the ability to avoid PRs on branches, I suppose).

jdno:
> I'm not concerned about the breaking change because we control sync-team so we can redeploy it after changing the format

Theoretically, this is a public API and we're not necessarily the only consumers. But a quick [code search](https://github.com/search?q=lang%3ARust+rust_team_data%3A%3Av1%3A%3ARepo&type=code) doesn't show any uses of the `Repo` struct outside of `sync-team`. And it's definitely not worth the effort to bump the API version for this...

Turbo87:
> But a quick [code search](https://github.com/search?q=lang%3ARust+rust_team_data%3A%3Av1%3A%3ARepo&type=code) doesn't show any uses of the `Repo` struct outside of `sync-team`.

btw not everyone is necessarily using this as a git dependency... (https://github.com/rust-lang/crates.io/blob/e922d6806cfcb90a4eb8838ec3a29029b65fe9e0/src/team_repo.rs#L63 😉)

Kobzol:
I was specifically looking for usage of the `repos` endpoint and the `required_approvals` attribute [before](https://github.com/search?type=code&auto_enroll=true&q=org%3Arust-lang+required_approvals), which didn't seem to show anything.

MarcoIeni (approved):

MarcoIeni:
merged as asked by Kobzol

PR #1381, opened 2024-03-16
Retire wg-grammar

ehuss:
wg-grammar has not met in several years, so I think it is time to go ahead and retire it. I don't foresee it starting up again. Also, I'm presuming that the spec team will be picking up this work in some form.

cc @rust-lang/wg-grammar 
cc @rust-lang/spec

### DIFF
--- a/repos/archive/rust-lang/wg-grammar.toml
+++ b/repos/archive/rust-lang/wg-grammar.toml
@@ -4,4 +4,3 @@ description = "Where the work of WG-grammar, aiming to provide a canonical gramm
 bots = []
 
 [access.teams]
-wg-grammar = "write"
--- a/teams/archive/wg-grammar.toml
+++ b/teams/archive/wg-grammar.toml
@@ -3,12 +3,11 @@ subteam-of = "lang"
 kind = "working-group"
 
 [people]
-leads = ["eddyb"]
-members = [
+leads = []
+members = []
+alumni = [
     "eddyb",
     "CAD97",
-]
-alumni = [
     "Centril",
     "qmx",
 ]


repos/archive/rust-lang/gll.toml
```diff
@@ -0,0 +1,6 @@
+org = "rust-lang"
+name = "gll"
+description = " GLL parsing framework."
```
rylev, line 3: Should we put something on the README and/or description explaining any detail about why this is archived? In other organizations I'm a part of, we make a habit of putting a prominent message in the README explaining where those looking at the repo should direct their attention.

rylev (commented):
r? @eddyb

repos/archive/rust-lang/gll.toml
ehuss: Sure! I posted https://github.com/rust-lang/gll/pull/145 and https://github.com/rust-lang/wg-grammar/pull/70. However, I'm not sure how active @eddyb is to be able to see this.

Kobzol:
The `wg-grammar` repo now has the deprecation notice (https://github.com/rust-lang/wg-grammar/pull/70). eddyb has asked for `gll` to be moved back under his org, so I suggest to remove `gll` from `team`, so that it can be moved out without `sync-team` recreating it.

repos/archive/rust-lang/gll.toml
eddyb: > However, I'm not sure how active @eddyb is to be able to see this.

My notifications are a disaster because I didn't get myself out of @rust-lang/compiler the moment I stopped being able to pay attention to them, so while I keep every GitHub email for posterity, I have no easy way to be notified when people want me to look at anything under `rust-lang/`.
You pretty much have to DM me on Zulip or Discord for the time being, sorry!
<sub>(I've been trying to fix it but life keeps handing me lemons - also, theoretically the `@eddyb` on GitHub may have been enough to bypass that but maybe I need to do more email automation because I don't remember the GH mobile app catching it, and it's also terrible in other ways)</sub>

---

Also, IMO `gll` does not belong in `rust-lang/`, it's a toy parsing library developed originally under the `@LykenSol` org and it got moved because wg-grammar wanted some level of "vendoring". Completely unnecessary and it would be best to move it back if that's an option. (see also https://github.com/rust-lang/gll/pull/145#issuecomment-2039830588)

eddyb:
Oops, raced a bit with @Kobzol's comment, I left mine further above as a review comment (https://github.com/rust-lang/team/pull/1381#discussion_r1553676133), leaving this note here so it doesn't get missed (tho it mostly just adds more context, the conclusion is the same).

ehuss:
Sure! I have removed gll, and I think it should be safe to transfer it.

Kobzol:
Ah, I didn't realize that `gll` wasn't even tracked before this PR. Yeah, so in that case we can transfer it independently of this PR.

PR #1403, opened 2024-03-16
Require subteam-of to be specified.

ehuss:
This changes it so that Teams, Working Groups, and Project Groups set the `subteam-of` field. Marker teams do not require it.

The reason is to ensure that teams are connected, and to make it a little easier to connect the top-level teams to the leadership-council.

### DIFF
--- a/tests/static-api/teams/wg-test.toml
+++ b/tests/static-api/teams/wg-test.toml
@@ -1,5 +1,6 @@
 name = "wg-test"
 kind = "working-group"
+subteam-of = "leadership-council"
 
 [people]
 leads = ["user-2"]
--- a/tests/static-api/teams/wg-test.toml
+++ b/tests/static-api/teams/wg-test.toml
@@ -1,6 +1,6 @@
 name = "wg-test"
 kind = "working-group"
-subteam-of = "leadership-council"
+top-level = true
 
 [people]
 leads = ["user-2"]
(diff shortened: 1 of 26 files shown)


ehuss:
This is a draft because it requires https://github.com/rust-lang/www.rust-lang.org/pull/1937 to avoid breaking the website.

ehuss:
This should be ready now that the website has deployed.

rylev (changes requested):
This seems fine to me, but I can imagine some might view the top-level teams as being subteams of the leadership council as not correct since according the RFC, the leadership Council is really just a collection of the top-level teams authority and not really "sovereign". 

This is somewhat nitpicky, but I do think this might send the wrong signal. Perhaps we should have a `top-level = true` option and require that a team is either marked as top-level *or* has a parent team.

ehuss:
Sounds good, I went ahead and added a top-level property.

tests/static-api/teams/wg-test.toml
```diff
@@ -1,5 +1,6 @@
 name = "wg-test"
 kind = "working-group"
+top-level = true
```
rylev, line 3: I think we should also validate that working groups cannot be top-level teams.

rylev (changes requested):

tests/static-api/teams/wg-test.toml
ehuss: Sure! I have added a check.

rylev (approved):

PR #1404, opened 2024-03-17
Add `datafrog` under automation

Kobzol:
Repo: https://github.com/rust-lang/datafrog

CC @lqd, looks good?

Extracted from GH:
```toml
org = "rust-lang"
name = "datafrog"
description = "A lightweight Datalog engine in Rust"
bots = []

[access.teams]
compiler = "admin"
core = "admin"
infra = "admin"
wg-polonius = "write"
security = "pull"

[access.individuals]
lcnr = "admin"
davidtwco = "admin"
pietroalbini = "admin"
nagisa = "admin"
pnkfelix = "admin"
compiler-errors = "admin"
badboy = "admin"
ecstatic-morse = "write"
michaelwoerister = "admin"
oli-obk = "admin"
kennytm = "admin"
rust-lang-owner = "admin"
lqd = "write"
cjgillot = "admin"
shepmaster = "admin"
amandasystems = "write"
Kobzol = "admin"
petrochenkov = "admin"
jackh726 = "admin"
wesleywiser = "admin"
nikomatsakis = "write"
rylev = "admin"
eddyb = "admin"
estebank = "admin"
Mark-Simulacrum = "admin"
jdno = "admin"
Aaron1011 = "admin"
matthewjasper = "admin"
```

### DIFF
--- a/repos/rust-lang/datafrog.toml
+++ b/repos/rust-lang/datafrog.toml
@@ -1,0 +1,9 @@
+org = "rust-lang"
+name = "datafrog"
+description = "A lightweight Datalog engine in Rust"
+bots = []
+
+[access.teams]
+wg-polonius = "write"
+compiler = "write"
+compiler-contributors = "write"


lqd:
The teams look good yes.

lqd:
Actually I don't understand what the description says vs the toml file, I only looked at the "extracted from GH" description, while the toml doesn't seem equivalent?

Kobzol:
The "Extracted from GH part" is basically the current raw data from GH, the content of the PR is a "normalized state" that removes obsolete stuff and access for individuals.

lqd:
The toml says only the wg-polonius will have access: are there implicit teams that will have access in addition to what the toml says? Other infra, admins, etc and compiler will lose the access they have.

Kobzol:
> are there implicit teams that will have access in addition to what the toml says

No.

> Otherwise infra, admins, compiler, etc will lose the access they currently have.

Yes, that's kind of the idea :sweat_smile: We want to ideally reduce the existing accesses as much as possible, and only expand them when someone actually needs a broader access. Out of the previous teams, only `t-compiler` makes sense to me. Do you think that we should add access to it for this repo?

lqd:
Most, if not all, rust-lang crates used by the compiler should at least have compiler (and IMHO compiler-contributors), infra, admins, etc in case anyone needs to merge a PR and do a release. IIRC there's an old "crates.io conformity policy" from Niko somewhere which described this.

Kobzol:
CC @ehuss

rylev:
infra-admins have implicit access as they are admins of the entire GitHub org. Other than that, I imagine the infra team would not feel comfortable cutting a release for this, so giving them access does not make sense to me.

lqd:
Sure, I'm not advocating for infra specifically: if they don't want to have access anymore, it's fine and their prerogative. They're of course never expected to cut a release, but seems to have always been a fallback in case of emergency I guess (for example if an issue in a crate prevents a beta bump, or another `rustc-serialize` situation happens again, things like that).

The crates policy from t-compiler (to my knowledge, it was never fully applied consistently) is https://github.com/rust-lang/compiler-team/blob/master/content/procedures/crates.md?plain=1#L61.

So let's just add t-compiler and t-compiler-contributors here, that should be enough.

Kobzol:
Added `compiler` and `compiler-contributors`.

rylev (approved):

PR #1409, opened 2024-03-18
feat: convert rustlings membership to a team

shadows-withal:
I want to add a co-maintainer to Rustlings (which will be done in a later step), so I'm converting the single person repo permission into a team permission. This is not a formal team or anything, just a convenience to let multiple people have the same permission, hence why the team itself is very minimal.

### DIFF
--- a/teams/rustlings.toml
+++ b/teams/rustlings.toml
@@ -1,0 +1,10 @@
+name = "rustlings"
+kind = "team"
+
+[people]
+leads = ["shadows-withal"]
+members = ["shadows-withal"]
+alumni = []
+
+[[github]]
+orgs = ["rust-lang"]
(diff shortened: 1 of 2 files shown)


shadows-withal:
Oh, I also changed the permission to `maintain` because I realized that with `write` I can't really do things like re-trigger CI runs, which is annoying.

teams/rustlings.toml
```diff
@@ -0,0 +1,10 @@
+name = "rustlings"
+kind = "team"
```
rylev, line 2:
```suggestion
kind = "marker-team"
```

rylev (changes requested):
This looks good if you make the team a marker team. Marker teams are used to indicate "just a collection of people" rather than an actual team in the governance sense.

rylev (approved):

PR #1411, opened 2024-03-20
Add the new *Clippy-Contributors* team 🎉

xFrednet:
This PR adds the new *Clippy-Contributors* team which has been proposed in rust-lang/rust-clippy#12481 and was accepted on Zulip yesterday! :tada:

As part of this PR, I also want to welcome the four initial heroes, who will start be part of this team:
* @GuillaumeGomez
* @J-ZhengLi
* @m-rph
* @pitaj

Thank you, for all the work you already put into Clippy. We're happy to have you! :heart:

---

cc: rust-lang/clippy

r? @flip1995 & @Manishearth

### DIFF
--- a/teams/clippy-contributors.toml
+++ b/teams/clippy-contributors.toml
@@ -21,4 +21,4 @@ zulip-stream = "clippy"
 repo = "https://github.com/rust-lang/rust-clippy"
 
 [[zulip-groups]]
-name = "T-clippy"
+name = "T-clippy-contributors"
--- a/teams/clippy-contributors.toml
+++ b/teams/clippy-contributors.toml
@@ -4,6 +4,10 @@ subteam-of = "clippy"
 [people]
 leads = []
 members = [
+    "GuillaumeGomez",
+    "J-ZhengLi",
+    "m-rph",
+    "pitaj",
 ]
 alumni = []
 
--- a/teams/clippy-contributors.toml
+++ b/teams/clippy-contributors.toml
@@ -1,0 +1,20 @@
+name = "clippy-contributors"
+subteam-of = "clippy"
+
+[people]
+leads = []
+members = [
+]
+alumni = []
+
+[[github]]
+orgs = ["rust-lang"]
+
+[website]
+name = "Clippy-Contributors"
+description = "Regular contributors to the Clippy linter"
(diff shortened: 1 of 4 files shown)


m-rph:
Hej that's me! :D Thank you for all your help and patience along the way! I am very proud to be a member of this team!

Manishearth (approved):

teams/clippy-contributors.toml
```diff
@@ -0,0 +1,24 @@
...
+orgs = ["rust-lang"]
+
+[website]
+name = "Clippy-Contributors"
+description = "Regular contributors to the Clippy linter"
+zulip-stream = "clippy"
+repo = "https://github.com/rust-lang/rust-clippy"
+
+[[zulip-groups]]
+name = "T-clippy"
```
rylev, line 24: Should this be

```suggestion
name = "T-clippy-contributors"
```

xFrednet: I think it makes sense to include them in the normal `T-clippy` Zulip group. We get very few pings, like one every two weeks when we have a team meeting, where the contributors are also invited to.

Even if another team would ping us for a design decision, the feedback from them would be useful. That's why I set it to `T-clippy`. Though, I'm open to change this :)

rylev: Then you need to add `clippy-contributors` to the Zulip group definition in clippy.toml using the `extra-teams` key. Right now, adding a duplicate team definition causes the last definition to be processed to override all other definitions which is obviously a bug. I would argue we need to throw a validation error in this case instead of allowing the group definitions be split across files.

flip1995: In that case, I would go with Ryan's original suggestion and just add them to `T-clippy-contributors`. No need to ping them for things they don't necessarily have to pay attention to.

flip1995 (approved):
LGTM, apart from Ryan's comment. Should be merged together with/after the RFC in the Clippy repo.

rylev (approved):

rylev:
I'll wait until the RFC is officially merged before merging this. Ping me if I forget!

xFrednet:
Hey @rylev, we just merged the RFC. Can you merge this as well to make the team official? :D

PR #1422, opened 2024-04-05
update and archive project-thir-unsafeck

lcnr:
THIR unsafeck is now stable since 1.77 (https://github.com/rust-lang/rust/pull/117673/) and MIR unsafeck has since been removed (https://github.com/rust-lang/rust/pull/123322) :tada: I think we should therefore archive the working group.

It was also outdated. I added @LeSeulArtichaut to alumni, which afaict was not possible when they left the project in https://github.com/rust-lang/team/pull/621. I also added @matthewjasper as a member as they've been doing most of the work towards actually getting it stabilized.

Unsure what to do with the afaict dead zulip stream (https://rust-lang.zulipchat.com/#narrow/stream/278509-project-thir-unsafeck). Ideally it should be archived and locked.

cc @nikomatsakis

### DIFF
--- a/repos/archive/rust-lang/project-thir-unsafeck.toml
+++ b/repos/archive/rust-lang/project-thir-unsafeck.toml
@@ -4,4 +4,3 @@ description = "Tracking the project to refactor the unsafe check to operate on T
 bots = []
 
 [access.teams]
-project-thir-unsafeck = "maintain"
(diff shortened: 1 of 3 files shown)


Kobzol:
Can you please also add a commit that moves the `repos/rust-lang/project-thir-unsafeck` directory under `repos/archived/rust-lang/project-thir-unsafeck`? And also remove the access for the `project-thir-unsafeck` project group. That will archive the repo and should fix the CI error.

Kobzol:
Ah, one more error, all members of archived teams must be alumni.

lcnr:
Should have maybe looked at the CI more quickly myself to save you some work :blush: thanks :heart:

PR #1437, opened 2024-04-23
Remove branch protection from staging branch of `bors`

Kobzol:
I need to force-push into `staging` to fix a bad rebase that I did before. And frankly, force-pushing from main into staging seems like the simplest deployment option for now anyway, to avoid additional merge commits.

### DIFF
--- a/repos/rust-lang/bors.toml
+++ b/repos/rust-lang/bors.toml
@@ -12,8 +12,3 @@ mods = "maintain"
 pattern = "main"
 ci-checks = ["Test", "Test Docker"]
 required-approvals = 0
-
-[[branch-protections]]
-pattern = "staging"
-ci-checks = ["Test", "Test Docker"]
-required-approvals = 0


repos/rust-lang/bors.toml
```diff
@@ -12,8 +12,3 @@ mods = "maintain"
 pattern = "main"
 ci-checks = ["Test", "Test Docker"]
 required-approvals = 0
-
-[[branch-protections]]
```
rylev, line 16: We have the `allowed-merge-teams` on branches which give teams the ability to push against protected branches. I think it's fine to merge this for now, but perhaps a better solution is adding a `allowed-merge-individuals` option and adding yourself to that? That way we can more easily document this work around.

Kobzol: Allowed merge teams doesn't allow force pushes, IIRC, the way sync-team is configured. It only allows to select a subset of teams that have write+ perms on the repo, which will have push/merge rights to the branch.

jackh726 (approved):

PR #1438, opened 2024-04-24
Validate homepage URLs for repositories

jdno:
Repositories can be configured with a homepage, which is prominently featured on GitHub as a link. For repositories under the `rust-lang` organization, we want to make sure that those links only point to domains that are explicitly allowed. Ideally, only domains owned and operated by the Rust project itself will be whitelisted.

The risk with other domains is that they might expire silently and get taken over by malicious actors, who can then host phishing campaigns or malware on sites "advertised" by the Rust project.

An initial selection of domains has been added to the allowlist for homepage URLs. The domains are either owned and operated by the infra-team or belong to GitHub.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -260,6 +260,8 @@ name = "my-repo"
 # A description of the repo (required)
 description = "A repo for awesome things!"
 # A URL that is displayed next to the description.
+# The URL or domain must be explicitely allowed in `allowed-domains` in the
+# configuration of the team repo.
 homepage = "https://www.rust-lang.org/"
 # The bots that this repo requires (required)
 bots = ["bors", "rustbot", "rust-timer"]


jdno:
This proposal is being discussed here: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Domains.20on.20GitHub

src/validate.rs
```diff
@@ -921,6 +922,24 @@ fn validate_member_roles(data: &Data, errors: &mut Vec<String>) {
...
 
+/// Ensure homepage URLs on GitHub are owned and managed by the Rust project
+fn validate_homepage_urls(data: &Data, errors: &mut Vec<String>) {
+    let allowed_domains: Vec<Regex> = data.config().allowed_domains().iter().map(|domain| Regex::new(domain).unwrap()).collect();
+
+    wrapper(data.repos(), errors, |repo, errors| {
+        if let Some(homepage) = &repo.homepage {
+            if !allowed_domains.iter().any(|domain| domain.is_match(homepage)) {
+                errors.push(format!(
+                    "homepage URL for {}/{} is not on an allowed domain: {}",
```
Kobzol, line 933: Maybe this could print the list of allowed domains to make debugging faster?

config.toml
```diff
@@ -18,6 +18,18 @@ allowed-github-orgs = [
     "rust-analyzer",
 ]
 
+allowed-domains = [
+    "https://(.*)rust-lang.org/?.*",
+    "https://areweasyncyet.rs/?.*",
+    "https://crates.io/?.*",
+    "https://docs.rs/?.*",
+    "https://github.com/rust-lang/rust/issues/35437",
```
apiraino, line 26: I'm curious: why also the URL of this github issue? :)

apiraino:
@jdno since this patch in in draft, I assume this list is still incomplete, correct?

src/validate.rs
jdno: Hm I'm not sure if that is helpful in the output. What I haven't done yet is document the allowlist in the schema and point to the configuration file. I think that should hopefully be enough for folks to quickly find the list.

config.toml
jdno: Because someone set it as the URL on a repo. 😅 I don't want to allow all of GitHub with something like `https://github.com/?.*`, because that could be abused with typosquatting. But we can change this to allow anything under the `rust-lang` organization with a rule like `http://github.com/rust-lang/?.*`.

jdno:
> @jdno since this patch in in draft, I assume this list is still incomplete, correct?

Yes, there are three URLs that are currently missing from the list. As far as I can tell, they are not owned/managed by the infra-team:

```
[ERROR rust_team::validate] validation error: homepage URL for rust-lang/rustlings is not on an allowed domain: https://rustlings.cool
[ERROR rust_team::validate] validation error: homepage URL for rust-lang/this-week-in-rust is not on an allowed domain: https://this-week-in-rust.org/
[ERROR rust_team::validate] validation error: homepage URL for rust-lang/wg-allocators is not on an allowed domain: http://bit.ly/hello-wg-allocators
```

src/validate.rs
Kobzol: Ok, fair enough, if it's documented than that should be enough.

apiraino:
> Yes, there are three URLs that are currently missing from the list. As far as I can tell, they are not owned/managed by the infra-team:

`rustup.rs` also? I didn't check them all, I'm sure you will find many more.

That `bit.ly` URL is ... ugh :sweat_smile:

jdno:
> `rustup.rs` also? I didn't check them all, I'm sure you will find many more.

I didn't add all the domains that we own. Instead, I looked at the links that are currently set as a `homepage` and added most of them to the allowlist. I'm not sure if there's a lot of value to preemptively add all our domains to the list or try to keep them in sync as we add more. I think it's easier to add them on a case-by-case basis.

config.toml
```diff
@@ -18,6 +18,18 @@ allowed-github-orgs = [
     "rust-analyzer",
 ]
 
+allowed-domains = [
+    "https://(.*)rust-lang.org/?.*",
```
Nilstrieb, line 22:
```suggestion
    "https://(.*)\\.rust-lang.org/?.*",
```
instead of all of this regex, would it make more sense to parse the URLs and check whether the host part ends with a list of allowed host suffixes?

jdno: That was the first approach that I considered, but then dropped when I came across the link to the GitHub issue and the `bit.ly` link. There are probably valid reasons to link to domains that we don't own, but in that case I would only want to allow specific URLs and not all of `bit.ly` or GitHub.

jdno:
I've [proposed](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Domains.20on.20GitHub) this to the infra-team, but there isn't enough support for the check. The only comment on the proposal was that teams can already add any URLs they want to the README, so adding this check for the repository URL doesn't have a lot of benefits.

PR #1442, opened 2024-05-03
Add `gsoc-contributors` team

Kobzol:
Adds a `gsoc-contributors` team with access to dev desktops. Eventually we should add all the GSoC participants, but for now it would be great if we could unblock people that need access to dev desktops.

CC @Oneirical if you want me to modify your details or name (we can also do that later).


jackh726 (approved):
This is a good start - as you said, we should eventually add all the participants.

PR #1461, opened 2024-05-24
Join the spec team

traviscross:
Joel Marcey, the owner of the spec work, proposed that I join the team, and the team confirmed his nomination with unanimous consent.

Correspondingly, in this commit, we add me (TC) to the spec team.

cc @JoelMarcey @m-ou-se @pnkfelix @ehuss


ehuss (approved):

jackh726 (approved):
Typically need a +1 from a team lead, but given the approvals here and the statement that there was unanimous consent, I'm going to take that as an approval.

PR #1467, opened 2024-05-27
Add TC to lang-docs

ehuss:
TC has volunteered to help with some lang-docs work.


ehuss:
cc @traviscross @rust-lang/lang-docs

jackh726 (approved):

PR #1483, opened 2024-06-24
Give the spec team write-access to the reference.

ehuss:
This gives the spec team permissions for the reference repository.

cc @rust-lang/spec

### DIFF
--- a/repos/rust-lang/reference.toml
+++ b/repos/rust-lang/reference.toml
@@ -7,6 +7,7 @@ bots = ["rustbot"]
 [access.teams]
 lang = "write"
 lang-docs = "write"
+spec = "write"
 
 [[branch-protections]]
 pattern = "master"


JoelMarcey (approved):

traviscross (approved):

rylev (approved):

traviscross:
Regarding the review requested of lang, we discussed this in the lang call today, and we had no objections and agreed that this is a spec team + rustdocs matter.

For broader context, this PR from Eric is part of a plan that we agreed to among the spec team for adopting the Reference as the spec.  The currently active maintainers of the Reference are all also current spec team members.

jackh726:
(Minutes link for reference: https://hackmd.io/JqUIKN5bShSP05t7FYAyug?view)

Its unclear to me from this whether or not the team (particularly Niko or Tyler as leads) explicitly +1ed this, or if this was just skipped over and assumed yes.

I'm not sure that it *really* matters in this case, but I'd rather not accidentally end up in a situation where anyone is surprised and, in the future, its probably easier for a lang team lead to explicitly +1 PRs.

jackh726 (approved):

PR #1484, opened 2024-06-24
Remove myself from wg-wasm.toml members

fitzgen:


teams/wg-wasm.toml
```diff
@@ -5,13 +5,13 @@ kind = "working-group"
 [people]
 leads = ["fitzgen"]
```
jackh726, line 6: I assume you also want to remove yourself here too?

fitzgen: Done

jackh726 (approved):

PR #1486, opened 2024-06-28
Create a `rustc_codegen_c` repository

tgross35:
One of the OSPP projects this year is a C codegen backend. Create a repository where this will be developed, modeled off the codegen_cranelift repo file.

r? @Kobzol since I think you are in the loop with this

### DIFF
--- a/repos/rust-lang/rustc_codegen_c.toml
+++ b/repos/rust-lang/rustc_codegen_c.toml
@@ -1,0 +1,9 @@
+org = "rust-lang"
+name = "rustc_codegen_c"
+description = "C based backend for rustc"
+bots = []
+
+[access.teams]
+codegen-c-maintainers = "maintain"
+compiler = "maintain"
+compiler-contributors = "write"
--- a/teams/codegen-c-maintainers.toml
+++ b/teams/codegen-c-maintainers.toml
@@ -1,0 +1,13 @@
+name = "codegen-c-maintainers"
+subteam-of = "compiler"
+
+[people]
+leads = []
+members = [
+    "Amanieu",
+    "tgross35"
+]
+alumni = []
+
+[[github]]
+orgs = ["rust-lang"]


repos/rust-lang/rustc_codegen_c.toml
```diff
@@ -0,0 +1,10 @@
+org = 'rust-lang'
+name = 'rustc_codegen_c'
+description = 'C based backend for rustc'
+bots = []
+
+[access.teams]
+compiler = 'maintain'
+
+[access.individuals]
+tgross35 = 'maintain'
```
ehuss, line 10: We have a policy of not giving individual access anymore. Can you find a team that would be appropriate to add here instead?

tgross35: Compiler makes sense (and is included). I'm just not a member of any relevant team but am the mentor for this project. Any suggestions?

Maybe I could join e.g. compiler contributors, but I have no clue what is required there.

ehuss: The only other suggestion I have is to create a subteam of compiler responsible for this backend. You might want to discuss with the compiler team.

tgross35:
I updated this to be both compiler and libs with contributors, which covers me. I think this is fine, if not then we might just need a new group.

@wesleywiser can you confirm?

Amanieu (approved):

wesleywiser (approved):

jackh726:
Not sure that adding libs/libs-contributors here is the right choice. I think the most lightweight option is to make a marker team like "ospp-2024-codegen-c" to add both you and potentially the ospp participant?

tgross35:
Sorry @jackh726, I missed your comment. I changed this to create a subteam `codegen-c-maintainers` (opted against including OSPP-2024 since this project will outlive it) with Amanieu (OSPP organizer) and myself (advisor for this project) as members. This team and `compiler` can maintain, `compiler-contributors` has write.

This is about the same as cg_gcc.

jackh726 (commented):
Going to go ahead and merge this. 

Am going to cc @wesleywiser and @davidtwco here - let me know if you have concerns

jackh726 (approved):

PR #1487, opened 2024-06-28
Grant myself try permissions

tgross35:
Some of my recent work has been flaky on certain platforms, and I would love to make use of the new `try-job`. Not sure what the rules for permissions are but it seems like maybe this is okay?

Also added perf because they are sometimes related (and I'm often online if others request a perf run). No specific need here though.

### DIFF
--- a/people/tgross35.toml
+++ b/people/tgross35.toml
@@ -2,3 +2,7 @@ name = "Trevor Gross"
 github = "tgross35"
 github-id = 13724985
 zulip-id = 532317
+
+[permissions]
+bors.rust.try = true
+perf = true


DianQK:
I think wg-binary-size can be granted try and perf permissions.

tgross35:
That works too, I only didn't propose it because the group is fairly inactive. @Kobzol any thoughts here?

Kobzol:
Adding try and perf to wg-binary-size seems fine to me.

tgross35:
That covers me, https://github.com/rust-lang/team/pull/1490

PR #1494, opened 2024-07-07
Add Liwa Gu

albertlarsan68:
She is an OSPP mentee

cc @Kobzol

### DIFF
--- a/people/liwagu.toml
+++ b/people/liwagu.toml
@@ -1,0 +1,4 @@
+name = "liwagu"
+github = "liwagu"
+github-id = 69970253
+zulip-id = 716586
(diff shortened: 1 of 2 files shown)


albertlarsan68:

rylev (changes requested):
Adding a person without adding them to a team is rejected by CI because it doesn't actually do anything. Was the idea to add them to this? 

https://github.com/rust-lang/team/blob/16fc8a96bf2733bc0e7ca553a645f3840ed0a7a4/teams/ospp.toml#L6

Kobzol:
They should be added to ospp-contributors, but it doesn't exist yet. It can be created in this PR I guess, as a copy paste of gsoc-contributors.

rylev (approved):

tgross35:
@rylev OSPP is getting into its final ~month. Would this be able to merge so we could add the other mentees?

jackh726 (approved):

PR #1499, opened 2024-07-16
Add Marco Ieni

MarcoIeni:


jdno (approved):
🥳

teams/infra-admins.toml
```diff
@@ -7,6 +7,7 @@ members = [
     "Mark-Simulacrum",
     "pietroalbini",
     "jdno",
+    "MarcoIeni",
```
rylev, line 10: Just want to draw attention to this since infra-admins have a high level of permissions. I can fully see how this is the intention in the medium term, but I wanted to make sure it's desired now.

jdno: Thanks for double checking, Ryan. It is indeed desired now. 🙂 One reason is that there is very little automation around `infra-admins` and we can stretch out provisioning of accounts manually. The other is that some level of immediate access is required to onboard and start working on our cloud infrastructure.

PR #1576, opened 2024-10-11
Add `merge-bots` configuration per branch protection

Kobzol:
Instead of just assuming that when bors (homu) is enabled for a repository, all its branches will be automatically managed by it, this PR adds an opt-in mechanism to use bors/homu for specific branch protections.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -352,4 +352,12 @@ required-approvals = 1
 # can push/merge to the branch.
 # (optional)
 allowed-merge-teams = ["awesome-team"]
+# Determines the merge queue bot(s) that manage pushes to this branch.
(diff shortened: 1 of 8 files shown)


MarcoIeni:
> all its repositories will be automatically managed by it

What do you mean?

A repo doesn't have multiple repositories.
Do you mean branches?

Kobzol:
Yes, I indeed meant branches, sorry =D

This PR shouldn't have any user-facing changes, so hopefully should be relatively safe. Once it lands, I'll change sync-team to read bors permissions from the branch protection instead from the array of bots. That should also be safe, because there is currently no repository in `team` that is managed by homu xD

MarcoIeni:
Merged because it only changes the schema and code 👍

PR #1590, opened 2024-10-22
Switch cargo from bors to merge-queue

Kobzol:
Accompanying PR: https://github.com/rust-lang/cargo/pull/14718

**Merge should be synchronized with an infra-admin switching the repo to merge queues!**

### DIFF
--- a/repos/rust-lang/cargo.toml
+++ b/repos/rust-lang/cargo.toml
@@ -2,13 +2,15 @@ org = "rust-lang"
 name = "cargo"
 description = "The Rust package manager"
 homepage = "https://doc.rust-lang.org/cargo"
-bots = ["bors", "rustbot", "rfcbot", "renovate"]
+bots = ["rustbot", "rfcbot", "renovate"]
 
 [access.teams]
 cargo = "write"
 
 [[branch-protections]]
 pattern = "master"
+ci-checks = ["conclusion"]
 
 [[branch-protections]]
 pattern = "rust-1.*"
+ci-checks = ["conclusion"]
--- a/people/Mark-Simulacrum.toml
+++ b/people/Mark-Simulacrum.toml
@@ -5,6 +5,3 @@ zulip-id = 116122
 irc = "simulacrum"
 email = "mark.simulacrum@gmail.com"
 discord-id = 442403469580566528
-
-[permissions]
-bors.cargo.review = true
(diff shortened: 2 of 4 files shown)


repos/rust-lang/cargo.toml
```diff
@@ -2,13 +2,14 @@ org = "rust-lang"
...
 homepage = "https://doc.rust-lang.org/cargo"
-bots = ["bors", "rustbot", "rfcbot", "renovate"]
+bots = ["rustbot", "rfcbot", "renovate"]
 
 [access.teams]
 cargo = "write"
 
 [[branch-protections]]
 pattern = "master"
+ci-checks = ["conclusion"]
```
ehuss, line 12: ci-checks will be needed on the `rust-1.*` branch protection, too. We use those for beta backports.

ehuss: Or...hm.. Merge queues don't support `*` patterns. That's...unfortunate.

Kobzol: If it's needed only for cargo, I would just configure it manually. TIL about the `*` limitation, that's annoying.

ehuss: Just checking, I think it would be good to also add ci-checks to the beta branch to ensure that nothing is merged with failures.

ehuss:
The cargo team discussed switching, and agreed we would like to try it out. I just have a minor caveat above, and then I think we should be good to go.

repos/rust-lang/cargo.toml
Kobzol: Changed.

repos/rust-lang/cargo.toml
```diff
@@ -2,13 +2,18 @@ org = "rust-lang"
...
+
+[[branch-protections]]
+pattern = "beta"
+ci-checks = ["conclusion"]
```
ehuss, line 16: Oh, sorry, to be clear, the pattern isn't "beta". It should be the `rust-1.*` pattern below. That's where we put our beta backports (it works a little differently than rust-lang/rust).

Kobzol: Right, sorry, my bad. Fixed :)

Kobzol:
We should have ideally someone from the Cargo team online for performing the switch (and we also need an infra-admin). If you're willing to help, @ehuss, let me and @MarcoIeni and we can coordinate the time.

ehuss:
I should be around tomorrow (Friday) if you want.

MarcoIeni:
I'm on a train for RustLab but it works for me. I'll be available in 4 hours from now for 7 hours (10.30 to 17.30 cet)

MarcoIeni (approved):
Approving and merging to switch cargo to merge queues

PR #1601, opened 2024-11-01
update membership for t-compiler/meeting

davidtwco:
It is preferred that the membership of this Zulip ping group be managed in an ad-hoc way, some members do not attend, and some members prefer not to be pinged so they get notified about subsequent direct pings, etc.

### DIFF
--- a/teams/compiler.toml
+++ b/teams/compiler.toml
@@ -110,3 +110,17 @@ name = "T-compiler"
 
 [[zulip-groups]]
 name = "T-compiler/meeting"
+include-team-members = false
+extra-people = [
+    "davidtwco",
+    "wesleywiser",
+    "apiraino",
+    "bjorn3",
+    "SparrowLii",
+    "estebank",
+    "BoxyUwU",
+    "jieyouxu",
+    "jswrenn",
+    "spastorino",
+    "tmiasko"
+]


teams/compiler.toml
```diff
@@ -107,6 +107,3 @@ include-subteam-members = true
 
 [[zulip-groups]]
 name = "T-compiler"
-
-[[zulip-groups]]
-name = "T-compiler/meeting"
```
Urgau, line 112: We should maybe set `include-team-members` to `false`
https://github.com/rust-lang/team/blob/8d055a5d1693107b41ddd4ea99dc646991e8700c/docs/toml-schema.md?plain=1#L189-L192
and use `extra-people` instead?

rylev: Setting `include-team-members` to `false`  would be the way it's "supposed to work" in the eyes of the team repo maintainers.

davidtwco: I've changed it to this.

teams/compiler.toml
```diff
@@ -110,3 +110,8 @@ name = "T-compiler"
 
 [[zulip-groups]]
 name = "T-compiler/meeting"
+include-team-members = false
+extra-people = [
+    "davidtwco",
+    "wesleywiser"
```
davidtwco, line 116: @rust-lang/compiler who wants to be pinged by `T-compiler/meeting`?

bjorn3: +1

SparrowLii: +1

estebank: +1

BoxyUwU: +1

jieyouxu: +1

spastorino: +1

davidtwco: Updated with everyone above

@apiraino: we should add to the team meeting agenda a small note that anyone who wants pinged for the meeting can make a PR to add to themselves to this list

wesleywiser (approved):

PR #1604, opened 2024-11-04
sync zulip stream membership

davidtwco:
See rust-lang/sync-team#92 for the other half of this.

t-compiler want to use our private Zulip stream for some communications but we've held off on doing this as the membership of that stream is manually managed, that is inconvenient and we sometimes forget. In testing this patch, I found the stream membership was still out-of-sync!

I've tested this locally with the sync-team changes locally and absent the untested parts noted in that PR's description, it all works.

I've deliberately avoided adding an exhaustive schema for the streams, just their name as that's all that is necessary to adjust the membership, and that's all that the compiler team needs.

There's a reasonable amount of duplication with the types and functions for Zulip user groups, but we'll likely extend these types to manage more about streams, I think that's okay. Everything I've added should be reusable and compatible with representing streams fully, but I'll leave that for other patches like rust-lang/team#1244.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -209,6 +209,37 @@ excluded-people = [
     "rylev",
 ]
 
+# Define the Zulip streams used by the team
+# It's optional, and there can be more than one.
+#
+# This will remove anyone who isn't in the team from the stream
+# so it should only be used for private streams at the moment.
+[[zulip-streams]]
+# The name of the Zulip stream (required)
+name = "t-overlords/private"
+# This can be set to false to avoid including all the team members in the stream
+# It's useful if you want to create the stream with a different set of members
+# It's optional, and the default is `true`.
+include-team-members = true
+# Include the following extra people in the Zulip stream. Their email address
+# or Zulip id will be fetched from their TOML in people/ (optional).
+extra-people = [
+    "alexcrichton",
+]
+# Include the following Zulip ids in the Zulip stream (optional).
+extra-zulip-ids = [
+    1234
+]
+# Include all the members of the following teams in the Zulip stream
+# (optional).
+extra-teams = [
+    "bots-nursery",
+]
+# Exclude the following people in the Zulip stream (optional).
+excluded-people = [
+    "rylev",
+]
+
 # Roles to define in Discord.
 [[discord-roles]]
 # The name of the role.
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -127,6 +127,26 @@ pub struct ZulipGroups {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ZulipStream {
+    pub name: String,
+    pub members: Vec<ZulipStreamMember>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum ZulipStreamMember {
+    // TODO(rylev): this variant can be removed once
+    // it is verified that no one is relying on it
+    Email(String),
+    Id(u64),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ZulipStreams {
+    pub streams: IndexMap<String, ZulipStream>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Permission {
     pub people: Vec<PermissionPerson>,
     pub github_users: Vec<String>,
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -179,6 +179,8 @@ pub(crate) struct Team {
     lists: Vec<TeamList>,
     #[serde(default)]
     zulip_groups: Vec<RawZulipGroup>,
+    #[serde(default)]
+    zulip_streams: Vec<RawZulipStream>,
     discord_roles: Option<Vec<DiscordRole>>,
 }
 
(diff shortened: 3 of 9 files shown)


ehuss:
Just so I'm clear, is the intention that this is only for private streams? I'm a little unclear, but it looks like it removes people if they are not explicitly listed in the team database, is that correct?

Can you add some docs to https://github.com/rust-lang/team/blob/master/docs/toml-schema.md documenting the new TOML structure?

davidtwco:
> Just so I'm clear, is the intention that this is only for private streams? I'm a little unclear, but it looks like it removes people if they are not explicitly listed in the team database, is that correct?
> 
> Can you add some docs to https://github.com/rust-lang/team/blob/master/docs/toml-schema.md documenting the new TOML structure?

Yeah, this would primarily be used for private streams.

davidtwco:
> Can you add some docs to https://github.com/rust-lang/team/blob/master/docs/toml-schema.md documenting the new TOML structure?

Added

docs/toml-schema.md
```diff
@@ -209,6 +209,37 @@ excluded-people = [
     "rylev",
 ]
 
+# Define the Zulip streams used by the team
+# It's optional, and there can be more than one.
+#
+# This will remove anyone who isn't in the team from the stream
+# so it should primarily be used for private streams.
```
MarcoIeni, line 216: I don't see a `private` boolean or similar. Does that mean that all zulip-streams are private?
if yes, should we change this comment?
```suggestion
# so it should only be used for private streams.
```

MarcoIeni:
The `static_api` test is failing in CI.

docs/toml-schema.md
davidtwco: I'll update the comment - there's no `private` boolean at the moment because all this will do initially is add and remove members, it can be extended later to create streams (where we'd want to specify a stream is private) and to only add members (so it makes sense to use it on a public stream).

davidtwco:
Fixed the test and removed a bunch of the duplicated code

docs/toml-schema.md
```diff
@@ -209,6 +209,37 @@ excluded-people = [
...
+# This will remove anyone who isn't in the team from the stream
+# so it should only be used for private streams at the moment.
+[[zulip-streams]]
+# The name of the Zulip stream (required)
+name = "t-overlords/private"
+# This can be set to false to avoid including all the team members in the stream
+# It's useful if you want to create the stream with a different set of members
+# It's optional, and the default is `true`.
+include-team-members = true
+# Include the following extra people in the Zulip stream . Their email address
```
MarcoIeni, line 224:
```suggestion
# Include the following extra people in the Zulip stream. Their email address
```

rust_team_data/src/v1.rs
```diff
@@ -126,6 +126,26 @@ pub struct ZulipGroups {
...
+pub struct ZulipStream {
+    pub name: String,
+    pub members: Vec<ZulipStreamMember>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum ZulipStreamMember {
+    // TODO(rylev): this variant can be removed once
+    // it is verified that noone is relying on it
```
MarcoIeni, line 139:
```suggestion
    // it is verified that no one is relying on it
```
right?

src/schema.rs
```diff
@@ -368,6 +370,49 @@ impl Team {
         Ok(lists)
     }
 
+    fn expand_zulip_membership(
```
MarcoIeni, line 373:
```suggestion
    /// `on_exclude_not_included` is a function that is returned when an excluded member
    /// wasn't included.
    fn expand_zulip_membership(
```

src/validate.rs
```diff
@@ -733,6 +735,30 @@ fn validate_zulip_group_ids(data: &Data, errors: &mut Vec<String>) {
     });
 }
 
+/// Ensure every member of a team that has a Zulip stream either has a Zulip id
+fn validate_zulip_stream_ids(data: &Data, errors: &mut Vec<String>) {
+    wrapper(data.teams(), errors, |team, errors| {
+        let streams = team.zulip_streams(data)?;
+        // Returns if stream is empty or all the groups don't include the team members
```
MarcoIeni, line 742:
```suggestion
        // Returns if stream is empty or all the streams don't include the team members
```

src/validate.rs
```diff
@@ -733,6 +735,30 @@ fn validate_zulip_group_ids(data: &Data, errors: &mut Vec<String>) {
     });
 }
 
+/// Ensure every member of a team that has a Zulip stream either has a Zulip id
```
MarcoIeni, line 738:
```suggestion
/// Ensure team members in Zulip streams have a Zulip id
```
I think this reads more clear. If you like it, can you also update the function `validate_zulip_group_ids`?

MarcoIeni:
Overall looks good to me! I just left a few minor comments 👍

docs/toml-schema.md
davidtwco: Fixed!

rust_team_data/src/v1.rs
davidtwco: Fixed!

src/schema.rs
davidtwco: Fixed!

src/validate.rs
davidtwco: Fixed!

MarcoIeni (approved):
This looks good from my perspective.
Maybe we can merge it once https://github.com/rust-lang/sync-team/pull/92 is also approved?

davidtwco:
> Do you agree to merge this first and merge [rust-lang/sync-team#92](https://github.com/rust-lang/sync-team/pull/92) after?

Works for me

PR #1618, opened 2024-12-03
Add niko to spec team

m-ou-se:
@nikomatsakis @rust-lang/spec

PR #1619, opened 2024-12-03
Add Niko to the spec team

traviscross:
We discussed this on a spec call 2024-12-03, and we're all in enthusiastic agreement for adding Niko to the team.

cc @rust-lang/spec @nikomatsakis


traviscross:
Looks like @m-ou-se and I raced.  Closing in favor of #1618.

PR #1620, opened 2024-12-04
Define a team for Project Directors on the Foundation board

carols10cents:
So that:

- Project Directors can post to Inside Rust as "on behalf of the Project Directors team"
- We can have an email address for people to contact all of us
- ~~We can be listed on rust-lang.org~~ I don't think this is true now that I've changed the team to be type marker-team

I've probably forgotten something that should be set in the toml or that needs to be done when adding a new team. Please let me know and I'm happy to fix! ❤️

### DIFF
--- a/teams/foundation-board-project-directors.toml
+++ b/teams/foundation-board-project-directors.toml
@@ -1,4 +1,4 @@
-name = "project-directors"
+name = "foundation-board-project-directors"
 kind = "marker-team"
 
 [people]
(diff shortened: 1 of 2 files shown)


carols10cents:
```
[ERROR rust_team::validate] validation error: team `project-directors` must specify `subteam-of` or top-level=true
```

Well we're not a top-level team with a rep on the leadership council. So I decided to go with marker-team? I think this doesn't put us on the website though, which I guess is fine, but can we still post to Inside Rust? (That's the most important thing to me).

Mark-Simulacrum (approved):

carols10cents:
Thank you @Mark-Simulacrum !

PR #1628, opened 2024-12-15
Add `rfcbot` to compiler-team repo

Urgau:
Add `rfcbot` to the list of bots to the compiler-team reposiroty so that the bot can edit labels.

cf. https://www.github.com/rust-lang/compiler-team/issues/785#issuecomment-2544090585

### DIFF
--- a/repos/rust-lang/compiler-team.toml
+++ b/repos/rust-lang/compiler-team.toml
@@ -2,7 +2,7 @@ org = "rust-lang"
 name = "compiler-team"
 description = "A home for compiler team planning documents, meeting minutes, and other such things."
 homepage = "https://rust-lang.github.io/compiler-team/"
-bots = ["rustbot"]
+bots = ["rustbot", "rfcbot"]
 
 [access.teams]
 compiler = "write"


apiraino:
In [triagebot#1747](https://github.com/rust-lang/triagebot/pull/1747) (and then fixed in [triagebot#1863](https://github.com/rust-lang/triagebot/pull/1747)) I have added the syntax `@rfcbot concern` to issue a fake command to register MCP concerns. Currently that command is a "no-op", just a device to let me keep track of MCP concerns.

I wonder what happens if the `@rfcbot` will be enabled for real. I guess we will discover when this PR is merged.

Urgau:
`rfcbot` != `rustbot` (aka `triagebot`), I don't think there will be any issue.

apiraino:
> `rfcbot` != `rustbot` (aka `triagebot`), I don't think there will be any issue.

Ah ok, so that means that the invocation of `@rfcbot concern`[^1] is handled by another codebase https://github.com/rust-lang/rfcbot-rs

[^1]: looking into the issues :thinking:

jackh726 (approved):

PR #1637, opened 2025-01-06
Adopt rust-by-example as a subteam of lang-docs

traviscross:
We're working to move things out from under the launching-pad team to more appropriate places.  It makes the most sense for the rust-by-example working group to be adopted as a subteam of lang-docs, so let's do that.

Note that, in doing this, we'll make it a proper subteam rather than a working group.

See:

- https://github.com/rust-lang/leadership-council/issues/139

cc @ehuss @marioidival @jamesmunns

### DIFF
--- a/repos/rust-lang/rust-by-example.toml
+++ b/repos/rust-lang/rust-by-example.toml
@@ -5,4 +5,4 @@ homepage = "https://doc.rust-lang.org/stable/rust-by-example/"
 bots = ["rustbot"]
 
 [access.teams]
-wg-rust-by-example = "write"
+rust-by-example = "write"
--- a/teams/rust-by-example.toml
+++ b/teams/rust-by-example.toml
@@ -1,6 +1,5 @@
-name = "wg-rust-by-example"
-subteam-of = "launching-pad"
-kind = "working-group"
+name = "rust-by-example"
+subteam-of = "lang-docs"
 
 [people]
 leads = ["marioidival"]
@@ -16,6 +15,6 @@ alumni = [
 orgs = ["rust-lang"]
 
 [website]
-name = "Rust by Example working group"
+name = "Rust by Example team"
 description = "Maintaining and updating the official Rust by Example book"
 repo = "https://github.com/rust-lang/rust-by-example"


teams/wg-rust-by-example.toml
```diff
@@ -1,5 +1,5 @@
 name = "wg-rust-by-example"
-subteam-of = "launching-pad"
+subteam-of = "lang-docs"
 kind = "working-group"
```
ehuss, line 3: I think while we're here making changes, it would make sense to change it to a normal team (and drop the `wg-` prefix). "working group" doesn't really make too much sense here, and everything else is a team.

(or we can do that in a separate PR)

traviscross: Makes sense.  I had almost done that originally.  Done now.

traviscross:

The first bit of context is that I'm making this PR with my lang-docs, lang, and lang-ops hats on -- that is, the hat of the adopting team, the hat of its parent team, and the hat of the team that supports that parent team and tends to handle many administrative and policy matters like this -- not with my council hat on.

Regarding the lang-docs hat, while, in practice, @ehuss and I tend to handle much of the coordination work for lang-docs, this should of course be approved by one or both of the current leads for lang-docs, @ehuss and/or @JohnTitor.  This PR can be seen as my suggestion to my fellow team members about the appropriate handling.  As it is, @ehuss and I had discussed this ahead of time.

On the council side, @jamesmunns is of course the one coordinating this, and is of course in my view doing a great job in that.  Thanks to James for opening the tracking issues such as the one [here](https://github.com/rust-lang/leadership-council/issues/139), for bringing the various parties together, and for updating the council on all the various progress on this endeavor.

JohnTitor (approved):

ehuss (approved):

nikomatsakis:
I'm in favor of this move in general. I would however like to ask that there be some kind of charter -- it doesn't have to be a lot -- documenting the role of each team. As the family of lang subteams grows, it occurs to me that we should probably go through and host these charters somewhere, I'm imagining lang-team.rust-lang.org.

As an example of what I have in mind, I wrote up a brief charter for the proposed goals team here:

https://rust-lang.github.io/rust-project-goals/admin/team.html

traviscross:
Proposed charter:

> The Rust by Example team is responsible for maintaining and updating the official Rust by Example book.
>
> This team maintains the following repositories:
>
> - https://github.com/rust-lang/rust-by-example
>
> Members of this team both perform original authorship on the Rust by Example book and oversee reviewing and accepting the proposed changes from others.  The team works to expand the book with new content where appropriate, with an eye toward best helping new users to learn Rust and existing users to understand Rust better, and to keep the book up to date with changes in the languages and with the ongoing and changing needs of our users.

jamesmunns (approved):
Thank you all! @rust-lang/team-repo-admins could we please get an r+ and merge?

rylev (approved):

PR #1638, opened 2025-01-06
Adopt rustlings team as a subteam of lang-docs

traviscross:
We're doing some minor reorganization.  As part of that, we're moving the book team and rust-by-example to be subteams of lang-docs. Given that, it makes sense for lang-docs to also adopt rustlings as a subteam, so let's do that.

Note that, unlike the other two, rustlings had been notated as a marker team rather than a subteam of launching-pad.  This change will make it into a normal subteam, which seems more appropriate.

As discussed in the PR (https://github.com/rust-lang/team/pull/1638#discussion_r1904685144), apparently mo8it has been maintaining rustlings solo for about nine months, so this PR also makes him a lead of the team.

cc @ehuss @shadows-withal @mo8it @jamesmunns

### DIFF
--- a/teams/rustlings.toml
+++ b/teams/rustlings.toml
@@ -2,7 +2,7 @@ name = "rustlings"
 subteam-of = "lang-docs"
 
 [people]
-leads = ["shadows-withal"]
+leads = ["mo8it", "shadows-withal"]
 members = [
     "shadows-withal",
     "mo8it",


people/mo8it.toml
```diff
@@ -1,3 +1,4 @@
 name = "Mo"
 github = "mo8it"
 github-id = 76752051
+email = false
```
mo8it, line 4:
```toml
email = "mo8it@proton.me"
zulip-id = 632194
```

teams/rustlings.toml
```diff
@@ -1,5 +1,5 @@
 name = "rustlings"
-kind = "marker-team"
+subteam-of = "lang-docs"
 
 [people]
 leads = ["shadows-withal"]
```
mo8it, line 5: Might be a good point to make me a lead since I am maintaining Rustlings alone for about 9 months now :D

mo8it (changes requested):

people/mo8it.toml
traviscross: Thanks; added.

teams/rustlings.toml
traviscross: Yes, that would make sense then.  Updated.

mo8it (approved):

traviscross:

The first bit of context is that I'm making this PR with my lang-docs, lang, and lang-ops hats on -- that is, the hat of the adopting team, the hat of its parent team, and the hat of the team that supports that parent team and tends to handle many administrative and policy matters like this -- not with my council hat on.

Regarding the lang-docs hat, while, in practice, @ehuss and I tend to handle much of the coordination work for lang-docs, this should of course be approved by one or both of the current leads for lang-docs, @ehuss and/or @JohnTitor.  This PR can be seen as my suggestion to my fellow team members about the appropriate handling.  As it is, @ehuss and I had discussed this ahead of time, and he has approved it in https://github.com/rust-lang/team/pull/1638#pullrequestreview-2532978123.

On the council side, @jamesmunns is of course the one coordinating this, and is of course in my view doing a great job in that.  Thanks to James for opening the tracking issues, for bringing the various parties together, and for updating the council on all the various progress on this endeavor.

JohnTitor (approved):
👍, thank you for your work!

shadows-withal (approved):
sgtm

rylev (approved):

PR #1639, opened 2025-01-06
Adopt wg-async as a subteam (WG) of lang

traviscross:
We're working to move things out from under the launching-pad team to more appropriate places.  It makes the most sense for the async working group to be adopted as a subteam of lang, so let's do that.

Note that there are ongoing discussions of how we on lang might want to further refactor this working group.  This PR doesn't seek to resolve those.  It just puts the working group in our domain.  We can later adjust as we deem proper.

See:

- https://github.com/rust-lang/leadership-council/issues/137

cc @nikomatsakis @tmandry @eholk @yoshuawuyts @compiler-errors @jamesmunns

### DIFF
--- a/teams/wg-async.toml
+++ b/teams/wg-async.toml
@@ -1,5 +1,5 @@
 name = "wg-async"
-subteam-of = "launching-pad"
+subteam-of = "lang"
 kind = "working-group"
 
 [people]


eholk:
This seems like a good way to make this reorganization tractable. We can start by getting it out of the launching pad, but any more substantive changes can happen later.

traviscross:

The first bit of context is that I'm making this PR with my lang/lang-ops hat on -- that is, the hat of the adopting team (and the ops team that supports it in many administrative matters like this) -- and not with my council hat on.  This PR should of course be approved by @nikomatsakis and/or @tmandry, who were CCed above, and who are incidentally the leads of both lang and of WG-async, and this can be seen as my suggestion to my fellow team members about the appropriate handling.  As mentioned, there have been ongoing discussions within the lang team about how to refactor the async working group, and this PR does not seek to resolve those discussions, but simply to take the step of putting it within our domain, leaving us of course later free to make other adjustments.

On the council side, @jamesmunns is of course the one coordinating this, and is of course in my view doing a great job in that.  Thanks to James for opening the tracking issues, such as the one [here](https://github.com/rust-lang/leadership-council/issues/137), for bringing the various parties together, and for updating the council on all the various progress on this endeavor.

ehuss:
Just a heads up that this will delete the page https://www.rust-lang.org/governance/wgs/wg-async.

If we don't want to break all the links on the internet, we could add a redirect to https://github.com/rust-lang/www.rust-lang.org/blob/HEAD/src/redirect.rs.

traviscross:
Draft PR for the redirect up in:

- https://github.com/rust-lang/www.rust-lang.org/pull/2106

tmandry (approved):
Moving it under lang makes sense; let's just get it out of the launching pad for now. Thanks for the suggestion @traviscross, and thanks @jamesmunns for driving the overall effort.

nikomatsakis (approved):

jamesmunns:
Thanks all! Seems like everyone is on board, @rylev or @Mark-Simulacrum, would you mind assisting with an r+ and merge?

rylev (approved):

PR #1642, opened 2025-01-13
Fix marcoieni username change

ehuss:
The validation is complaining about the case change on this username. I didn't even know it was possible to change your username like that.

### DIFF
--- a/people/marcoieni.toml
+++ b/people/marcoieni.toml
@@ -1,5 +1,5 @@
 name = "Marco Ieni"
-github = "MarcoIeni"
+github = "marcoieni"
 github-id = 11428655
 email = "marcoieni@rustfoundation.org"
 discord-id = 650270302198169611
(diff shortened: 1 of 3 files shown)


ehuss:
cc @marcoieni

people/MarcoIeni.toml
```diff
@@ -1,5 +1,5 @@
 name = "Marco Ieni"
-github = "MarcoIeni"
+github = "marcoieni"
```
Kobzol, line 2: The github username should ideally match the name of the file.

ehuss: Good point, thanks!

jackh726 (approved):

marcoieni:
Sorry about this! I did it because too many people thinks my surname is *Leni* instead of *ieni* 😅

> I didn't even know it was possible to change your username like that.

Yes, you can do it from GitHub settings. 👍

ehuss:
Curious, how do you pronounce Ieni?

marcoieni:
The google translate pronunciation is correct 👍
https://translate.google.com/?sl=it&tl=en&text=ieni&op=translate

image

PR #1643, opened 2025-01-13
Add a website page for the launching pad

ehuss:
This adds a dedicated website page for the launching pad. See https://github.com/rust-lang/www.rust-lang.org/pull/2105 for more information.

### DIFF
--- a/teams/launching-pad.toml
+++ b/teams/launching-pad.toml
@@ -6,3 +6,9 @@ top-level = true
 leads = []
 members = []
 alumni = []
+
+[website]
+page = "launching-pad"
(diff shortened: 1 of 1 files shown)


ehuss:
This is a draft since we need to determine if we want to do this.

ehuss:
@rust-lang/team-repo-admins I think I'd like to move forward with this.

cc @jamesmunns

jamesmunns:
@ehuss I don't have a specific preference, this looks like a fine change to me. Having a page for the launching pad seems reasonable!

Mark-Simulacrum (approved):

PR #1649, opened 2025-01-28
Add types-fcp team and remove exclusion list from types team

oli-obk:
As discussed in today's T-types meeting. This mirrors the structure the compiler team uses.

### DIFF
--- a/teams/types-fcp.toml
+++ b/teams/types-fcp.toml
@@ -1,0 +1,20 @@
+name = "types-fcp"
+subteam-of = "types"
+
+[people]
+leads = []
+members = [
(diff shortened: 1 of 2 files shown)


rylev (approved):

jackh726:
This shouldn't have been merged yet.

First, from what I can tell, there wasn't actually consensus in our private thread whether this should be a new team or maintain the previous one-team with just "opt-out".

Second, this should have had explicit approval by one of lcnr or I as team leads.

jackh726:
That being said, this is not actively harmful to the point that it's worth going through the work of reverting it until actual consensus is reached.

PR #1658, opened 2025-02-17
Rust for Linux: remove Wedson

ojeda:
Wedson does not work on Rust for Linux anymore, thus remove him from the ping group.

Cc: @wedsonaf


ojeda:
Apparently I should also remove his `people/` entry to pass the CI since he is not anymore in any group, so I will do that.

rylev (approved):

PR #1660, opened 2025-02-17
Archive the core team

ehuss:
This archives the core team and the core-team repo.

I am not aware of any infrastructure still depending on this. If there is, it would be good to know so it can be resolved.

None of the mailing list aliases look like they are particularly important to keep around. I do not see `legal@crates.io` used anywere, and I assume Mark and Ryan have not been getting any significant email as of recently. If anyone really cares, I'm fine with moving them to council instead.

### DIFF
--- a/repos/archive/rust-lang/core-team.toml
+++ b/repos/archive/rust-lang/core-team.toml
@@ -4,4 +4,3 @@ description = "A place to house minutes and other documents related to the core 
 bots = []
 
 [access.teams]
-core = "maintain"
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
Plenty of email, but with the spam volume hard to find signal in it. I think it's fine to do this, though I suspect that we'd be losing some email pointers somewhere if we drop "legal@crates.io" and "user-logos@rust-lang.org" in particular. But web search isn't finding anything either there, so 🤷 seems ok. If someone really wants to reach us I think there's clear ways to do that.

I'll go ahead and merge, we can always revert AFAICT.

Mark-Simulacrum (approved):

PR #1663, opened 2025-02-19
Add dry run workflow for testing the applied changes

Kobzol:
This PR adds a dry run CI workflow that runs after every push to a PR. On a PR push, a GitHub artifact with its rendered JSON API is uploaded. When the PR workflow finishes successfully, the dry run workflow will run on `main` (in a privileged context). It downloads `sync-team`, downloads the built JSON API from the PR, passes it to `sync-team` and gathers its dry run output.

It then hides all previous bot comments on that PR and posts a new [comment](https://github.com/Kobzol/team/pull/1#issuecomment-2668621163) that prints the output of the dry run.

Using this workflow, we should be able to quickly and easily observe what changes would happen after a given team PR is merged.

Blocked on
- [x] https://github.com/rust-lang/sync-team/pull/100.
- [ ] Configuring a read-only GitHub (and ideally also Zulip) token for the sync-team dry run.

(Very painfully) debugged at https://github.com/Kobzol/team/pull/1#issuecomment-2668621163.

### DIFF
--- a/.github/workflows/dry-run.yml
+++ b/.github/workflows/dry-run.yml
@@ -1,0 +1,88 @@
+# This workflow executes a dry-run of the sync-team tool after a push to any pull request.
+# This allows us to see which changes would be applied to live services after the PR
+# would be merged.
+#
+# The workflow uses the `workflow_run` trigger, which should always run in the default branch of
+# this repository. This is required so that the workflow has permissions to post PR comments.
+# We should not check out any code from the PR, as it could present a security hazard.
+# Instead, we simply download a GitHub artifact with a directory of JSON files and use that as
+# input for sync-team.
+# This artifact is created and uploaded on PR pushes using the CI workflow in `main.yml`.
+# Details about `workflow_run`:
+# https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run:
+name: sync-team dry-run
+
+on:
+  workflow_run:
+    workflows: [ CI ]
+    types:
+      - completed
+
+concurrency:
+  # Only run this once at a time on any given PR
+  group: dry-run-${{ github.event.workflow_run.pull_requests[0].number }}
+  cancel-in-progress: true
+
+jobs:
+  dry-run:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request' }}
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: rust-lang/sync-team
+          persist-credentials: false
+
+      - name: Install Rust Stable
+        run: |
+          rustc -vV
+          rustup update stable
+          rustup default stable
+          rustc -vV
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Download built JSON API
+        uses: actions/download-artifact@v4
+        with:
+          name: team-api-output
+          path: team-api
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run sync-team dry-run check
+        run: |
+          # Perform build and execution separately to avoid any potential output from
+          # cargo leaking into the output file.
+          cargo build ./target/debug/sync-team print-plan --team-json team-api > output.txt
+
+      - name: Prepare comment
+        run: |
+          cat > comment.txt << EOL
+          <details>
+          <summary>Dry-run check results</summary>
+          <pre><code>
+          EOL
+          cat output.txt >> comment.txt
+          printf "</pre></code>\n</details>\n" >> comment.txt
+
+          cat comment.txt
+
+      - name: Send comment
+        env:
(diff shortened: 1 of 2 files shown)


Kobzol:
This still requires the read-only tokens to be configured, but otherwise should be ready for a review.

.github/workflows/dry-run.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Run sync-team dry-run check
+        run: |
+          cargo build
+          ./target/debug/sync-team print-plan --team-source team-api > output.log
+
+      - uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
```
marcoieni, line 45: This script is a bit complex. What do you think about storing the script in a separate file and invoking it from here?
Instructions [here](https://github.com/actions/github-script?tab=readme-ov-file#run-a-separate-file-with-an-async-function)

Having the script in a separate file will have various benefits, such as syntax highlight and IDE autocomplete

.github/workflows/dry-run.yml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          name: team-api-output
+          path: team-api
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run sync-team dry-run check
+        run: |
+          cargo build
+          ./target/debug/sync-team print-plan --team-source team-api > output.log
```
marcoieni, line 40: why not simply `cargo run` 🤔

marcoieni:
This approach is flagged as not secure by [zizmor](https://woodruffw.github.io/zizmor/audits/#dangerous-triggers).

We are confident that here there won't be issues because CI will never run the `dry-run.yml` from the fork, but only the one from the main branch, right?

We should write some comments to explain the risks and justify why we are using this approach 👍

Kobzol:
Wut, `workflow_run` was explicitly [designed](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) to avoid the problems with `pull_request_target` :confused:

.github/workflows/dry-run.yml
```diff
@@ -0,0 +1,87 @@
...
+jobs:
+  dry-run:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request' }}
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: rust-lang/sync-team
```
marcoieni, line 18:
```suggestion
          repository: rust-lang/sync-team
          persist-credentials: false
```
If we don't use the git token we can set this for additional security

marcoieni:
> Wut, `workflow_run` was explicitly [designed](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) to avoid the problems with `pull_request_target` 😕

Right. I think it's fine, we should just justify our choices.

.github/workflows/dry-run.yml
Kobzol: I can auto-inject syntax highlight and autocomplete into the snipper in my IDE (CLion) :grin:

We would have to checkout the team repo for that. I don't have a problem with that, but it could be one additional potential attack vector.

.github/workflows/dry-run.yml
Kobzol: Just to make sure that no stdout leaks from `cargo` into the log file.

Kobzol:
Added explanation :)

.github/workflows/dry-run.yml
marcoieni: oooh, I see, nevermind 👍

.github/workflows/dry-run.yml
marcoieni: can you add a comment about this?

Kobzol:
https://github.com/rust-lang/sync-team/pull/103 well.. it's probably not worth it, due to GHA limitations :(

.github/workflows/dry-run.yml
Kobzol: Done :)

Kobzol:
Removed the complex script and removed it with (hopefully slightly less complex) `gh cli` + bash-fu.

marcoieni:
LGTM 👍
Should I add the GitHub app token in the GitHub Actions secrets?
For the zulip token api, I need to create a new one with read-only access probably. Also, this zulip token is not set for the sync-team dry-run workflow. Should we also set it there? 🤔

marcoieni:
I added the secrets `SYNC_TEAM_GH_APP_ID` and `SYNC_TEAM_GH_APP_PRIVATE_KEY`.
I will now push to this PR to edit the CI workflow to generate the tokens.

marcoieni:
Should we merge this?

Kobzol:
Since the workflow runs from master, it's quite hard to test in production without merging. So yeah, I would try it, possibly there will be some issues that we can fix incrementally. The comment contents are collapsed by default, so even if there is a lot of output, it should not be too annoying.

marcoieni (approved):

marcoieni:
done. Will you open a PR to test this?

PR #1669, opened 2025-02-28
Add review permissions for rustfmt team members

ytmimi:
These review permissions will only be used for rustfmt subtree syncs and the occasional in-tree rustfmt PR.

### DIFF
--- a/teams/rustfmt.toml
+++ b/teams/rustfmt.toml
@@ -16,7 +16,7 @@ alumni = [
 [[github]]
 orgs = ["rust-lang"]
 
-[leads-permissions]
+[permissions]
 # Primarily for subtree syncs and other rustfmt-related review.
 bors.rust.review = true
 


compiler-errors:
I see no reason why T-rustfmt shouldn't have review permissions for rust-lang/rust. Rustfmt is a subtree, and afaict all other subtree owners have review permissions for at *least* subtree syncs, but also for in-tree reviews for when it's necessary :)

jackh726 (commented):
Am I misreading or misremembering something? Shouldn't this go in `teams/rustfmt.toml`? (edit: yes, CI is failing, I'm not crazy)

Rustfmt leads already have permission (https://github.com/rust-lang/team/blob/18d2963ea88b9d0c4d698eb9bf57c8985cdc5cb9/teams/rustfmt.toml#L19). Looks like originally it was the entire team and then got limited to leads (https://github.com/rust-lang/team/pull/610).

I personally think it's okay to just give the whole team permissions, but maybe @calebcartwright  or @ytmimi can chime in.

compiler-errors:
Oh yes, this is a typo bc this should go in the teams file, not the repo.

As for the previous decision about only allowing the lead to have bors permissions, this is something that was decided quite some years ago, and it definitely warrants revisiting today.

ytmimi:
@jackh726 @compiler-errors thanks for pointing out that this should be in the teams file instead. To be honest I didn't know what I was doing 😅. I'll make that update right now!

calebcartwright (approved):
Support this change. The current leads-only setting was done in the context where the team was only 1 person and I was heavily trying to recruit people to the rustfmt team to help, some of whom were very new to the project, wouldn't have needed permissions in r-l/rust and shouldn't have had them.

jackh726 (approved):

jackh726:
Seems like a clear decision.

PR #1676, opened 2025-03-04
Move Nemo157 and Byron into docs.rs alumnis

GuillaumeGomez:
This is a sad news but they didn't contribute in a long time, so I think it's time for us to let them go.

Thanks a lot for your contributions!

Please note though that if you ever want to contribute again on docs.rs, you're more than welcome and if it happens, you'll be fast-tracked back on the team. :)

cc @syphar @Nemo157 @Byron

### DIFF
--- a/teams/docs-rs-reviewers.toml
+++ b/teams/docs-rs-reviewers.toml
@@ -5,10 +5,11 @@ subteam-of = "docs-rs"
 leads = []
 members = [
     "GuillaumeGomez",
-    "Nemo157",
     "syphar",
 ]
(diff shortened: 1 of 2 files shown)


rylev (approved):

Byron:
Thanks a lot for releasing me, it was my honour :)! 

I was mostly a satellite for maintaining at least one crate that was (is?) important to `docs.rs`, but there also wasn't much to do nor much intersection with the actual team. So me being on the team officially might have been too much to begin with, making this action a welcome correction.

PR #1681, opened 2025-03-06
Fix dry run workflow for PR from forks

Kobzol:
GitHub strikes again.. We need to get the PR number into the dry-run workflow somehow. This seemed like the simplest possibility (https://github.com/orgs/community/discussions/25220).

### DIFF
--- a/.github/workflows/dry-run.yml
+++ b/.github/workflows/dry-run.yml
@@ -150,11 +150,24 @@ jobs:
 
           cat comment.txt
 
+      - name: Extract PR number
+        run: |
+          # We read the PR number that is stored in the uploaded archive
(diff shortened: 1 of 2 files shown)


.github/workflows/dry-run.yml
```diff
@@ -42,7 +42,8 @@ jobs:
           rustup default stable
           rustc -vV
 
-      - uses: Swatinem/rust-cache@v2
+      # Pinning a specific version to avoid surprises
+      - uses: Swatinem/rust-cache@27b8ea9368cf428f0bfe41b0876b1a7e809d9844
```
marcoieni, line 46: you might also want to pin this action in other workflows

PR #347, opened 2020-05-28
Add windows group

spastorino:
r? @nikomatsakis

PR #1685, opened 2025-03-06
Add `codegen-{cranelift,gcc}` marker teams for custom codegen backend ping groups

jieyouxu:
Part of https://github.com/rust-lang/rust/issues/138116.
Discussion: [#t-infra > Triagebot ping groups for GCC and cranelift backends](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Triagebot.20ping.20groups.20for.20GCC.20and.20cranelift.20backends/with/503862444).
rust-lang/rust side triagebot configuration + docs: https://github.com/rust-lang/rust/pull/138120

cc @GuillaumeGomez @bjorn3 @antoyo

### DIFF
--- a/teams/codegen-cranelift.toml
+++ b/teams/codegen-cranelift.toml
@@ -1,0 +1,8 @@
+name = "codegen-cranelift"
+kind = "marker-team"
+
+[people]
+leads = []
+members = [
+    "bjorn3",
+]
--- a/teams/codegen-gcc.toml
+++ b/teams/codegen-gcc.toml
@@ -1,0 +1,9 @@
+name = "codegen-gcc"
+kind = "marker-team"
+
+[people]
+leads = []
+members = [
+    "antoyo",
+    "GuillaumeGomez",
+]


Urgau:
Why I we creating marker teams? We already have a wg-gcc-backend working group.

I don't see one for cranelift, but we should do the same and create one instead of creating marker teams.

jieyouxu:
> Why are we creating marker teams? We already have a wg-gcc-backend working group.

For ping groups. You can't ping rust-lang/wg-gcc-backend if u're not an org member.

Urgau:
Oh, the llvm one is actually for icebreakers-llvm and not wg-llvm. Fair enough.

I still wish we didn't have to create marker teams.

jieyouxu:
Oh I raced with ur comment here over at the r-l/r side PR lol.

Yeah, I don't think you can use a real wg team as ping group.

jieyouxu:
Friendly ping @rust-lang/team-repo-admins

jieyouxu:
> I don't think you can ping a marker group if you're not a team member, either.

Yes, the reason for the marker-group is so it can be registered through r-l/r triagebot, see for instance https://github.com/rust-lang/rust/blob/65899c06f117ddac9c8399479ddcdc122c92fddf/triagebot.toml#L49-L58. That way, non-org members can ping through triagebot via `@rustbot ping codegen-{cranelift,gcc}`. They can be *aliases* on the r-l/r triagebot side for e.g. `@rustbot ping cg_gcc` if people want that.

> I'm not quite clear why these would be created. It sounds like Ralf couldn't remember the team names, but I'm not sure how you would remember the ping group names, either. Also you can just type @gcc, and GitHub's autoomplete's first entry is the team you want to notify.

This is not intended primarily for org members who can already ping `@rust-lang/*` teams, but for non-org members through triagebot.

The intention is not to "remember them", because they will be documented in rustc-dev-guide, and CI job failure logs can backlink to rustc-dev-guide dedicated test job/component pages (in this case, custom codegen backend pages) and/or dedicated ping group pages. See the r-l/r side companion PR https://github.com/rust-lang/rust/pull/138120.

jieyouxu:
> Oh, the llvm one is actually for icebreakers-llvm and not wg-llvm. Fair enough.

FTR, I removed the `@rustbot ping llvm` alias in https://github.com/rust-lang/rust/pull/138735

jieyouxu:
See for example https://github.com/rust-lang/team/pull/347 which adds a "windows" marker-team for r-l/r side ping group.

ehuss:
> That way, non-org members can ping through triagebot via `@rustbot ping codegen-{cranelift,gcc}`

I'm pretty sure only team members can ping: https://github.com/rust-lang/triagebot/blob/d45a79f6a37a9cbd84b4160a13020f7dacf9f3f3/src/handlers/ping.rs#L28-L35

jieyouxu:
... Wait, what?

PR #1686, opened 2025-03-06
Also print sync-team output to CI logs

Kobzol:
So that we can debug [errors](https://github.com/rust-lang/team/actions/runs/13705594955/job/38329976933).

### DIFF
--- a/.github/workflows/dry-run.yml
+++ b/.github/workflows/dry-run.yml
@@ -135,7 +135,7 @@ jobs:
           # Perform build and execution separately to avoid any potential output from
           # cargo leaking into the output file.
           cargo build --release
-          ./target/release/sync-team print-plan --services github --team-json team-api 2> output.txt
+          ./target/release/sync-team print-plan --services github --team-json team-api 2>&1 | tee -a output.txt
 
       - name: Prepare comment
         run: |


marcoieni (approved):

apiraino:
@Kobzol pull requests now get an automated comment with the CI debug output ([example](https://github.com/rust-lang/team/pull/1700#issuecomment-2735091212)). Subscribers of this repository now receive a notification (and a rather long email, 85 KiB) with the CI debug output which is not super interesting. Unsure if this behaviour started with this patch or in #1684.

Can the debug be available only in the CI logs instead of posting a comment on the pull request?

Thanks

Kobzol:
I do acknowledge that it is annoying, sorry.

However, we would really like to have the comment on the PR rather than in CI, to make it easier for reviewers to see what will be the applied diff. We had it only in CI earlier and no one was looking at it :upside_down_face:

The fact that the log is so long is a bug, it should be mostly empty for most PRs. However, we're currently starting a large restructuring of the team and sync-team repos (https://github.com/rust-lang/team/pull/1697) and I'm not sure if we want to fix the bug before or after that :thinking:

@marcoieni Maybe we could just temporarily disable the posting of the comment before we finish the team/sync-team merge?

marcoieni:
yes, sure 👍

Kobzol:
https://github.com/rust-lang/team/pull/1701

apiraino:
Thanks folks.

PR #1697, opened 2025-03-15
Import sync team into team

Kobzol:
This PR shows how it would look like if we imported sync-team into team. It doesn't do anything else than import the `sync-team` history, moved into the `sync-team` directory using [josh-filter](https://josh.readthedocs.io/en/latest/cli.html).


marcoieni (approved):
Looks good. I was waiting to reach consensus on zulip before merging this. So that we have a clear plan on how to migrate from that repo to this one.

Kobzol:
Updated to the latest version of sync-team. If we modify sync-team in the meantime, it should be fine to still add the new commits in a "git-history-compatible" using josh, until we finish the migration.

PR #1700, opened 2025-03-19
Lowercase `DianQK` to `dianqk`

dianqk:
This is the pinyin abbreviation of the Chinese pen name (靛青 K) when I was a child. Typing `DianQK` might be a bit troublesome, so I lowercase it to `dianqk`.

For this, I added an English name: Quentin Dian, I feel this makes the problem worse, my Chinese name is Xutao Song. I think you can call me dianqk or any combination of the above.

### DIFF
--- a/people/dianqk.toml
+++ b/people/dianqk.toml
@@ -1,5 +1,5 @@
-name = 'DianQK'
-github = 'DianQK'
+name = 'dianqk'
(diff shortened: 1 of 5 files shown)


marcoieni (approved):
approving to unblock CI errors 👍

PR #1706, opened 2025-03-20
give the `mods` team write access to `team` repo

marcoieni:
Hey @rust-lang/mods, I was wondering if `write` permission is sufficient for this repository, or whether you need `maintain` for certain settings.

I'm asking because this repo will soon run [sync-team](https://github.com/rust-lang/sync-team) in CI, so we are making sure everyone has the least permissions as possible to mitigate the risk of account compromised and similar issues 👍

### DIFF
--- a/repos/rust-lang/team.toml
+++ b/repos/rust-lang/team.toml
@@ -4,7 +4,7 @@ description = "Rust teams structure"
 bots = []
 
 [access.teams]
-mods = "maintain"
+mods = "write"
 team-repo-admins = "write"
 infra = "triage"
 


marcoieni:
converting to draft to wait for the mods team response before merging of course 👍

oli-obk:
As long as we can lock issues and PRs that's fine. Ideally we can ban ppl directly from comments, which I can do from the rust repo but not from the cargo repo. I need to copy the name and manually block them from the org

marcoieni:
in the rust repo you have write permissions, so "write" should be enough here 👍 
image

while in the cargo repo, mods don't have any permissions. You might want to add your team [here](https://github.com/rust-lang/team/blob/f517b662bbb2ecbcc8ab9f34602587360cdd7856/repos/rust-lang/cargo.toml#L8). Maybe "triage" is enough and you don't even need "write"

Mark-Simulacrum (approved):

PR #1717, opened 2025-03-25
Upload GitHub Pages using the standard action

Kobzol:
Rather than using the bespoke script, which shouldn't be needed anymore. If this works, the `GITHUB_DEPLOY_KEY` secret can be removed from this repository.

I wanted to land this separately from https://github.com/rust-lang/team/pull/1715 to make sure that it works and to avoid too many changes in that PR. If this deployment works here, it should also work in the merge queue, without us having to reconfigure any terraform settings.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,7 @@ jobs:
     if: github.repository == 'rust-lang/team'
     permissions:
       id-token: write
+      pages: write
     steps:
 
       - uses: actions/checkout@main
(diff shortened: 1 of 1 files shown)


marcoieni (approved):

Kobzol:
https://github.com/rust-lang/team/actions/runs/14064397710/job/39383066087 Huh, that's weird. I modelled this after https://github.com/Kobzol/phd/blob/main/.github/workflows/ci.yml, where it doesn't require any environment..

PR #1729, opened 2025-03-28
don't run dry-run periodically

marcoieni:
now that the sync runs every 4 hours, there's no reason to also perform the dry run at the same time.

but maybe I'm missing something.

also, for some reason, the `CI` event is not running in actions: https://github.com/rust-lang/sync-team/actions?query=event%3Aschedule
I fill try to fix this separately


marcoieni:
wrong file

PR #1730, opened 2025-03-28
fix: run sync-team on schedule

marcoieni:
as you can see [here](https://github.com/rust-lang/team/actions/runs/14124599167), the deploy step is skipped

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -77,7 +77,7 @@ jobs:
     permissions:
       id-token: write
       pages: write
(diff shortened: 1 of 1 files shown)


Kobzol (commented):
Ah, I was thinking about this, but forgot to change it in the YAML :man_facepalming: I would just do `if: github.event_name != 'pull_request'` :sweat_smile:

Kobzol (approved):

jdno (approved):
:shipit:

PR #1463, opened 2024-05-25
Add rust repository under automation

Kobzol:
Repository: https://github.com/rust-lang/rust

Extracted from GH:
```toml
org = "rust-lang"
name = "rust"
description = "Empowering everyone to build reliable and efficient software."
bots = []

[access.teams]
cargo = "write"
project-stable-mir = "write"
compiler = "write"
release = "write"
rustdoc = "write"
rustfmt = "pull"
security = "pull"
rfcbot = "write"
opsem = "write"
clippy = "write"
style = "write"
libs-api = "write"
libs-contributors = "write"
wg-prioritization = "triage"
wg-triage = "write"
highfive = "write"
project-exploit-mitigations = "write"
compiler-contributors = "write"
mods = "write"
types = "write"
wg-rustc-dev-guide = "pull"
infra = "write"
lang = "write"
libs = "write"
bootstrap = "write"
core = "admin"
lang-ops = "write"
bors = "write"

[access.individuals]
BurntSushi = "write"
rust-lang-glacier-bot = "write"
alexcrichton = "write"
flip1995 = "write"
xFrednet = "write"
Enselic = "write"
rylev = "admin"
weihanglo = "write"
JohnTitor = "write"
nnethercote = "write"
Mark-Simulacrum = "admin"
cramertj = "write"
hameerabbasi = "triage"
MarcoIeni = "admin"
davidtwco = "write"
durin42 = "write"
shepmaster = "write"
tgross35 = "write"
scottmcm = "write"
lcnr = "write"
pitaj = "write"
SimonSapin = "write"
rfcbot = "write"
camelid = "write"
WaffleLapkin = "write"
Manishearth = "write"
craterbot = "write"
albertlarsan68 = "write"
fmease = "write"
fee1-dead = "write"
pnkfelix = "write"
clubby789 = "write"
traviscross = "write"
oskgo = "write"
Centri3 = "write"
workingjubilee = "write"
onur-ozkan = "write"
est31 = "write"
bstrie = "write"
notriddle = "write"
estebank = "write"
llogiq = "write"
SparrowLii = "write"
KittyBorgX = "write"
b-naber = "write"
nikomatsakis = "write"
epage = "write"
tmandry = "write"
petrochenkov = "write"
eddyb = "write"
joshtriplett = "write"
aliemjay = "write"
digama0 = "write"
jackh726 = "write"
alex-semenyuk = "write"
rustbot = "write"
rust-lang-owner = "admin"
kennytm = "write"
RalfJung = "write"
[...]

PR #1746, opened 2025-03-31
Add bots needed for rust-lang/rust

Kobzol:
The [dry-run](https://github.com/rust-lang/team/actions/runs/14170134329/job/39691899954) on https://github.com/rust-lang/team/pull/1463 showed that we will need to include these bots.

(It's so great to have the option to modify both team and sync-team in the same PR :heart:)

### DIFF
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -205,6 +205,9 @@ pub enum Bot {
     Rustbot,
     RustTimer,
     Rfcbot,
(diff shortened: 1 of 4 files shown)


Mark-Simulacrum (approved):
Let's aim for parity of permissions first, but I'm actually not sure crater or log analyzer need write permissions. Glacier I'm not sure is still running (I don't think it is on our infra at least?)

PR #1747, opened 2025-03-31
Remove unused bors repos from permission allowlist

Kobzol:
These repos no longer use bors (neither the old one nor the new one). Homu no longer loads any other repo than `rust`, so removing the (empty) JSON bors permission files from the rendered output should hopefully be fine.


jdno (approved):
> should hopefully be fine

With homu, that seems like the most confidence we'll ever have. 😆

PR #1749, opened 2025-03-31
Make it possible to run sync-team against local data

Kobzol:
Before, `sync-team` performed sync against the live service. Now it runs against the local `team` checkout by default. This PR also makes it easier to run `sync-team` on the local checkout data (as that is probably a common use-case, if you use it locally), without having to prebuild the JSON data directory first.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -25,6 +25,7 @@ use clap::Parser;
 use log::{error, info, warn};
 use std::collections::{BTreeMap, HashMap};
 use std::path::PathBuf;
+use std::str::FromStr;
 use sync_team::run_sync_team;
 use sync_team::team_api::TeamApi;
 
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -114,19 +118,62 @@ struct SyncOpts {
...
-    /// Path to a directory with prebuilt JSON data from the `team` repository.
-    #[clap(long, global(true))]
-    team_json: Option<PathBuf>,
-
+    /// Source of the data against which is the sync performed.
+    /// Possible values:
+    /// - in-tree => use the current team checkout
+    /// - production => use the live REST API team endpoint
+    /// - <path> => use prebuilt directory at <path>
+    #[arg(
```
Kobzol, line 126: I wanted to implement it as two commands at first, but it was quite tricky in clap, especially since `DataSource::Prebuilt` actually takes a value (path).

marcoieni:
looks good. can you resolve the conflict?

marcoieni (approved):

PR #1751, opened 2025-04-01
Do not require PRs for branches managed by homu

Kobzol:
To make the rust-lang/rust branch protections closer to their original state. Should help if bors merges fail after the migration of rust-lang/rust under `team`.

### DIFF
--- a/sync-team/src/github/mod.rs
+++ b/sync-team/src/github/mod.rs
@@ -506,17 +506,21 @@ pub fn construct_branch_protection(
     branch_protection: &rust_team_data::v1::BranchProtection,
 ) -> api::BranchProtection {
     let uses_homu = branch_protection.merge_bots.contains(&MergeBot::Homu);
-    let required_approving_review_count: u8 = if uses_homu {
-        0
+    // When homu manages a branch, we should not require a PR nor approvals
+    // for that branch, because homu pushes to these branches directly.
+    let branch_protection_mode = if uses_homu {
+        BranchProtectionMode::PrNotRequired
     } else {
-        match branch_protection.mode {
-            BranchProtectionMode::PrRequired {
-                required_approvals, ..
-            } => required_approvals
-                .try_into()
-                .expect("Too large required approval count"),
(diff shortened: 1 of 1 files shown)


sync-team/src/github/mod.rs
```diff
@@ -506,17 +506,21 @@ pub fn construct_branch_protection(
     branch_protection: &rust_team_data::v1::BranchProtection,
 ) -> api::BranchProtection {
     let uses_homu = branch_protection.merge_bots.contains(&MergeBot::Homu);
-    let required_approving_review_count: u8 = if uses_homu {
-        0
+    // When homu manages a branch, we should not require a PR nor approvals
+    // for that branch, because homu pushes to these branches directly.
```
marcoieni, line 510: but homu is in the allowlist, so why is this change required?
image

Kobzol: Conceptually, when a bors manages a branch, PRs (and PR reviews) are not required, because bors pushes to these branches directly. So I think that the branch protections should reflect that (even though GH also allows the bot to push even if PRs are required, since we didn't configure any required CI checks).

marcoieni: I see. Sounds good to me. before merging this, is there a way to visualize the branch protection difference in the dry run?

Kobzol: Hmm, we currently don't log the branch protection diffs fully. Let me fix that.

Kobzol: https://github.com/rust-lang/team/pull/1756

marcoieni:
I ran the dry run locally, given that the token we have in CI doesn't have permissions to diff against branch protections:
```
    📝 Editing repo 'rust-lang/rust':
      Branch Protections:
          master
            Requires PR: true => false
          stable
            Requires PR: true => false
          beta
            Requires PR: true => false
          try
            Requires PR: true => false
          auto
            Requires PR: true => false
          *
            Requires PR: true => false
```

PR #1752, opened 2025-04-01
Add debug logging of team and repo diffs during actual sync

Kobzol:
To help observe the progress better when real synces are being performed. Currently, we are staring at an empty screen for ~3 minutes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,7 +97,7 @@ jobs:
           ZULIP_API_TOKEN: ${{ secrets.ZULIP_API_TOKEN }}
           ZULIP_USERNAME: ${{ secrets.ZULIP_USERNAME }}
         run: |
-          cargo run sync apply --team-json build
+          RUST_LOG="sync_team=debug" cargo run sync apply --team-json build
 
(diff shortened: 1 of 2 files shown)


Kobzol:
Note that I didn't add the logging to the dry run, because there we want to have an almost empty output if there is no diff, rather than a lot of log data.

marcoieni (dismissed):

.github/workflows/main.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           ZULIP_API_TOKEN: ${{ secrets.ZULIP_API_TOKEN }}
           ZULIP_USERNAME: ${{ secrets.ZULIP_USERNAME }}
         run: |
-          cargo run sync apply --team-json build
+          RUST_LOG=sync_team=debug cargo run sync apply --team-json build
```
marcoieni, line 100:
```suggestion
          RUST_LOG="sync_team=debug" cargo run sync apply --team-json build
```
this seems a bit cleaner

Kobzol: Changed :+1:

marcoieni (approved):

PR #1756, opened 2025-04-01
Use destructuring in diffs to make it harder to forget to include diff fields

Kobzol:
Some branch protection diff fields weren't logged previously, this should prevent that from happening in the future. This PR also denies warnings on CI.

### DIFF
--- a/sync-team/src/github/mod.rs
+++ b/sync-team/src/github/mod.rs
@@ -950,9 +950,21 @@ fn log_branch_protection(
     new: Option<&api::BranchProtection>,
     mut result: impl Write,
 ) -> std::fmt::Result {
+    let api::BranchProtection {
+        // Pattern identifies the branch protection, so it has to be same between `current`
+        // and `new`.
+        pattern: _,
+        is_admin_enforced,
+        dismisses_stale_reviews,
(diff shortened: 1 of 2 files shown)


sync-team/src/github/mod.rs
```diff
@@ -676,25 +676,33 @@ impl CreateRepoDiff {
...
 
         writeln!(f, "➕ Creating repo:")?;
-        writeln!(f, "  Org: {}", self.org)?;
-        writeln!(f, "  Name: {}", self.name)?;
-        writeln!(f, "  Description: {:?}", description)?;
-        writeln!(f, "  Homepage: {:?}", homepage)?;
-        writeln!(f, "  Auto-merge: {}", auto_merge_enabled)?;
+        writeln!(f, "  Org: {org}")?;
+        writeln!(f, "  Name: {name}")?;
+        writeln!(f, "  Description: {description}")?;
```
Kobzol, line 697: The description was originally `Option<String>`, which is why it was logged with `:?`. But recently we switched it to just `String`, so it doesn't make sense to print it as `Debug` anymore.

marcoieni (approved):
great, thanks!

PR #1758, opened 2025-04-01
allow team leaders to edit their teams

marcoieni:
- [ ] If this is approved, merge the `leads = "write"` before this PR, so that team leads have write access

We want to give team leaders permission to approve PRs in this repository, so that they can unblock their teams.

The question is: how much permission can we give them? If you look at the CODEOWNERS file, you can see that anyone
with write access can approve PRs in the `/people` `/repos` and `/teams` directories.
But what if an attacker compromises a team-lead's account? They could then approve their own PRs
to remove branch protection rules, kick out team members, or archive all rust-lang repositories.

To prevent this, we want to limit the permissions of team-leads to only the directories they own,
i.e. their `/teams` and the `/repos` owned by their teams.

With this PR, I start by giving team-leads write access to their own `/teams` directory.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,10 +11,93 @@
 # on every change to the data files.
 people/**/*.toml
 repos/**/*.toml
+# Useful for teams without leaders.
 teams/**/*.toml
 
 # Do not require admin approvals for Markdown file modifications.
 *.md
+
+# Team leads can approve changes to their own team files.
+/teams/book.toml @team-repo-admins @mods @carols10cents @chriskrycho
+/teams/bootstrap.toml @team-repo-admins @mods @Mark-Simulacrum @onur-ozkan
+/teams/cargo.toml @team-repo-admins @mods @ehuss
+/teams/clippy.toml @team-repo-admins @mods @Manishearth @flip1995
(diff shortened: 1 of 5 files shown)


.github/CODEOWNERS
```diff
@@ -9,24 +9,142 @@
 # Data files can be approved by users with write access.
 # We don't list these users explicitly to avoid notifying all of them
 # on every change to the data files.
-people/**/*.toml
-repos/**/*.toml
-teams/**/*.toml
+people/**/*.toml @rust-lang/team-repo-admins @rust-lang/mods @Mark-Simulacrum @pietroalbini @jdno @marcoieni
+repos/**/*.toml  @rust-lang/team-repo-admins @rust-lang/mods @Mark-Simulacrum @pietroalbini @jdno @marcoieni
+# Useful for teams without leaders.
+teams/**/*.toml  @rust-lang/team-repo-admins @rust-lang/mods @Mark-Simulacrum @pietroalbini @jdno @marcoieni
```
marcoieni, line 15: unfortunately with this approach we need to add specific owners in these generic directories, otherwise the team-leads will have the same permissions of team-repo-admins and mods, i.e. they could approve PRs on all these toml files

PR #1763, opened 2025-04-04
Archive inactive compiler WGs

apiraino:
This patch does some cleanup.

* Archives:
  * wg-debugging
  * wg-polymorphization
  * wg-self-profile
  * wg-incr-comp
* Renames active groups where work is happening, removing the "Working Group" wording (which T-compiler is sunsetting).

This is part of the work stemming from https://github.com/rust-lang/compiler-team/issues/848

Thanks


Kobzol:
There is [wg-debugging](https://github.com/rust-lang/wg-debugging) repo, that should also be archived (and its `access.teams` should be thus empty).

Btw if the "working group" suffix is dropped, it might look a bit weird on the [website](https://www.rust-lang.org/governance/teams/compiler#team-wg-binary-size) (it will show just "Binary size" as the header).

marcoieni:
I tried to improve the error message in https://github.com/rust-lang/team/pull/1764 👍

apiraino:
I get a new error locally, unrelated to this PR:
```
[ERROR rust_team::validate] validation error: GitHub user `m-rph` changed username to `PartiallyUntyped`
```

Should I fix this first in a separate PR? What do we do in these cases?

Kobzol:
Yeah, we should do that in a separate PR. It's a bit unfortunate, but when a user in the team DB renames themselves, it currently blocks `team`'s CI :sweat_smile:

apiraino:
> Yeah, we should do that in a separate PR. It's a bit unfortunate, but when a user in the team DB renames themselves, it currently blocks `team`'s CI 😅

Must be my lucky day. Exactly while I was changing stuff here :D

marcoieni:
In case this is ready, probably a team-repo-admin should review this (not infra admins) since it changes teams 👍

apiraino:
> Btw if the "working group" suffix is dropped, it might look a bit weird on the [website](https://www.rust-lang.org/governance/teams/compiler#team-wg-binary-size) (it will show just "Binary size" as the header).

I thought it was ok but since you mention maybe it was not *that* ok, so I have added a "working area" suffix for some of them, which is not great either but it feels better (and I am uninspired to find a better title anyway).

jackh726:
r? @davidtwco @wesleywiser 

(particulary on "working area" naming, I guess)

davidtwco (approved):

teams/wg-compiler-performance.toml
```diff
@@ -23,9 +23,10 @@ alumni = [
 orgs = ["rust-lang"]
 
 [website]
-name = "Compiler performance working group"
+name = "Compiler performance working area"
```
spastorino, line 26: Here we change working group to working area.

teams/wg-const-eval.toml
```diff
@@ -18,9 +18,10 @@ alumni = [
 name = "Compile-time Function Evaluation Working Group"
```
spastorino, line 18: Here and in other cases we just keep Working Group.

teams/wg-compiler-performance.toml
Kobzol: It was like that originally, I suggested to add some suffix (https://github.com/rust-lang/team/pull/1763#issuecomment-2778353451).

spastorino: In any case, what I meant with my 3 comments is that it seems it lacks some consistency. In some place is working area, in some others is working group and in others nothing.
I was wondering if that have a reason that I'm not understanding.

apiraino:
@Kobzol I think we're done here

(to be clear: I've intentionally only touched files marked as compiler subteams)

Kobzol:
Fine by me :) (I can't approve it though)

apiraino:
cc @technetos for a last look and push the merge button :^)

jackh726 (changes requested):
Please remove all leads from archived teams too, please.

jackh726:
Archived teams still have the leads?

jackh726 (approved):

PR #1764, opened 2025-04-04
improve repo validation error message

marcoieni:
A user complained in https://github.com/rust-lang/team/pull/1763#issuecomment-2778295890 that this error was confusing. 

tested against that PR:

```
$ GITHUB_TOKEN=$(gh auth token) rust-team check                                                                  
[WARN  rust_team::validate] couldn't perform checks relying on the Zulip API, some errors will not be detected
[WARN  rust_team::validate] cause: missing ZULIP_USER and/or ZULIP_TOKEN environment variables
[ERROR rust_team::validate] validation error: Access for the repo rust-lang/wg-debugging is invalid: the team 'rust-lang/wg-debugging' is archived. Please remove it from the repo.
```

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -868,20 +868,25 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
         }
         for team_name in repo.access.teams.keys() {
             if !github_teams.contains(&(repo.org.clone(), team_name.clone())) {
+                let error_reason = if data.is_team_archived(team_name, &repo.org) {
+                    "is archived. Please remove the team access from the `[access.teams]` section of the repo"
+                } else {
+                    "is not present in the `/teams/` folder"
+                };
+
                 bail!(
-                        "access for {}/{} is invalid: '{}' is not configured as a GitHub team for the '{}' org",
-                        repo.org,
-                        repo.name,
-                        team_name,
-                        repo.org
-                    )
+                    "Access for the repo {}/{} is invalid: the team '{}/{team_name}' {error_reason}.",
+                    repo.org,
+                    repo.name,
+                    repo.org,
+                );
             }
         }
 
         for name in repo.access.individuals.keys() {
             if data.person(name).is_none() {
                 bail!(
-                    "access for {}/{} is invalid: '{}' is not the name of a person in the team repo",
+                    "Access for the repo {}/{} is invalid: '{}' is not the name of a person in the team repo",
                     repo.org,
                     repo.name,
                     name
--- a/src/data.rs
+++ b/src/data.rs
@@ -210,6 +210,13 @@ impl Data {
         }
         result
     }
+
+    pub(crate) fn is_team_archived(&self, team_name: &str, org: &str) -> bool {
+        self.archived_teams()
+            .filter_map(|team| team.github_teams(self).ok())
+            .flatten()
+            .any(|github_team| github_team.name == team_name && github_team.org == org)
+    }
 }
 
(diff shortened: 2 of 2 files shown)


src/validate.rs
```diff
@@ -868,20 +868,26 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
         }
         for team_name in repo.access.teams.keys() {
             if !github_teams.contains(&(repo.org.clone(), team_name.clone())) {
+                let is_team_archived = data.archived_teams().any(|t| t.name() == team_name);
```
Kobzol, line 871: We should start taking organizations into account :) Ideally by creating some form of `Team/RepoId`, which contains both the org and the name. To avoid naming collisions across different organizations.

marcoieni: right. Can this be a separate issue?

Kobzol: Sure, but I would rather not add new code that doesn't consider organizations. There's a possibly bigger problem here though, as you should check the github team name (and organization), not the `team` team name (yeah, it's confusing.. xD).

marcoieni: you mean I should check the [github](https://github.com/rust-lang/team/blob/92c25e89785dbedaaa5c8ed4f164cd53b3ecbf37/rust_team_data/src/v1.rs#L26) field? https://github.com/rust-lang/team/blob/92c25e89785dbedaaa5c8ed4f164cd53b3ecbf37/rust_team_data/src/v1.rs#L49

Kobzol: Yeah, because `access.teams` refers to github team names, not `team` team names xD

apiraino:
> Access for the repo rust-lang/wg-debugging is invalid: the team 'rust-lang/wg-debugging' is archived. Please remove it from the repo.

The new version of the validation error is still not very clear to me (that I don't know a lot about the `team` repository). What should I remove? Which repo? I've deleted [this line](https://github.com/rust-lang/team/pull/1763/files#diff-d27f973d2f9f506a2965ad33e384ce338f67ff2d1e4aaad0d861a74d9171a0cbL7) and the validation now pass but I can't correlate my change to the validation error message.

How about this:
> Access for the repo rust-lang/wg-debugging is invalid: the team 'rust-lang/wg-debugging' is archived. Please remove the team from the `[access.teams]` section.

marcoieni:
rephrased, thanks a lot! This kind of feedback is very useful 🙏

src/validate.rs
```diff
@@ -868,20 +868,25 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
         }
         for team_name in repo.access.teams.keys() {
             if !github_teams.contains(&(repo.org.clone(), team_name.clone())) {
+                let error_reason = if data.is_team_archived(team_name, &repo.org) {
+                    "is archived. Please remove the team access from the `[access.teams]` section of the repo"
```
jdno, line 872: I'm wondering if something like this wouldn't be even easier to understand. Might require a change to the data model, though, to deserialize a repository without the `access.teams` field set. 🤔 

```suggestion
                    "is archived. Please remove the `[access.teams]` section in the repo's toml"
```

marcoieni: this change sounds more like https://github.com/rust-lang/team/pull/1754
Why should we suggest removing the entire `[access.teams]` section of the repo here?
That particular team is archived, but not the repo.

jdno: Oh yeah, my bad. When looking how `[access.teams]` was used in the repository TOMLs, I forgot that this was about an archived team. 🙈

jdno (approved):
LGTM :shipit:

PR #1767, opened 2025-04-06
Add branch protection for ci-mirrors

pietroalbini:


repos/rust-lang/ci-mirrors.toml
```diff
@@ -6,3 +6,8 @@ bots = []
 
 [access.teams]
 infra = "write"
+
+[[branch-protections]]
+pattern = "main"
+ci-checks = ["Build finished"]
+required-approvals = 1
```
Kobzol, line 13: `1` is the default, so no need to spell it out, unless you want to be explicit for some reason :)

Kobzol (approved):

repos/rust-lang/ci-mirrors.toml
pietroalbini: Didn't know that! Added `dismiss-stale-review = true` in the meantime tho.

Mark-Simulacrum (approved):

PR #1769, opened 2025-04-08
[t-spec] Move m-ou-se to alumni status

JoelMarcey:


teams/spec-contributors.toml
```diff
@@ -3,10 +3,10 @@ subteam-of = "spec"
 
 [people]
 leads = []
-members = [
+members = []
+alumni = [
     "chorman0773",
 ]
-alumni = []
```
WaffleLapkin, line 10: you should probably change the title, since this is not m-ou-se

JoelMarcey: I just appended this to the original PR that was `m-ou-se`, but ok, I will.

jackh726 (approved):

PR #1781, opened 2025-04-17
give mods permissions in the cargo repo

oli-obk:
Otherwise we can't unhide comments (hiding works amusingly), lock issues, or ban people directly from comments they leave on that repo

cc @rust-lang/mods


ehuss (approved):
Seems like we should automate this (like with a default or something).

steffahn:
This is superseded by
* #1848

PR #1784, opened 2025-04-22
Set more WG teams as marker team

apiraino:
Follow up to #1763 

With this patch I am moving some WG teams from `working-group` to `marker-team`. The reasoning is (as per this [comment](https://github.com/rust-lang/team/pull/1763#issuecomment-2786645629) and this other [comment on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Difference.20between.20marker.20and.20archived.20teams.3F/near/512676551)) that these are more "ping lists" rather than full-fledged teams.

Please review these changes for inaccuracies, I am building a picture of the Rust project structure as I go :-)

If these changes are conceptually correct, then I'll figure out something to fix the validation error.

Thanks!

cc: @davidtwco 
r? @ehuss

### DIFF
--- a/repos/archive/rust-lang-deprecated/error-chain.toml
+++ b/repos/archive/rust-lang-deprecated/error-chain.toml
@@ -1,0 +1,6 @@
+org = "rust-lang-deprecated"
+name = "error-chain"
+description = "Error boilerplate for Rust"
+bots = []
+
+[access.teams]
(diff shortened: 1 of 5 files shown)


apiraino:
> I don't think I would agree these would be good to set as marker teams

All of them? Then I don't have clear the border that qualifies a team to be *marker* (as discussed on Zulip). You mean to just drop this patch or there's something to salvage? The WG-prio team has *never* been a team so at least that one should be ok :-)

ehuss:
I think for now my guidance would be: If the *only* purpose is to support `@rustbot ping foo`, then that would make sense to use a marker team. From what I can tell, all of these seem to be something more than that.

PR #1786, opened 2025-04-23
Move directors to launching pad, give them blog permissions

ehuss:
I think we can trust directors to post on the blog, so lets give them permissions to do that so they don't need to bug anyone else to do it.

To accommodate the addition of permissions, this promotes it from a marker team to a team in the launching pad (which we are planning to rebrand as a home for cross-cutting teams). There's no expectation that this would change anything in practice, and the membership is still chosen by the council.

cc @rust-lang/leadership-council @carols10cents @JakobDegen @rylev @scottmcm @spastorino

### DIFF
--- a/teams/foundation-board-project-directors.toml
+++ b/teams/foundation-board-project-directors.toml
@@ -1,5 +1,5 @@
 name = "foundation-board-project-directors"
-kind = "marker-team"
+subteam-of = "launching-pad"
 
 [people]
 leads = []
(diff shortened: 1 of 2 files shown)


cuviper (approved):

traviscross (approved):

traviscross:
Agreed this is the right handling.

jamesmunns (approved):
Seems reasonable to me!

oli-obk (approved):

traviscross:
Five of us on the council have signed off on this, and I don't think this needs a formal council decision as it comports well enough with our existing ones.  I'm raising my hand here and saying this is OK to merge.

jackh726 (approved):

PR #1790, opened 2025-04-25
Enable rustbot in `rustc_codegen_cranelift`&`rustc_codegen_gcc`

Urgau:
This is going to be useful for enabling [`[no-mentions]`](https://forge.rust-lang.org/triagebot/no-mentions.html) and [`[issue-links]`](https://forge.rust-lang.org/triagebot/issue-links.html) in both repository.

Which we should anyway to before enabling them in rust-lang/rust as otherwise subtree sync risk being flagged with warnings.

cc @bjorn3 @GuillaumeGomez @antoyo

### DIFF
--- a/repos/rust-lang/rustc_codegen_cranelift.toml
+++ b/repos/rust-lang/rustc_codegen_cranelift.toml
@@ -1,7 +1,7 @@
 org = 'rust-lang'
 name = 'rustc_codegen_cranelift'
 description = 'Cranelift based backend for rustc'
-bots = []
+bots = ["rustbot"]
 
 [access.teams]
 compiler = 'maintain'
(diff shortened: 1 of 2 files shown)


Kobzol:
I think that adding triagebot.toml is the main thing, this just adds write access to the triagebot account, but that shouldn't be needed for these checks (it's a good idea regardless though, to track that we use rustbot there).

antoyo (approved):
LGTM

tgross35:
> I think that adding triagebot.toml is the main thing, this just adds write access to the triagebot account, but that shouldn't be needed for these checks (it's a good idea regardless though, to track that we use rustbot there).

Write access meaning to post comments, or to write to the repo? I can't think of any time it does the latter.

Kobzol:
I think that anyone can post comments, write access means merging PRs and pushing commits.

Urgau:
Doesn't it also allow editing comments, hiding comments, ...? Both are required for these two triagebot features.

Kobzol:
Good point! I guess it could hide its own comments, but definitely couldn't edit PR descriptions without the write permission.

bjorn3 (approved):

jackh726 (approved):

PR #1794, opened 2025-04-29
Add `rfcbot` to the bots used in the reference repo

Urgau:
cf. https://github.com/rust-lang/reference/pull/1792#issuecomment-2824692734 where the bot wasn't able to add the label because it doesn't have the rights to do so.


ehuss (approved):
Thanks!

jackh726 (approved):

PR #1809, opened 2025-05-02
Track the FLS repo

pietroalbini:
Transferred the FLS repo from `ferrocene/specification` to `rust-lang/fls`. This will have the same permissions and access levels as the reference. cc @rust-lang/spec


JoelMarcey (approved):
This is awesome! Glad this has finally happened!

Mark-Simulacrum (approved):

PR #1814, opened 2025-05-08
Add Google Summer of Code 2025 contributors to the `gsoc-contributors` team

Kobzol:
I will fill the PR properly this evening, once the accepted projects are published by Google.


jackh726 (approved):
Not needed for this, but I wonder if we should make use of roles? (I think this only makes sense if alumni can have roles!)

Kobzol:
@jackh726 Could we merge this already to provide access to the new contributors ([#gsoc > GSoC 2025 participants @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/421156-gsoc/topic/GSoC.202025.20participants/near/517079331))? I'll take a look at the roles and the website later.

PR #1815, opened 2025-05-13
Update Rustlings team summary

manyinsects:
The (I assume temporary) description talks about a site, when in fact Rustlings is mostly something you download onto your computer (although we do have a website).


apiraino (approved):
thanks!

jackh726 (approved):

PR #1818, opened 2025-05-16
Update my account information

0xPoe:
People always ask me about my name, so I figured it was time for a new one. 😆

### DIFF
--- a/people/0xPoe.toml
+++ b/people/0xPoe.toml
@@ -1,6 +1,6 @@
-name = 'Rustin'
-github = 'Rustin170506'
+name = 'Poe'
+github = '0xPoe'
 github-id = 29879298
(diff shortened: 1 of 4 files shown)


apiraino (approved):
lgtm

people/0xPoe.toml
```diff
@@ -1,6 +1,6 @@
-name = 'Rustin'
-github = 'Rustin170506'
+name = 'Poe'
```
weihanglo, line 1: So we should stop calling you Rustin, is that correct?

0xPoe: You can call me Dongpo if you prefer. Ed told me that my Chinese name is not that hard. 😂

jackh726 (approved):

PR #1829, opened 2025-05-21
Unarchive gha-self-hosted

pietroalbini:
As discussed at the All Hands, let's unarchive the gha-self-hosted repo, so that we can bring it to a working state again and do some benchmarks on it. We're not sure yet whether we actually want to start using it again though.

### DIFF
--- a/repos/rust-lang/gha-self-hosted.toml
+++ b/repos/rust-lang/gha-self-hosted.toml
@@ -4,3 +4,7 @@ description = "GitHub Actions self-hosted runners infrastructure"
 bots = []
 
 [access.teams]
+infra = "write"
+
+[[branch-protections]]
+pattern = "main"


marcoieni (approved):

repos/rust-lang/gha-self-hosted.toml
```diff
@@ -4,3 +4,7 @@ description = "GitHub Actions self-hosted runners infrastructure"
 bots = []
 
 [access.teams]
+infra = "write"
```
Kobzol, line 7: Not needed right now for experiments ofc, but if we do end up using self-hosted runners, we should eventually probably avoid giving the whole of `t-infra` write access. To avoid the previous situation where any t-infra member could essentially arbitrarily affect CI through `rust-lang-ci/rust` :laughing:

PR #1838, opened 2025-05-23
Sync wg-embedded-arm.toml

thejpster:
Synchronised with https://github.com/rust-embedded/wg/blob/master/README.md.

robamu joined the team and ithinuel came out of hibernation.

PR #1840, opened 2025-05-24
Bumps schannel to a version that isn't buggy.

thejpster:
See https://github.com/steffengy/schannel-rs/issues/106.

Resolves the issue I saw in https://github.com/rust-lang/team/pull/1838.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1671,12 +1671,11 @@ dependencies = [
 
 [[package]]
 name = "schannel"
-version = "0.1.20"
+version = "0.1.27"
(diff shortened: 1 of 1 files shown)


apiraino:
Just out of curiosity:
```
[dependencies.schannel]
version = "0.1.27"
```

Would that solve the problem the same? I often destroy/desync my Cargo.lock[^1]

(Though I didn't experience that issue on Linux, so I can't say for sure if it fixes)

[^1]: One possible reply could be "then don't it :smile: "

PR #1850, opened 2025-05-28
add jyn514 back to rustc-dev-guide

jyn514:
i want to do a bunch of work around stability, the unstable-book, and forge. we don't really have a team for any of that, but rustc-dev-guide seems close enough.

can someone confirm that this will *not* give me r+ for rust-lang/rust? i don't want that access.


Kobzol:
There's no `bors.rust.review = true` in this file, so you shouldn't get `r+` on r-l/r.

jieyouxu (approved):
Welcome back!

PR #1852, opened 2025-05-29
Add back two removed homepages

Kobzol:
I noticed in the dry-run results of https://github.com/rust-lang/team/pull/1797 that two homepages were removed, I missed that. I think that my backporting scripts skipped some homepages. It didn't seem to happen in the other recently merged PRs, only here.


Kobzol:
Uhh, crap, the repos are archived, so `team` won't actually merge the homepages back. The homepages should thus be changed manually by an infra-admin (sorry!). This PR should still be merged though.

PR #1805, opened 2025-04-30
Backport rust-analyzer repos (batch 1)

Kobzol:
Backporting the following repos:

- [expect-test](https://github.com/rust-analyzer/expect-test)
    <details markdown="1">
    <summary>Original content</summary>

    ```toml
    org = "rust-analyzer"
    name = "expect-test"
    description = "Minimalistic snapshot testing for Rust."
    bots = []

    [access.teams]
    triage = "triage"

    [access.individuals]
    MissHolmes = "admin"
    pnkfelix = "admin"
    skade = "admin"
    Xanewok = "admin"
    kjeremy = "triage"
    edwin0cheng = "triage"
    JoshMcguigan = "triage"
    pietroalbini = "admin"
    wesleywiser = "admin"
    Veykril = "admin"
    kiljacken = "triage"
    lnicola = "admin"
    bjorn3 = "triage"
    flodiebold = "admin"
    matklad = "admin"
    Veetaha = "triage"
    lf- = "triage"
    quad = "write"
    dustypomerleau = "triage"
    rust-lang-owner = "admin"
    SomeoneToIgnore = "triage"

    [[branch-protections]]
    pattern = "master"
    ci-checks = ["test"]
    required-approvals = 0
    pr-required = false
    ```

    </details>

- [rowan](https://github.com/rust-analyzer/rowan)
    <details markdown="1">
    <summary>Original content</summary>

    ```toml
    org = "rust-analyzer"
    name = "rowan"
    description = ""
    bots = []

    [access.teams]
    triage = "triage"

    [access.individuals]
    SomeoneToIgnore = "triage"
    pietroalbini = "admin"
    skade = "admin"
    pnkfelix = "admin"
    matklad = "admin"
    kjeremy = "triage"
    jD91mZM2 = "write"
    JoshMcguigan = "triage"
    wesleywiser = "admin"
    kiljacken = "triage"
    MissHolmes = "admin"
    lnicola = "admin"
    lf- = "triage"
    Veykril = "admin"
    edwin0cheng = "triage"
    Xanewok = "admin"
    dustypomerleau = "triage"
    flodiebold = "admin"
    Veetaha = "triage"
    rust-lang-owner = "admin"
    CAD97 = "write"
    bjorn3 = "triage"
    ```

    </details>

- [smol_str](https://github.com/rust-analyzer/smol_str)
    <details markdown="1">
    <summary>Original content</summary>

    ```toml
    org = "rust-analyzer"
    name = "smol_str"
    description = ""
[...]

PR #1854, opened 2025-05-30
Add `expect-test` team

Kobzol:
After https://github.com/rust-lang/team/pull/1805, @quad no longer access to maintain the `expect-test` repository. In [#t-compiler/rust-analyzer > Rust-analyzer org repository access and status @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Rust-analyzer.20org.20repository.20access.20and.20status/near/521110084), we have discussed how to resolve this situation, and the outcome was to create a `expect-test` team that would be responsible only for this repository.

We can do this in two ways, I suppose:
- Create a marker team (which is what this PR does)
- Create the team as a subteam of `rust-analyzer` (?), and add a website entry for it

The second approach seemed a bit heavyweight for a tiny team designed to maintain a single small Rust Analyzer dependency, so I went with the first.

### DIFF
--- a/people/quad.toml
+++ b/people/quad.toml
@@ -1,0 +1,4 @@
+name = "Scott Robinson"
+github = "quad"
+github-id = 24025
+zulip-id = 705090
--- a/repos/rust-analyzer/expect-test.toml
+++ b/repos/rust-analyzer/expect-test.toml
@@ -5,6 +5,7 @@ bots = []
 
 [access.teams]
 rust-analyzer = "write"
+expect-test = "write"
 
 [[branch-protections]]
 pattern = "master"
--- a/teams/expect-test.toml
+++ b/teams/expect-test.toml
@@ -1,0 +1,10 @@
+name = "expect-test"
+kind = "marker-team"
+
+[people]
+leads = ["quad"]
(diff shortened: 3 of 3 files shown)


ehuss:
I think in general we should avoid using marker teams to give permissions for things like this. I don't think there are any inherent downsides to just making these normal teams, whereas making them marker teams muddles the definition of what that team-kind is for and makes the responsibilities unclear. It's also not really clear why this is maintained by the rust-analyzer team since it seems like a general-purpose testing library unrelated to rust-analyzer (other than it is a user?).

Kobzol:
As noted in the PR description, I wasn't sure if this should be a marker team or not. Happy to send a PR that changes that and makes it a "proper" team with a website entry. Regarding the second question, I thought the same, it was discussed here: [#t-compiler/rust-analyzer > Rust-analyzer org repository access and status @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Rust-analyzer.20org.20repository.20access.20and.20status/near/521237984).

PR #1866, opened 2025-06-04
Sync organization members

Kobzol:
This PR finishes the work from https://github.com/rust-lang/sync-team/pull/91 (I used @me-diru's code, adapted it to the modified codebase and marked them as an author).

I guess that it would be good to repost this PR by an infra-admin, so that we could get a dry-run for it.

Fixes: https://github.com/rust-lang/team/issues/1731

PR #1868, opened 2025-06-04
Sync organization members

### LINKED ISSUE #1731: Remove users without a team from GitHub
Right now, we invite new users to the `rust-lang` GitHub organization when they are added to a team. Eventually, they might stop contributing to the team and leave it again. When that happens, they are removed from the team in the [rust-lang/team](https://github.com/rust-lang/team) repo.

When a user is no longer part of any team or working group, we should also remove them from the `rust-lang` GitHub organization. This a) cleans up the members of the organization, making it easier to manage and review them. And b) frees up seats in the organization. The latter will become more relevant in the future, when admins will need to manually add seats in case we run out of our current allocation.

emilyalbini:
Re-submitting #1866 *again* as requested:

> This PR finishes the work from https://github.com/rust-lang/sync-team/pull/91 (I used @me-diru's code, adapted it to the modified codebase and marked them as an author).
>
> I guess that it would be good to repost this PR by an infra-admin, so that we could get a dry-run for it.
>
> Fixes: https://github.com/rust-lang/team/issues/1731

cc @Kobzol 

- [x] Reach out to people that we are removing to thank them for their work and explain why we are removing them

Discussed in [#t-infra > removing inactive members from the org](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/removing.20inactive.20members.20from.20the.20org/with/528798074).

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -15,6 +15,12 @@ allowed-github-orgs = [
     "rust-analyzer",
 ]
 
+# GitHub organizations where member management is independent
+# and should not be automatically synchronized
+independent-github-orgs = [
+    "rust-embedded",
+]
+
 permissions-bors-repos = [
     "bors-kindergarten",
     "rust",
@@ -26,4 +32,26 @@ permissions-bools = [
     "crates-io-admin",
     "dev-desktop",
     "sync-team-confirmation",
+]
+
+# GitHub accounts that are allowed to stay in the GitHub organizations,
+# even if they may not be members of any team.
+# Note: Infra admins are automatically included from the infra-admins team.
+special-org-members = [
+    # Bots.
+    "bors",
+    "craterbot",
+    "rust-embedded-bot",
+    "rust-highfive",
+    "rust-lang-core-team-agenda-bot",
+    "rust-lang-glacier-bot",
+    "rust-lang-owner",
+    "rust-log-analyzer",
+    "rust-timer",
+    "rustbot",
+
+    # Ferrous systems employees who manage the github sponsors
+    # for rust-analyzer.
+    "MissHolmes",
+    "skade",
 ]
--- a/sync-team/src/github/mod.rs
+++ b/sync-team/src/github/mod.rs
@@ -3,6 +3,7 @@ mod api;
 mod tests;
 
 use self::api::{BranchProtectionOp, TeamPrivacy, TeamRole};
+use crate::Config;
 use crate::github::api::{GithubRead, Login, PushAllowanceActor, RepoPermission, RepoSettings};
 use log::debug;
 use rust_team_data::v1::{Bot, BranchProtectionMode, MergeBot};
@@ -18,8 +19,9 @@ pub(crate) fn create_diff(
     github: Box<dyn GithubRead>,
     teams: Vec<rust_team_data::v1::Team>,
     repos: Vec<rust_team_data::v1::Repo>,
+    config: Config,
 ) -> anyhow::Result<Diff> {
-    let github = SyncGitHub::new(github, teams, repos)?;
+    let github = SyncGitHub::new(github, teams, repos, config)?;
     github.diff_all()
 }
 
@@ -29,6 +31,7 @@ struct SyncGitHub {
     github: Box<dyn GithubRead>,
     teams: Vec<rust_team_data::v1::Team>,
     repos: Vec<rust_team_data::v1::Repo>,
+    config: Config,
     usernames_cache: HashMap<u64, String>,
     org_owners: HashMap<OrgName, HashSet<u64>>,
     org_members: HashMap<OrgName, HashMap<u64, String>>,
@@ -39,6 +42,7 @@ impl SyncGitHub {
         github: Box<dyn GithubRead>,
         teams: Vec<rust_team_data::v1::Team>,
         repos: Vec<rust_team_data::v1::Repo>,
+        config: Config,
     ) -> anyhow::Result<Self> {
         debug!("caching mapping between user ids and usernames");
         let users = teams
@@ -72,6 +76,7 @@ impl SyncGitHub {
             github,
             teams,
             repos,
+            config,
             usernames_cache,
             org_owners,
             org_members,
@@ -90,7 +95,7 @@ impl SyncGitHub {
         })
     }
 
-    // Collect all org members from the respective teams
+    /// Collect all org members from the respective teams
     fn get_org_members_from_teams(&self) -> HashMap<OrgName, HashSet<u64>> {
         let mut org_team_members: HashMap<OrgName, HashSet<u64>> = HashMap::new();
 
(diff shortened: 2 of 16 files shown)


marcoieni:
It's removing rust-lang-owner and other bots

marcoieni:
I tried pushing one commit that also updated master, but failed miserably, so I reverted it 😅

marcoieni:
Now Kobzol is not in the authors of the commit anymore. Sorry about that but that's the maximum I can get with my git skills 😂
image

sync-team/src/github/mod.rs
```diff
@@ -60,30 +64,105 @@ impl SyncGitHub {
...
+        for (org, toml_members) in toml_org_team_members {
+            let Some(gh_org_members) = self.org_members.get(&org) else {
+                panic!("GitHub organization {org} not found");
+            };
+
+            // Initialize `members_to_remove` to all GitHub members in the org.
+            // Next, we'll delete members from `members_to_remove` that don't respect certain criteria.
+            let mut members_to_remove = gh_org_members.clone();
+
+            // Remove all members that are in TOML teams.
```
samueltardieu: I find the "remove" here confusing. Maybe use "subtract" when talking about removal from the set?

It currently reads: "Remove all members that are in TOML teams", and a few lines later, "The rest are members that should be removed". I think this would be clearer: "Subtract all members that are in TOML teams" / "The rest are members that should be removed".

marcoieni: yeah, I added more comments because it was confusing to me as well in the first place. I will implement this suggestion, thanks!

marcoieni: I simplified the comments and extracted the code to a function. Let me know what you think.

config.toml
```diff
@@ -27,3 +27,29 @@ permissions-bools = [
     "dev-desktop",
     "sync-team-confirmation",
 ]
+
+# GitHub accounts that are allowed to stay in the orgs,
+# even if they are not members of any team.
```
tshepang: less confusing?
```suggestion
# even if they may not be members of any team.
```

marcoieni: changed this comment, thanks!

sync-team/src/github/mod.rs
samueltardieu: Much clearer indeed!

config.toml
```diff
@@ -27,3 +27,29 @@ permissions-bools = [
     "dev-desktop",
     "sync-team-confirmation",
 ]
+
+# GitHub accounts that are allowed to stay in the orgs,
+# even if they may not be members of any team.
+allowed-org-members = [
```
Kobzol: nit: `allowed` makes it sound like only these org members may be a part of an org :D Which of course doesn't make sense, and the comment does clarify it. But I don't have a great alternative. Maybe just `special-org-members`? But it doesn't really matter.

config.toml
```diff
@@ -27,3 +27,29 @@ permissions-bools = [
...
+    "rust-embedded-bot",
+    "rust-highfive",
+    "rust-lang-core-team-agenda-bot",
+    "rust-lang-glacier-bot",
+    "rust-lang-owner",
+    "rust-timer",
+    "rustbot",
+
+    # Infra admins.
+    "jdno",
```
Kobzol: Hmm, we start having a bunch of places where we hardcode infra-admins. Maybe it would be better if we loaded the actual contents of the `infra-admins` team (so that we hardcode only the team in code, but not its members) in code and ignored them?

rust_team_data/src/v1.rs
```diff
@@ -1,8 +1,15 @@
+use std::collections::BTreeSet;
+
 use indexmap::IndexMap;
 use serde::{Deserialize, Serialize};
 
 pub static BASE_URL: &str = "https://team-api.infra.rust-lang.org/v1";
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
```
Kobzol: I don't think we need to expose this to the JSON API? It's an internal implementation detail. Can't you just pass `&Config` from `team` to `sync-team` in `run_sync_team`?

Kobzol, line 8: I don't think we need to expose this to the API? It's an internal implementation detail. I guess that we don't currently pass the config from `team` to `sync-team`, right?

config.toml
tshepang: I like `special-org-members`

marcoieni:
addressed the feedback, thanks. Ready for another look 👍

config.toml
```diff
@@ -27,3 +33,24 @@ permissions-bools = [
...
+    # Bots.
+    "bors",
+    "craterbot",
+    "rust-embedded-bot",
+    "rust-highfive",
+    "rust-lang-core-team-agenda-bot",
+    "rust-lang-glacier-bot",
+    "rust-lang-owner",
+    "rust-timer",
+    "rustbot",
```
fmease, line 51: I think you want to add `rust-log-analyzer` here too, it's currently getting removed looking at the dry-run results.

marcoieni: done, thanks!

config.toml
```diff
@@ -27,3 +33,25 @@ permissions-bools = [
...
+    "rust-lang-core-team-agenda-bot",
+    "rust-lang-glacier-bot",
+    "rust-lang-owner",
+    "rust-log-analyzer",
+    "rust-timer",
+    "rustbot",
+
+    # Ferrous systems employees who manage the github sponsors
+    # for rust-analyzer.
+    "missholmes",
```
emilyalbini:
```suggestion
    "MissHolmes",
```

config.toml
emilyalbini: Special also doesn't clarify much, maybe call it `dont-remove-from-orgs = []`?

config.toml
emilyalbini: This is a good point, but also, I'm not sure if hardcoding the infra-admins team in the code is the best either. As we are not managing org owners with the team repo for now, I'd skip removing any org owner, and do a manual pass removing inactive org owners.

Kobzol: We already hardcode infra-admins for CODEOWNERS, but ignoring org owners is a good point.

Kobzol (approved):
Looks good, thanks!

jieyouxu:
Triage: waiting on the communication part [#t-infra > removing inactive members from the org](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/removing.20inactive.20members.20from.20the.20org/with/533124684).

programmerjake:
> Now Kobzol is not in the authors of the commit anymore. Sorry about that but that's the maximum I can get with my git skills 😂

if you want more than one author to be listed, you can use[ `Co-Authored-By: Their Name <their@email.example.com>`](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)

marcoieni (approved):

PR #1872, opened 2025-06-11
code cleaning

tshepang:


marcoieni:
CI is failing 😿

marcoieni (approved):

marcoieni:
Thanks! Do you want to raise another PR to bump the edition to 2024?

Cargo.toml
```diff
@@ -1,8 +1,6 @@
 [package]
 name = "rust-team"
-version = "0.1.0"
-authors = ["Alex Crichton <alex@alexcrichton.com>", "Pietro Albini <pietro@pietroalbini.org>"]
-edition = '2018'
+edition = '2021'
```
Kobzol, line 3: Why remove the authors?

tshepang: it would not be too feasible adding them all

apiraino: If your intent was to include everyone maybe a good alternative could be: "The Rust Project team" or something similar.

In any case: I agree it was not such a fundamental thing to remove :slightly_smiling_face:

tshepang: that would be redundant in that all repos in rust-lang are maintained by members of the Rust project

also, the field is deprecated

apiraino: > also, the field is deprecated

Oh this is news to me, I didn't know
https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field

Thanks, TIL :smiley:

PR #1889, opened 2025-06-26
add trusted publishing action repo

marcoieni:
[Trusted publishing](https://rust-lang.github.io/rfcs/3691-trusted-publishing-cratesio.html) is working on crates.io staging.

I worked on an [action](https://github.com/marcoieni/trusted-publishing-action-draft) to allow users to use trusted publishing on GitHub Actions.

In this PR, I create the repo for the action. After the repo has been created, I will push my code in that repo.

After the repo is created, I will edit this file again to add branch protections and maybe give infra team write access, but we need to make sure that only infra admins can push tags to that repo.

After my code is merged, I'm happy to receive PRs to change what you don't like about the action 👍

Regarding the license of the repo, is it enough to copy these files in the repo:
- https://github.com/rust-lang/rust/blob/master/LICENSE-APACHE 
- https://github.com/rust-lang/rust/blob/master/LICENSE-MIT

And in the readme we could write the following:

```
Licensed under either of [./LICENSE-APACHE](Apache License, Version 2.0) or [./LICENSE-MIT](MIT license) at your option.
```

Not sure if we have any guidelines around licenses in the project.

### DIFF
--- a/repos/rust-lang/crates-io-auth-action.toml
+++ b/repos/rust-lang/crates-io-auth-action.toml
@@ -1,0 +1,8 @@
+org = "rust-lang"
+name = "crates-io-auth-action"
+description = "Get a crates.io temporary access token"
+bots = []
+
+[access.teams]
+infra-admins = "write"
+infra = "triage"


jdno:
I don't think we have general guidelines around licenses. Since the action will not be written in Rust and is not designed to be used as a crate in other Rust projects, I don't think it is necessary to dual license it. My suggestion would be to simply publish it under the MIT license.

repos/rust-lang/crates-io-auth-action.toml
```diff
@@ -0,0 +1,8 @@
+org = "rust-lang"
+name = "crates-io-auth-action"
```
jdno, line 2: Since the name of the repository will also be the official identifier of the GitHub Action and it will be hard to change after we go live, I'd like to get feedback from a few other folks on it.

@rust-lang/infra @rust-lang/crates-io

marcoieni: actually it's easy to change it after going live. you can just rename the repo, and github actions automatically redirects to the new one.

Anyway, this is the name used in the RFC.

jdno: Oh, I missed that it was named like this in the RFC. Guess everyone had a chance to comment on it, then. 😁

jdno (approved):
With the name coming from the RFC, I'm fine getting this rolling...

emilyalbini:
> I don't think we have general guidelines around licenses. Since the action will not be written in Rust and is not designed to be used as a crate in other Rust projects, I don't think it is necessary to dual license it. My suggestion would be to simply publish it under the MIT license.

There is some Rust code in the `mock/` directory inside the repository. Having `MIT OR Apache-2.0` as the license will allow to copy code from the crates.io codebase to the action (or the inverse). Not sure when or if it will be needed, but let's avoid licensing problems IMO.

Turbo87:
> we need to make sure that only infra admins can push tags to that repo

why is this limited to infra admins? wouldn't it make sense to give the crates.io team access too? 🤔

marcoieni:
> why is this limited to infra admins?

I thought that for security reasons we only want infra-admins to allow push tags. Because I wanted to setup CI so that if anyone pushes v1.2.3 than the tag v1 is updated to the latest automatically. One security threat would be that people edit the code of the action to send tokens to them.

emilyalbini:
A compromised account of a member of the crates.io team (at least one with Heroku access) could achieve the same thing by messing with the production deployment anyway. I think it's fine to give access to the crates.io team, or at least the subset of the crates.io team that has database access.

marcoieni:
Opened https://github.com/rust-lang/crates-io-auth-action/issues/13 to track the crates-io team discussion 👍

PR #1891, opened 2025-06-30
Add more blocking CI checks to rustc-perf

Kobzol:
The CI didn't block on compile benchmarks failing, which is annoying (https://github.com/rust-lang/rustc-perf/pull/2176).


Kobzol:
Looks like a spurious error.

PR #1895, opened 2025-07-04
Archive `rust-lang/generate-manifest-list`

pietroalbini:
The tool and the GitHub Actions cron running on it is not needed anymore, as `manifests.txt` is now updated by the release process.


Kobzol:
Yay, no more manual interventions when the cron job stops working after 6 months of inactivity :tada:

PR #1896, opened 2025-07-04
Require tests to pass on `josh-sync`

Kobzol:
CC @marcoieni :innocent:

https://github.com/rust-lang/josh-sync/blob/main/.github/workflows/test.yml#L7


Kobzol:
Deploy error looked spurious.

PR #1905, opened 2025-07-10
fix clippy issues

marcoieni:
Let's run clippy in CI to improve the quality of the codebase

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,11 +29,13 @@ jobs:
           fetch-depth: 50
 
       - name: Install Rust Stable
+        env:
+          RUST_VERSION: "1.85.0"
         run: |
           rustc -vV
-          rustup toolchain install 1.85.0
-          rustup default 1.85.0
-          rustup component add rustfmt
+          rustup toolchain install $RUST_VERSION
+          rustup default $RUST_VERSION
+          rustup component add rustfmt clippy
           rustc -vV
 
       - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6
@@ -48,6 +50,9 @@ jobs:
 
       - name: Run rustfmt
         run: cargo fmt -- --check
+
+      - name: Run clippy
+        run: cargo clippy --workspace --all-targets --all-features -- -D warnings
 
       - name: Run tests
         run: cargo test --workspace --all-features
--- a/rust_team_data/src/email_encryption.rs
+++ b/rust_team_data/src/email_encryption.rs
@@ -79,8 +79,8 @@ pub enum Error {
 impl std::fmt::Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
-            Error::GetRandom(e) => write!(f, "{}", e),
-            Error::Hex(e) => write!(f, "{}", e),
+            Error::GetRandom(e) => write!(f, "{e}"),
+            Error::Hex(e) => write!(f, "{e}"),
             Error::EncryptionFailed => write!(f, "encryption failed"),
             Error::DecryptionFailed => write!(f, "encryption failed"),
             Error::InvalidUtf8 => write!(f, "invalid UTF-8"),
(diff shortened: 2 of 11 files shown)


sync-team/src/github/tests/mod.rs
```diff
@@ -19,7 +19,11 @@ fn team_create() {
     let user = model.create_user("mark");
     let user2 = model.create_user("jan");
     let gh = model.gh_model();
-    model.create_team(TeamData::new("admins").gh_team(DEFAULT_ORG, "admins-gh", &[user, user2]));
+    model.create_team(TeamData::new_builder("admins").gh_team(
```
marcoieni: I had to rename `new` because that method wasn't returning Self. So I called the method `new_builder`.

An alternative is calling it just `builder`

Kobzol: `build`? It's annoying though, I would just silence the lint on this method.

Kobzol (approved):

.github/workflows/main.yml
```diff
@@ -49,6 +51,9 @@ jobs:
       - name: Run rustfmt
         run: cargo fmt -- --check
 
+      - name: Run clippy
+        run: cargo clippy --workspace --all-targets --all-features -- -D warnings
```
apiraino, line 55: this will imply that contributions to the codebase must now pass clippy, and no warnings allowed.
I see that a clippy lint is enforcing two rules:
- `#![allow(clippy::enum_variant_names)]`
- `#[allow(clippy::type_complexity)]`

apiraino: Also one question (for my curiosity): I think I don't see a specific clippy lint config file. What happens if clippy "stabilizes" and adds a new default?

apiraino:
Is the diff mostly stylistic changes? How do they improve the quality of the code?

.github/workflows/main.yml
marcoieni: > this will imply that contributions to the codebase must now pass clippy

yes. I find clippy is a good teacher for new contributors. They learn about idiomatic rust 👍

> and no warnings allowed

I think that all clippy lints are treated as warnings, so if you don't add `-D warnings` then this step will pass 😅



> What happens if clippy "stabilizes" and adds a new default?

clippy adds new defaults on every rust release more or less. We pin the rust version in CI, so whenever we bump the rust version we need to fix new clippy lints, too (if any)

marcoieni:
> How do they improve the quality of the code?

the code is more readable imo

e.g. 
image
image

.github/workflows/main.yml
```diff
@@ -29,11 +29,13 @@ jobs:
           fetch-depth: 50
 
       - name: Install Rust Stable
+        env:
+          RUST_VERSION: "1.85.0"
```
emilyalbini, line 33: If we are pinning the Rust version, can we do so with a `rust-toolchain.toml` rather than an environment variable? The way the variable is setup right now is quite hard to discover...

sync-team/src/github/tests/mod.rs
emilyalbini: I also agree to silence that lint.

.github/workflows/main.yml
marcoieni: Can I do it in another PR as it's unrelated to this one? This is just a small refactoring to avoid specifying the version twice.

marcoieni: The version was already pinned in CI

emilyalbini: Oh, I missed that we were already pinning it. Then sure.

sync-team/src/github/tests/mod.rs
jdno: Personally, I'd stick with the convention here and rename the method to something like `builder`. We're probably all using code completion, so there's imo no cost in adding a few characters to the method name. 😅 

But not a blocking concern, happy to go with whatever the majority thinks...

jdno (dismissed):
Great find! Will help a lot with long-term maintenance of the code base. 🙂

sync-team/src/github/tests/mod.rs
marcoieni: I went for `#[allow(clippy::new_ret_no_self)]` since it's the less destructive option. We can always remove it in another PR

jdno (approved):

PR #1919, opened 2025-07-28
ensure team name equals the name of the first GitHub team

marcoieni:
I find confusing that the github `mods` team doesn't match the `mods.toml` team. This PR solves this issues, but I'm not sure it's safe to apply or if there's historic context I'm missing.

- [ ] check that this won't break dev desktops (is the sync checking github api or team api?)
- [ ] check that no consumers of the team API will be impacted by this


src/validate.rs
```diff
@@ -1111,3 +1112,22 @@ fn validate_website(data: &Data, errors: &mut Vec<String>) {
...
     })
 }
+
+/// Ensure team name equals the name of the first GitHub team
+fn validate_team_name_matches_github_team(data: &Data, errors: &mut Vec<String>) {
+    wrapper(data.teams(), errors, |team, _| {
+        let github_teams = team.github_teams(data)?;
+        if let Some(first_github_team) = github_teams.first() {
+            let should_ignore_team =
+                team.name().starts_with("project-") || team.name().starts_with("wg-");
```
marcoieni, line 1122: this is what `cargo run check` looks like if we remove this:

```
[ERROR rust_team::validate] validation error: team name `project-async-crashdump-debugging` should match the first GitHub team-name `initiative-async-crashdump-debugging`
[ERROR rust_team::validate] validation error: team name `project-dyn-upcasting` should match the first GitHub team-name `initiative-dyn-upcasting`
[ERROR rust_team::validate] validation error: team name `project-generic-associated-types` should match the first GitHub team-name `initiative-generic-associated-types`
[ERROR rust_team::validate] validation error: team name `project-impl-trait` should match the first GitHub team-name `initiative-impl-trait`
[ERROR rust_team::validate] validation error: team name `project-keyword-generics` should match the first GitHub team-name `initiative-keyword-generics`
[ERROR rust_team::validate] validation error: team name `project-negative-impls` should match the first GitHub team-name `initiative-negative-impls`
[ERROR rust_team::validate] validation error: team name `project-trait-system-refactor` should match the first GitHub team-name `initiative-trait-system-refactor`
[ERROR rust_team::validate] validation error: team name `wg-embedded-arm` should match the first GitHub team-name `Arm`
[ERROR rust_team::validate] validation error: team name `wg-embedded-core` should match the first GitHub team-name `core`
[ERROR rust_team::validate] validation error: team name `wg-embedded-hal` should match the first GitHub team-name `embedded-hal`
[ERROR rust_team::validate] validation error: team name `wg-embedded-infra` should match the first GitHub team-name `infrastructure`
[ERROR rust_team::validate] validation error: team name `wg-embedded-libs` should match the first GitHub team-name `libs`
[ERROR rust_team::validate] validation error: team name `wg-embedded-linux` should match the first GitHub team-name `embedded-linux`
[ERROR rust_team::validate] validation error: team name `wg-embedded-msp430` should match the first GitHub team-name `msp430`
[ERROR rust_team::validate] validation error: team name `wg-embedded-resources` should match the first GitHub team-name `resources`
[ERROR rust_team::validate] validation error: team name `wg-embedded-riscv` should match the first GitHub team-name `risc-v`
[ERROR rust_team::validate] validation error: team name `wg-embedded-tools` should match the first GitHub team-name `tools`
[ERROR rust_team::validate] validation error: team name `wg-embedded-triage` should match the first GitHub team-name `triage`
[ERROR rust_team::validate] validation error: team name `wg-security-response` should match the first GitHub team-name `security`
```

Kobzol:
Probably another place where we might want to ignore `wg-embedded`, to avoid needless churn for them.. or just rename the `toml` files instead of the GH team, that's probably a better idea. But that reminds me of a problem with this approach - our teams are not scoped per org. So if there are two different (GH) teams in two different GH orgs, naming the TOML file after the GitHub team won't really work, because their names would collide.

marcoieni:
> Probably another place where we might want to ignore `wg-embedded`, to avoid needless churn for them

I already ignore `wg-*` in the validation 🤔



> or just rename the `toml` files instead of the GH team, that's probably a better idea

That's what I did 🤔


> So if there are two different (GH) teams in two different GH orgs, naming the TOML file after the GitHub team won't really work, because their names would collide.

Right but maybe this easily solvable by naming the gh team differently.

For example, I imagine that if there's a `rust-lang-nursery` moderation team that needs to be different from the rust-lang moderation team, we could call that team `nursery-mods`.

Or we could also explicitly disable this validation for that team. But this should be a sane default imo.

Like, why is it useful to have `cloud-compute-team` vs `cloud-compute`? 😵‍💫

Kobzol:
> For example, I imagine that if there's a rust-lang-nursery moderation team that needs to be different from the rust-lang moderation team, we could call that team nursery-mods.

It's weird to name GH teams with a prefix when they are already scoped by the organization :sweat_smile:

I'd personally rather change the semantics of the LHS rather than enforce this, seeing the edge cases that we have to skip and the issue with team names across organizations.

marcoieni:
I'll close this PR then, and work on the LHS change next 👍

PR #1920, opened 2025-07-30
add tomassedovic to vision doc team

nikomatsakis:
Give @tomassedovic access to the vision doc group so he can help with project tasks. Proposed by @traviscross.

cc @jackh726, project co-lead


nikomatsakis:
Discussed some with @jackh726. He made the solid point that we don't actually have a particular task that needs doing at the moment, and we'd prefer to keep the size of the team small so as to limit access to confidential data. Therefore, going to close this PR until such time as have a clear need.

PR #1923, opened 2025-07-31
team access refers to team name of toml instead of github

marcoieni:
We discussed with the infra team that it's confusing to use the github name as the LHS of the `[access.teams]`, so we want to use the team name (i.e. the `name` field of the team) instead.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -24,8 +24,11 @@
 /teams/team-repo-admins.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/Mark-Simulacrum.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/Nadrieril.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
+/people/Noratrieb.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
+/people/apiraino.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/jackh726.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/jdno.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
+/people/jieyouxu.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/marcoieni.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/oli-obk.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/pietroalbini.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
--- a/src/static_api.rs
+++ b/src/static_api.rs
@@ -93,23 +93,35 @@ impl<'a> Generator<'a> {
                         Bot::Renovate => v1::Bot::Renovate,
                     })
                     .collect(),
-                teams: r
-                    .access
-                    .teams
-                    .iter()
-                    .map(|(name, permission)| {
+                teams: {
+                    let mut teams = Vec::new();
+                    for (team_name, permission) in &r.access.teams {
                         let permission = match permission {
                             RepoPermission::Admin => v1::RepoPermission::Admin,
                             RepoPermission::Write => v1::RepoPermission::Write,
                             RepoPermission::Maintain => v1::RepoPermission::Maintain,
                             RepoPermission::Triage => v1::RepoPermission::Triage,
                         };
-                        v1::RepoTeam {
-                            name: name.clone(),
-                            permission,
+
+                        // Look up the team by name and get all its GitHub teams
+                        let team = self.data.team(team_name).with_context(|| {
+                            format!("failed to find team '{team_name}' in teams directory")
+                        })?;
+                        let github_teams = team.github_teams(self.data).with_context(|| {
+                            format!("failed to get GitHub teams for '{team_name}'")
+                        })?;
+                        for gh_team in github_teams {
+                            if gh_team.org == r.org {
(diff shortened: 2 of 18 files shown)


.github/CODEOWNERS
```diff
@@ -24,8 +24,11 @@
 /teams/team-repo-admins.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/Mark-Simulacrum.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/Nadrieril.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
+/people/Noratrieb.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
+/people/apiraino.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/jackh726.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
 /people/jdno.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
+/people/jieyouxu.toml @Mark-Simulacrum @pietroalbini @jdno @marcoieni
```
marcoieni, line 31: these people have write access to the repo, so it was a bug that they weren't present in the codeowners file

src/validate.rs
```diff
@@ -900,6 +890,33 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
...
+) -> anyhow::Result<()> {
+    let team = data
+        .team(team_name)
+        .with_context(|| format!("The file 'teams/{team_name}.toml' does not exist."))?;
+    let github_teams = team.github_teams(data)?;
+    let mut team_found = false;
+    for github_team in &github_teams {
+        if all_github_teams.contains(&(repo.org.clone(), github_team.name.to_string())) {
+            team_found = true;
+        } else if data.is_team_archived(github_team.name, &repo.org) {
```
Kobzol: The function name is confusing, it means `is_github_team_archived`, right? But I don't think we need to check GH team archival status after this redesign. We should just check if the team itself is archived or not, and that's it. GH teams don't have any additional archival status.

src/validate.rs
```diff
@@ -900,6 +890,33 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
...
+    repo: &Repo,
+    team_name: &str,
+) -> anyhow::Result<()> {
+    let team = data
+        .team(team_name)
+        .with_context(|| format!("The file 'teams/{team_name}.toml' does not exist."))?;
+    let github_teams = team.github_teams(data)?;
+    let mut team_found = false;
+    for github_team in &github_teams {
+        if all_github_teams.contains(&(repo.org.clone(), github_team.name.to_string())) {
```
Kobzol: Why do you check `all_github_teams`? I don't think that we expect the corresponding GH team to come from anywhere else than the `team` itself. So essentially we want something like this:
```rust
let gh_teams: Vec<_> = github_teams.iter().filter(|t| t.org == repo.org).collect();
if gh_teams.is_empty() {
    bail!(
        "The team '{team_name}' does not have any GitHub teams in the org '{}'.",
        repo.org
    );
}
```

src/validate.rs
```diff
@@ -900,6 +890,33 @@ fn validate_repos(data: &Data, errors: &mut Vec<String>) {
...
+fn check_team_access(
+    data: &Data,
+    all_github_teams: &HashSet<(String, String)>,
+    repo: &Repo,
+    team_name: &str,
+) -> anyhow::Result<()> {
+    let team = data
+        .team(team_name)
+        .with_context(|| format!("The file 'teams/{team_name}.toml' does not exist."))?;
+    let github_teams = team.github_teams(data)?;
```
Kobzol: Team archival status can be checked e.g. here.

src/static_api.rs
```diff
@@ -93,23 +93,37 @@ impl<'a> Generator<'a> {
...
-                            permission,
+
+                        // Look up the team by name and get all its GitHub teams
+                        let team = self.data.team(team_name).with_context(|| {
+                            format!("failed to find team '{team_name}' in teams directory")
+                        })?;
+                        let github_teams = team.github_teams(self.data).with_context(|| {
+                            format!("failed to get GitHub teams for '{team_name}'")
+                        })?;
+                        for gh_team in github_teams {
```
Kobzol, line 113: I'm not sure if this is correct/what we want. I think that at the logical level of the REST API, we still want to represent that a specific *team* has access to a repository, without expanding it into GH teams here. We should only do the expansion in `sync-team` where we convert this logical representation into actual GH permissions.

Of course this will require changes in sync-team, in particular around `calculate_permission_diffs`. We should then also explicitly represent both the org name and the team name in the diffs.

Kobzol (commented):
Thanks for working on this! I think that we should use a different model for expanding the teams, clarified in comments.

src/static_api.rs
marcoieni: I marked this as resolved after a discussion I had with Kobzol.
We both agree it's better to keep things as they are to avoid making breaking changes to the API.

src/static_api.rs
```diff
@@ -93,23 +93,37 @@ impl<'a> Generator<'a> {
...
+                            if gh_team.org == r.org {
+                                let new_team = v1::RepoTeam {
+                                    name: gh_team.name.to_string(),
+                                    permission: permission.clone(),
+                                };
+                                if !teams.contains(&new_team) {
```
Kobzol: Just wondering, how could this ever happen? Two teams shouldn't ever share the same GH team, right?

marcoieni: yes, removed this check 👍

Kobzol (approved):

jdno (approved):

PR #1926, opened 2025-08-05
Update Vision Doc team membership

jackh726:
This better reflects the activity of the team.


nikomatsakis:
(For context, our original plan was to let anybody join, but to stay on as a member you were expected to conduct 3 or more interviews and be an active participant. The folks we're narrowing down to are the ones who met this criteria.)

Mark-Simulacrum (approved):

PR #1929, opened 2025-08-06
Move tgnottingham to alumni status in wg-compiler-performance

tgnottingham:


tgnottingham:
Was great to contribute, but I haven't done so in a long time and don't expect to in the future. :)

jackh726 (approved):

PR #1930, opened 2025-08-07
put myself as spec team lead

nikomatsakis:
Per discussion, I am stepping in as interim lead of the T-spec team, with the goal of (among other things) finding a longer-term replacement. Many thanks to @JoelMarcey for leading the team thus far!


nikomatsakis:
cc @rust-lang/spec

PR #1937, opened 2025-08-08
Create the `rust-heroku-deploy-access` bot

emilyalbini:
This bot will have write access to the GitHub repositories we link to Heroku. Adding the bot to a repo will allow linking an Heroku app to the repo via the Heroku Dashboard.

I already created the bot, added it to the org, and created an Heroku account for it. Once this is merged, we will stop linking Heroku and GitHub through my personal account.

### DIFF
--- a/repos/rust-lang/crates.io.toml
+++ b/repos/rust-lang/crates.io.toml
@@ -2,7 +2,7 @@ org = "rust-lang"
 name = "crates.io"
 description = "The Rust package registry"
 homepage = "https://crates.io"
-bots = ["renovate", "rustbot"]
+bots = ["renovate", "rustbot", "heroku-deploy-access"]
 
 [access.teams]
 crates-io = "write"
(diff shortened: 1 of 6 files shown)


Kobzol:
This broke triagebot :sweat_smile: (https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Team.20API.20not.20accessible.20in.20GHA.20log.20viewer/with/533492196). I wonder if we can have some tests for this (or just use semver properly, but that would be a bit annoying).

marcoieni:
> just use semver properly

but this is not an incompatible change, right? maybe the issue is in triagebot that doesn't allow the list of bots to grow? 🤔

Kobzol:
Adding a new variant to an enum not marked as non exhaustive is a BC break, so this is a backwards incompatible change. Although for triagebot the issue was in serde::Deserialize seeing an unexpected variant.

marcoieni:
> Although for triagebot the issue was in serde::Deserialize seeing an unexpected variant.

yeah, I was talking about the API. If triagebot consumes the data with a rust library, than yes, we should have bumped the version. In the future we could just mark the invariant as non exhaustive.

PR #1932, opened 2025-08-07
updates twir reviewers

nellshamrell:

PR #1938, opened 2025-08-08
Recognize TWIR reviewers alumni

jieyouxu:
Follow-up to #1932 which only removed members to instead move to alumni, as suggested in https://github.com/rust-lang/team/pull/1932#issuecomment-3166296450.


jieyouxu:
FYI @nellshamrell

nellshamrell:
This makes much more sense! Thank you for submitting this pull request!

jackh726 (approved):

PR #1942, opened 2025-08-10
Add Kivooeo to cloud-compute

Kivooeo:
Hi there!

I'd like to request access to the `cloud-compute` marker team, as suggested by a member of the infra team during our recent conversation

To briefly outline the issue: my current machine struggles with resource constraints when running RA alongside compiler builds. This results in 5-10 minute build times during which all available RAM is consumed, making it quite uncomfortable to work during that period


jieyouxu (approved):
This was discussed internally by the infra team (the email requesting access for whatever reason never went through). The consensus was that as long as a team member (of any of the maintenance teams) vouches for the dev-desktop access, this is fine.

I'll vouch for Kivo for their compiler/library side contributions.

This will need another team-repo-admin to approve, as I should refrain from approving my own vouch.

PR #1943, opened 2025-08-11
create the crates-io-infra-admins marker team

marcoieni:
This team contains people with elevated crates-io priviledges.
In a follow-up PR, I would like to give these people write access to https://github.com/rust-lang/crates-io-auth-action/ to solve https://github.com/rust-lang/crates-io-auth-action/issues/13

### DIFF
--- a/teams/crates-io-infra-admins.toml
+++ b/teams/crates-io-infra-admins.toml
@@ -1,0 +1,17 @@
+# This team defines which members of the crates-io team have administrative access to the
+# crates.io infrastructure, such as the database.
+
+name = "crates-io-infra-admins"
+kind = "marker-team"
+
+[people]
+leads = []
+members = [
+    "carols10cents",
+    "jtgeibel",
+    "Turbo87",
+    "LawnGnome",
+]
+
+[[github]]
+orgs = ["rust-lang"]


teams/crates-io-infra-admins.toml
```diff
@@ -0,0 +1,17 @@
+# This team defines which members of the crates-io team have administrative access to the
+# crates.io infrastructure, such as the database.
+
+name = "crates-io-infra-admins"
+kind = "marker-team"
```
jieyouxu, line 5: Question: this file and files of the members should be protected by CODEOWNERS, right?

marcoieni: mmmh I don't think so because these people don't have admin access to this repo. If an attacker gets access to their credentials, they have no interest in making changes to this repo imo.

Wdyt? 🤔

jieyouxu: Ah right, good point, that SGTM :+1:

Turbo87 (approved):
I'm approving this, but I wonder if this should also include the regular infra admins? can teams "inherit" members from other teams? 😅

marcoieni:
mmmh, I'm not sure how useful it is to add infra admins here as well. We can add it later if needed.

marcoieni:
> can teams "inherit" members from other teams? 😅

btw, yes

https://github.com/rust-lang/team/blob/bd9df6889bcf7ef2e7498579f2b1a0987d06ac86/teams/inside-rust-reviewers.toml#L7

jieyouxu (approved):
This was discussed in https://github.com/rust-lang/team/pull/1889#issuecomment-3008175282 re. trusted publishing

> A compromised account of a member of the crates.io team (at least one with Heroku access) could achieve the same thing by messing with the production deployment anyway. I think it's fine to give access to the crates.io team, or at least the subset of the crates.io team that has database access.

PR #1945, opened 2025-08-13
remove davidtwco from some groups

davidtwco:
I'm not active in mir-opts or const generics and don't often have much to add to issues that these groups are pinged on, I can still be pinged individually when that's not the case, but this will cut down on my notification load.


jackh726:
Do you want to be added as alumni for these groups?

davidtwco:
> Do you want to be added as alumni for these groups?

I guess I might as well, changed to that

PR #1947, opened 2025-08-14
Add the8472 to libs-api

Amanieu:


jieyouxu (approved):
Proposed by libs co-lead

Issue #1842, opened 2025-05-24
Dry-check should check for minimum `write` access to `<repo>` in `bors.<repo>.review = true`


jieyouxu:
See e.g.https://github.com/rust-lang/team/pull/1841

If a team has `bors.rust.review = true`, they can r+ PRs on rust-lang/rust. Surely if they can r+ PRs they should at the minimum have write access to said repo? Otherwise, the team won't even have triaging perms.

jieyouxu:
(Note: I'm not sure what coherence checks this repo has)

Kobzol:
Hmm, I guess the issue is that bors permissions are kind of opaque to `team`. It treats `rust` as just an arbitrary string, and there's no connection to the `rust-lang/rust` repository in `team`. I guess that we could hardcode it though.

Also I wanted to move the bors permissions from teams/users to the repository itself, to make it clearer who has permissions for a given repository.

PR #1948, opened 2025-08-15
Grant rustfmt team write access to `rust-lang/rust`

jieyouxu:
rustfmt team already has bors r+ permissions on `rust-lang/rust`.

https://github.com/rust-lang/team/blob/b10e72642a582b455ab8378d9a8a1217e6dcb94a/teams/rustfmt.toml#L19-L21

They should receive write access too, otherwise weird situations like "rustfmt team members can approve a PR, but can't even get assigned" happens. Noticed this in https://github.com/rust-lang/rust/pull/144323#issuecomment-3104648106. Same thing as https://github.com/rust-lang/team/pull/1841 (but that was for r-a team).

cc @ytmimi

calebcartwright was assignable already because both testing-devex and style teams have write access to r-l/r already, and they're a member of those teams.

FYI @rust-lang/rustfmt 

cc #1842


ytmimi:
Thank you!

PR #1973, opened 2025-09-02
Add the internal-sites team

ehuss:
This adds a team which is responsible for supporting internal-facing websites (that aren't already managed by more specific teams). Currently this is the [Rust Forge](https://forge.rust-lang.org/) and the [RFC Book](https://rust-lang.github.io/rfcs/). This was discussed a while back on Zulip [#**t-infra>rfc and forge maintenance**](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/rfc.20and.20forge.20maintenance).

cc @rust-lang/infra 

The charter for the team is:

## Responsibilities

The internal-sites team is responsible for maintaining websites used internally by the project. These sites include:

- [Rust Forge](https://forge.rust-lang.org/)
- [RFC Book](https://rust-lang.github.io/rfcs/)

The team is largely not responsible for creating content on these sites. That is the responsibility of the teams who post content on the sites. There are a few rare exceptions, such as the front page of these sites, which the team can update in coordination with the affected teams.

Maintenance activities include:

- Ensure the sites continue to function.
- Supporting teams that need to use and publish on the sites.
    - This also includes ensuring teams have sufficient access to do what they need.
- Updating and maintaining tool configuration used to publish the sites. This includes:
    - mdBook configuration, and the extensions used with it.
    - GitHub Actions publishing workflows.
    - GitHub Pages website deployments.
    - CDN integration (currently CloudFront).
- Updating and maintaining tooling integrated into the sites. This includes:
    - [Blacksmith](https://github.com/rust-lang/rust-forge/tree/master/blacksmith)
    - [RFC book generator](https://github.com/rust-lang/rfcs/blob/master/generate-book.py)
- Updating and maintaining bot configurations used on the sites. These include:
    - Triagebot
    - Renovate
    - Dependabot
- Maintaining the general organization of the sites.
    - Some of this is delegated to individual teams. For example, on the Forge, teams can set up their own organization for pages maintained by those teams.

The parent infra team has oversight of the internal-sites team. They are responsible for helping the internal-sites team remain operational.

## Team membership

Team membership is open to any project member who expresses interest, and is largely familiar with most or all of the tooling mentioned above.

The team leads are responsible for:
- Coordinating the team.
- Providing a point of contact with other teams.

The team leads are selected by consensus of the existing members with no objections (or the parent infra team if there are no members).

## Decision process

Decisions around major changes are done with a consensus process. These include things such as:
- Significant changes to this charter. For example, changing the list of sites in the responsibility list.
- Significant changes to tooling, such as introducing new tooling.


jieyouxu:
cc infra co-leads @shepmaster as a parent team co-lead needs to sign-off

shepmaster:
I approve this.

This seems like a general good thing ("maintain") that falls roughly in the world of infra ("used internally"). It also seems like a decision that is easily undone or amended if it proves to be a bad idea for any reason.

Thanks!

PR #1974, opened 2025-09-02
chore: Add ci checks to annotate-snippets-rs

Muscraft:
This PR adds a CI check for [`annotate-snippets` `ci` job](https://github.com/rust-lang/annotate-snippets-rs/blob/974a75be76c36b35f28faba79c289bc3377a8f97/.github/workflows/ci.yml#L22)

### DIFF
--- a/repos/rust-lang/annotate-snippets-rs.toml
+++ b/repos/rust-lang/annotate-snippets-rs.toml
@@ -9,4 +9,5 @@ wg-diagnostics = "write"
 
 [[branch-protections]]
 pattern = "master"
+ci-checks = ["CI"]
 required-approvals = 0


repos/rust-lang/annotate-snippets-rs.toml
```diff
@@ -9,4 +9,5 @@ wg-diagnostics = "write"
 
 [[branch-protections]]
 pattern = "master"
+ci-checks = ["ci"]
```
marcoieni, line 12:
```suggestion
ci-checks = ["CI"]
```

Since the job has a name, I think the ci-check should respect that name. E.g. see

https://github.com/rust-lang/crates-io-auth-action/blob/9dc75e2fccbefc31e71a232eaee19726b2c9b446/.github/workflows/ci.yml#L181

and

https://github.com/rust-lang/team/blob/e9dc0fab73d8928993e794cba9530f0adf96b92e/repos/rust-lang/crates-io-auth-action.toml#L13

PR #1975, opened 2025-09-03
Add a project goal team

joshtriplett:
Needed for repository permissions.


Kobzol:
More specifically, this is needed for https://github.com/rust-lang/team/pull/1972.

Kobzol:
To provide some context, Josh would like to move this repo (https://github.com/joshtriplett/project-goal-reference-expansion) under rust-lang. The repository should serve for coordinating work on a Project goal related to the Rust reference.

We need to provide some people access to the repo, of course. These people do not necessarily fall under any existing organizational structure, which brings a common tension:
- The easiest solution is to just provide a bunch of individuals with access. However, we have been trying to move away from individual accesses as much as possible.
- We can create a sort of an artificial team for just this repository (which is what this PR does).

@traviscross Any suggestions?

jieyouxu:
Maybe launching-pad...?

traviscross (approved):
Approach makes sense.  Looks good to me.

cc @ehuss

traviscross:
> Maybe launching-pad...?

No, the PR as-is is right.  Lang-docs is the appropriate parent team for this.

PR #1970, opened 2025-08-30
Create `cloud-compute` zulip user group

jieyouxu:
This is needed to configure the infra announcement channel so the zulip user group membership is synced automatically with marker team membership.

Accompanies https://github.com/rust-lang/team/pull/1969.

See [#t-infra > Communication method for dev desktops @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Communication.20method.20for.20dev.20desktops/near/536886782) for more details.

PR #1982, opened 2025-09-17
Add Zulip id for myself

Kivooeo:
Seems like it causes some problems? According to https://github.com/rust-lang/team/pull/1970

So here is fix for me


people/kivooeo.toml
```diff
@@ -2,3 +2,4 @@ name = "Kivooeo"
 github = "Kivooeo"
 github-id = 75776246
 email = "Kivooeo123@gmail.com"
+zulip-id = 675515
```
jieyouxu, line 5: Remark: added a trailing newline.

PR #1986, opened 2025-09-20
Add archived teams to v1 API

Kobzol:
Required for https://github.com/rust-lang/www.rust-lang.org/issues/2181. I decided to use a separate file and directory for the archived teams, rather than just adding an `archived: bool` field, for two reasons:

- It maintains backwards compatibility, e.g. for the Rust website, otherwise it would immediately start rendering archived teams in the governance section.
- It correponds to how we store the archived teams on disk in the `team` repo.

It is inconsistent how we represent archived repositories in the API output though, those have the `archived` bool.

### DIFF
--- a/src/static_api.rs
+++ b/src/static_api.rs
@@ -163,13 +163,13 @@ impl<'a> Generator<'a> {
     }
 
     fn generate_teams(&self) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


marcoieni:
there's a merge conflict unfortunately

Kobzol:
Rebased.

PR #1997, opened 2025-09-25
Rename community survey team to survey team

Kobzol:
And make launching-pad its parent.

As per https://github.com/rust-lang/leadership-council/issues/133.

### DIFF
--- a/repos/rust-lang/surveys-private.toml
+++ b/repos/rust-lang/surveys-private.toml
@@ -5,4 +5,4 @@ bots = []
 private-non-synced = true
 
 [access.teams]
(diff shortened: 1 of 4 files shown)


Kobzol:
CC @apiraino

apiraino:
yeah, like I commented I am mostly neutral about this change. I just see that we are moved away from being under the Community team (which imo was a fit for the survey subteam as well).

jieyouxu (approved):
Given that launching-pad has no parent team co-lead, I'm going to approve since both survey team co-leads are on board or at least neutral.

PR #1999, opened 2025-09-26
Add a few website descriptions to archived teams

Kobzol:
Companion to https://github.com/rust-lang/www.rust-lang.org/pull/2202, to provide a proper name and description to a few archived teams.


jieyouxu:
Does this need to land before https://github.com/rust-lang/www.rust-lang.org/pull/2202 or after? I guess it doesn't matter, because we'll have to render the archive teams in the first place, right...

PR #2000, opened 2025-09-27
rustc-dev-guide: make Noratrieb member

tshepang:
@Noratrieb regularly reviews PRs, so adding to the team as recognition of that fact

cc @BoxyUwU


jieyouxu (approved):
Nora is already a compiler member so

PR #2007, opened 2025-10-01
Prepare for lang FCPs in stdarch

traviscross:
Increasingly, due to what lives in stdarch, there's the likelihood of the lang team needing to do FCPs there.  Let's arrange for that.

cc @rust-lang/lang @sayantn https://github.com/rust-lang/rust/pull/146560


sayantn:
@traviscross so can I add `T-lang-nominated` to the stdarch pr now?

Issue #1476, opened 2024-06-10
Resolve individual repository access.


ehuss:
Part of [RFC 2872](https://rust-lang.github.io/rfcs/2872-github-access-policy.html) specifies that individuals should not be given permissions to repositories. We have a few individuals currently with access that we should resolve. They currently are:

- areweasyncyet.rs:
    - `upsuper`
    - Recommend: Create a team, figure out relationship with web-presence.
- backtrace-rs:
    - `alexcrichton`
    - Recommend: Remove Alex, he's been uninvolved for a few years.
- cargo-bisect-rustc:
    - `ehuss`
    - Recommend: Create a team, subteam of compiler (@ehuss assigned).
- cc-rs
    - `NobodyXu`
    - Recommend: Create a team, subteam of libs.
- `flate2-rs`
    - `alexcrichton`
    - `brson`
    - Recommend: Remove Alex (specified he has [handed off](https://github.com/rust-lang/flate2-rs/pull/325#issuecomment-1329548584)) and brson (hasn't been involved).
- google-summer-of-code
    - `Kobzol`
    - `jackh726`
    - Recommend: Create a team? Working-group? Where to slot it?
- log
    - `Thomasdezeeuw`
    - Recommend: Create a team, subteam of libs.
- mdbook
    - `Dylan-DPC`
    - Recommend: Create a team, subteam of devtools (@ehuss assigned).
- pkg-config-rs
    - `sdroege`
    - Recommend: Create a team, subteam of libs.
- project-deref-patterns
    - `traviscross`
    - Recommend: Switch to lang-ops.
- rustc_codegen_cranelift
    - `bjorn3`
    - Recommend: Create a team, subteam of compiler.
- socket2
    - `chansuke`
    - `sfackler`
    - `Thomasdezeeuw`
    - `carllerche`
    - `Darksonn`
    - Recommend: Create a team, subteam of libs.

Kobzol:
We already have a team for `gsoc-contributors`, we could create a team for `gsoc-mentors` (and org admins), and make them manage the gsoc repo.

PR #2030, opened 2025-10-22
Add @NobodyXu and @madsmtm to crate-maintainers team

madsmtm:
Motivation: To make Jiahao show up on the website.

This also removes the `[access.individuals]` key for `cc-rs` as part of https://github.com/rust-lang/team/issues/1476.

CC @rust-lang/crate-maintainers, this would add @NobodyXu and myself to your team. I would limit my use of these rights to contributions to `cc-rs` (which admittedly has been very limited lately), I suspect @NobodyXu would do the same.

An alternative would be to create a separate team `@rust-lang/cc`, I'd be fine with either.

### DIFF
--- a/repos/rust-lang/cc-rs.toml
+++ b/repos/rust-lang/cc-rs.toml
@@ -7,10 +7,6 @@ bots = []
 [access.teams]
 crate-maintainers = 'maintain'
 
-[access.individuals]
-NobodyXu = 'write'
-madsmtm = 'write'
-
 [[branch-protections]]
 pattern = 'main'
 required-approvals = 0
--- a/teams/crate-maintainers.toml
+++ b/teams/crate-maintainers.toml
@@ -16,7 +16,9 @@ members = [
     "lcnr",
     "tgross35",
     "jongiddy",
-    "Kobzol"
+    "Kobzol",
+    "NobodyXu",
+    "madsmtm",
 ]
 alumni = []
 


madsmtm:
>     📝 Editing repo 'rust-analyzer/smol_str':
>       Permission Changes:
>         Removing user 'alexheretic''s write permission 

That looks wrong? CC @alexheretic, it looks like you're not part of the `rust-analyzer` team?

NobodyXu:
Thanks Mads!

I'll also most likely contributing to cc-ras, very occasionally if jobserver has a bug impacting cc-rs or cc needs a feature from it, I might post a PR there and asks for others to review.

alexheretic:
> > ```
> > 📝 Editing repo 'rust-analyzer/smol_str':
> >   Permission Changes:
> >     Removing user 'alexheretic''s write permission 
> > ```
> 
> That looks wrong? CC @alexheretic, it looks like you're not part of the `rust-analyzer` team?

I've just started to collaborate on _smol_str_ crate, so to that extent maybe I should be?

tgross35 (approved):

ChrisDenton:
I think this is ultimately on @Amanieu (as libs team lead) to approve but I this should be fine since they both are already maintaining a libs crate. I don't think we need a special group just for cc-rs.

Amanieu (approved):

PR #2031, opened 2025-10-22
Document that updating dependencies is security sensitive

madsmtm:
I've been thinking a lot about sandboxing build scripts and proc-macros lately, and it occurred to me that if I were to attack this repo, I'd go the route of adding a `ctor` to some deeply nested dependency that sends the `GITHUB_TOKEN` somewhere if it detects it's inside this repo, and then just wait for Dependabot to update the dependency here.

CC @Kobzol, opening this to ensure that this is something you / t-infra is aware of.


Kobzol:
Good point. We made some mitigations against this, but it's not perfect.

PR #2035, opened 2025-10-24
Remove website entry for the alumni team

Kobzol:
All alumni are now automatically shown in https://rust-lang.org/governance/all-team-members.html and https://rust-lang.org/governance/archived-teams.html.

Marker teams with a website entry should be reserved for teams that we actually want to show on the website in the page of a specific toplevel team.

Unblocks https://github.com/rust-lang/www.rust-lang.org/pull/2213.

### DIFF
--- a/teams/alumni.toml
+++ b/teams/alumni.toml
@@ -10,8 +10,3 @@ kind = "marker-team"
 leads = []
 members = []
 include-all-alumni = true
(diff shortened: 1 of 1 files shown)


jieyouxu (approved):
Alternative display methods are already available, so this is fine.

PR #2041, opened 2025-10-27
FLS team, contributors updates

PLeVasseur:
cc @traviscross


PLeVasseur:
cc @AlexCeleste @kirtchev-adacore

PR #2044, opened 2025-10-29
repos/crates.io: Disable PR requirement

Turbo87:
This will allow https://github.com/rust-lang/crates.io/pull/12225 to work as intended. All team members with write permissions are aware anyway that we are supposed to use the PR workflow.


marcoieni (approved):

marcoieni:
```
validation error: repo 'crates.io' uses a branch protection for main that does not require a PR, but has non-empty `ci-checks`
```

Turbo87:
sigh... looks like that won't help anyway 😞

PR #2047, opened 2025-10-29
Mention `rust_team_data` usage when changing it

Urgau:
This PR adds a triagebot mention to changes done to the v1 format of `rust_team_data`, as it's being used by multiple projects and breaking the format results in an outage of some functionality.

See [#triagebot > ✔ assign handler, error decoding response body?](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/.E2.9C.94.20assign.20handler.2C.20error.20decoding.20response.20body.3F/with/547641597) for the last one https://github.com/rust-lang/team/pull/2043.

Note that "triagebot is currently not setup in this repo, as it would need write access (which needs to be tightly controlled)"[^1], which is fair for this repo. Fortunately posting a comment on this repo doesn't need write access so we should be fine.

cc @Kobzol @senekor (as you were also impacted)

[^1]: https://github.com/rust-lang/team/blob/master/docs/labels.md#labels-are-currently-manually-applied

### DIFF
--- a/docs/labels.md
+++ b/docs/labels.md
@@ -5,7 +5,7 @@ contributors to filter issues/PRs by labels.
 
 ## Labels are currently manually applied
 
-[`triagebot`][triagebot] is not currently set up for the `team` repository, as it would need write
+[`triagebot`][triagebot] is not currently set up for the `team` repository with write
 access (which needs to be tightly controlled). Thus, labels are currently manually applied and
 adjusted. See [threat model](./threat-model.md).
 


Kobzol:
Thanks, this is a good idea, and sorry for the breakage. Next time we should mark the field with `Option`, then update the projects, and remove it and then update the projects again.

jieyouxu:
Actually, could you also update https://github.com/rust-lang/team/blob/master/docs/labels.md#labels-are-currently-manually-applied to say that only limited functionality from triagebot is available because triagebot doesn't have write access to this repo? (Since technically now triagebot *is* setup, just that it can't support the "fancier" functionality).

jieyouxu:
CI failure is unrelated, it's because

> [ERROR rust_team::validate] validation error: GitHub user `Osintedx` changed username to `iownrena`

UPD: https://github.com/rust-lang/team/pull/2048 should unblock this PR once that merges.

PR #2058, opened 2025-11-04
Add `t-website` Zulip stream to t-website website entry

Kobzol:
Discussed with Remo.


teams/website.toml
```diff
@@ -20,3 +20,4 @@ orgs = ["rust-lang"]
 name = "Website team"
 description = "Manages the www.rust-lang.org website"
 repo = "https://github.com/rust-lang/www.rust-lang.org/"
+zulip-stream = "t-website"
```
jieyouxu, line 23: Remark: note that there is no corresponding `T-website` zulip group, but we could add one if the website team would like to be able to be pinged via `@T-website` on zulip

PR #2061, opened 2025-11-06
Update how often is the website updated

Kobzol:
It's no longer updated so often now that it is a static website.


jieyouxu (approved):
Thanks

PR #2062, opened 2025-11-06
Add validation for duplicate team entries

vivekjami:
This PR adds validation to check for duplicate entries in team members and alumni lists.

Fixes rust-lang/team #1918

Changes:
- Added new `validate_duplicate_team_entries` function
- Checks team members and alumni
- Uses separate function for cleaner code design as suggested by @Kivooeo

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -517,6 +517,10 @@ impl Team {
         self.people.alumni.as_ref().map_or(&[], Vec::as_slice)
     }
 
+    pub(crate) fn explicit_leads(&self) -> &[String] {
+        &self.people.leads
+    }
+
     pub(crate) fn contains_person(&self, data: &Data, person: &Person) -> Result<bool, Error> {
         let members = self.members(data)?;
         Ok(members.contains(person.github()))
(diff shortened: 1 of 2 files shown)


vivekjami:
@marcoieni

Kivooeo:
I tried open the link from https://github.com/rust-lang/team/pull/2032#issuecomment-3443513184

And it seems like it was disabled due to inactivity, so I updated link, so you can see https://github.com/rust-lang/team/compare/main...Kivooeo:team:add-check?expand=1, how I made this and read this comment above about some details, but I lean to this solution because it does account leads as well without repeating logic in check function

vivekjami:
Yes @Kivooeo , I saw your pr and i agree with your approach but i do not know if i can change the schema.rs as that was not specified in the pr and that is the cleanest way to handle team leads and I want to see the errors that needs to resolve in a single check so that i can solve the issue, so that is why i avoided bail!() , if i used that then your code works perfectly. @marcoieni any thoughts?

marcoieni:
you can change schema.rs, for sure!
Extracting the logic so that you don't repeat it, also sounds great!

vivekjami:
@marcoieni just updated the code with the new requirements

PR #2069, opened 2025-11-11
zulip api: unsubscribe doesn't take an object

davidtwco:
#2066 didn't have the intended effect. `users/me/subscriptions` subscriptions takes [a object for subscribing](https://zulip.com/api/subscribe#parameter-subscriptions) and [a string for unsubscribing](https://zulip.com/api/unsubscribe#parameter-subscriptions).

### DIFF
--- a/sync-team/src/zulip/api.rs
+++ b/sync-team/src/zulip/api.rs
@@ -241,9 +241,7 @@ impl ZulipApi {
         }
 
         if !remove_ids.is_empty() {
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Huh, another problem with this serialization 😆 I wonder if this changed sometime in the Zulip API. Thanks!

marcoieni (approved):

PR #2073, opened 2025-11-11
Clippy: Add Zulip streams to the team

flip1995:
Inspired by #2066

### DIFF
--- a/teams/clippy.toml
+++ b/teams/clippy.toml
@@ -48,3 +48,6 @@ address = "clippy@rust-lang.org"
 
 [[zulip-groups]]
 name = "T-clippy"
+
+[[zulip-streams]]
+name = "clippy/private"


teams/clippy.toml
```diff
@@ -48,3 +48,9 @@ address = "clippy@rust-lang.org"
 
 [[zulip-groups]]
 name = "T-clippy"
+
+[[zulip-streams]]
+name = "clippy"
+
+[[zulip-streams]]
+name = "clippy/private"
```
flip1995: Not sure if that is the correct way to specify 2 streams.

teams/clippy.toml
```diff
@@ -48,3 +48,10 @@ address = "clippy@rust-lang.org"
 
 [[zulip-groups]]
 name = "T-clippy"
+
+[[zulip-streams]]
+name = "clippy"
+extra-teams = ["clippy-contributors"]
```
flip1995: Do I understand correctly that this means that the `clippy-contributors` team will also be added to this channel?

teams/clippy.toml
flip1995: > \# This will remove anyone who isn't in the team from the stream
> \# so it should only be used for private streams at the moment.

Oh, so we should only use it for the private one.

teams/clippy.toml
Kobzol: Yeah. But note that it is currently only possible to create *private* channels with team, not public ones.

teams/clippy.toml
flip1995: Reverted.

teams/clippy.toml
flip1995: Yes, that's what it does. But no longer necessary.

flip1995: Thanks! I found the documentation and RTFM https://github.com/rust-lang/team/pull/2073#discussion_r2515145418 :D

jieyouxu:
Oh right, this is an existing private zulip stream, which needs the Rust Owner Account to be manually invited to the private stream first.

jieyouxu:
Discussion: [#clippy > clippy/private management through team repo](https://rust-lang.zulipchat.com/#narrow/channel/257328-clippy/topic/clippy.2Fprivate.20management.20through.20team.20repo/with/555035245)

Waiting on Rust Owner Account be manually invited before this can merge.

jieyouxu:
Update: Rust Owner Account was invited, let's give this another try.

PR #2078, opened 2025-11-13
Sync crates.io Trusted Publishing configs

Kobzol:
This PR adds support for synchronizing Trusted Publishing configurations for crates published from repositories managed by `team`.

Currently, it works like this:
- For each crate configured in `team`, it lists the existing configs.
- If it finds a matching config, nothing is created.
- All non-matching configs will be deleted.
- If a matching config was not found, a new config will be created.

Some additional notes:
- It does not yet do anything with crate owners, that can be done as a follow-up step, now that we have a crates.io API client set up.
- We will not be able to remove a publishing config if someone removes a crate config from the `team` DB, because the crates.io API does not currently have a way of listing trusted publishing configs for an API token, without knowing the corresponding crate name.
- At most one trusted publishing configuration can be configured per *crate* (not repo). This both simplifies the implementation, and in general is what we should aim to have, for added security. This is not a limitation of crates.io, but rather a self-imposed requirement of the `team` DB that I think makes sense.

I already configured this to run on CI, although we do not have the crates.io token configured for dry-run, that might be nice to add (but I don't think that crates.io allows creating read-only tokens, so we would have to do the dry-runs with a full token).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Install Rust Stable
         env:
-          RUST_VERSION: "1.85.0"
+          RUST_VERSION: "1.87.0"
         run: |
           rustc -vV
           rustup toolchain install $RUST_VERSION
@@ -103,6 +103,7 @@ jobs:
           EMAIL_ENCRYPTION_KEY: ${{ secrets.EMAIL_ENCRYPTION_KEY }}
           ZULIP_API_TOKEN: ${{ secrets.ZULIP_API_TOKEN }}
           ZULIP_USERNAME: ${{ secrets.ZULIP_USERNAME }}
+          CRATES_IO_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
         run: |
           cargo run sync apply --src build
 


Kobzol:
Actually, given that we might want to have a first-class support for crates in team DB (https://github.com/rust-lang/infra-team/issues/204#issuecomment-3524180344), I will refactor the schema slightly to make it more future-compatible.

Turbo87:
> * Only one trusted publishing configuration can be configured per _crate_ (not repo)

might be worth mentioning that this is a restriction on the `team` repo side. crates.io does support having multiple trustpub configs :)

src/validate.rs
```diff
@@ -1000,6 +1001,23 @@ Please remove the attributes when using bors"#,
...
+    let mut crates = HashMap::new();
+    wrapper(data.repos(), errors, |repo, _| {
+        let repo_name = format!("{}/{}", repo.org, repo.name);
+        for publishing in &repo.trusted_publishing {
+            if let Some(prev_crate) = crates.insert(&publishing.krate, repo_name.clone()) {
+                return Err(anyhow::anyhow!(
+                    "Repository `{repo_name}` configures trusted publishing for crate `{}` that is also configured in `{prev_crate}`. Each crate can only be configured once.",
+                    publishing.krate
+                ));
+            }
```
Turbo87:
```suggestion
            if let Some(prev_repo) = crates.insert(&publishing.krate, repo_name.clone()) {
                return Err(anyhow::anyhow!(
                    "Repository `{repo_name}` configures trusted publishing for crate `{}` that is also configured in `{prev_repo}`. Each crate can only be configured once.",
                    publishing.krate
                ));
            }
```

docs/toml-schema.md
```diff
@@ -430,3 +430,16 @@ allowed-merge-teams = ["awesome-team"]
 # (optional)
 merge-bots = ["homu"]
 ```
+
+### Crates.io trusted publishing
+Configure crates.io Trusted Publishing for crates published from a given repository from GitHub Actions.
+
+```toml
+[[trusted-publishing]]
```
emilyalbini: Nit: I'd call it `crates-io-publishing`, if in the future we need to integrate with other ecosystems.

sync-team/src/crates_io/api.rs
```diff
@@ -0,0 +1,175 @@
...
+        path: &str,
+        data: Option<&T>,
+    ) -> anyhow::Result<reqwest::blocking::Response> {
+        let mut req = self
+            .client
+            .request(method, format!("{CRATES_IO_BASE_URL}{path}"))
+            .header(
+                "Authorization",
+                format!("Bearer {}", self.token.expose_secret()),
+            );
```
Turbo87:
```suggestion
            .bearer_auth(self.token.expose_secret());
```

I think this should be sufficient, but I might be wrong 😅

docs/toml-schema.md
```diff
@@ -430,3 +430,16 @@ allowed-merge-teams = ["awesome-team"]
...
+# Name of the crate that will be published from this repository (required)
+crate = "regex"
+# Name of a GitHub Actions workflow file that will publish the crate (required)
+workflow-filename = "ci.yml"
+# GitHub Actions environment that has to be used for the publishing (optional)
+environment = "deploy"
```
emilyalbini, line 444: I would personally enforce on the sync-team side that an environment is always configured when publishing. I don't see any case when we would want all workflows on a repository to publish crates.

emilyalbini, line 444: Only people with maintain+ access on a repo can create environments. We'd need a way to let people create environments without infra-admins involvement. Also, migrating all environments to be managed by the team repo might not be trivial, as a lot of environments nowadays are created by Terraform.

pietroalbini (commented):
Code looks good, left some general feedback.

sync-team/src/crates_io/api.rs
Kobzol: Yeah, that's better. Thanks!

docs/toml-schema.md
Kobzol: This might make it a bit harder to deploy TP though, as creating an environment will require an infra-admin for each repo? Or are they auto-created if you just specify them in the YML workflow file? :thinking:

Kobzol: Ah, I see, this answers my question above. So what would you prefer? Make environments optional for now? Or make them required and involve infra admins in their creation?

Mark-Simulacrum:
> Only one trusted publishing configuration can be configured per crate (not repo). 

Is this zero or one? It would be nice to have a good way to remove those configs (e.g., if we rename a repo or move it out of rust-lang org), otherwise I worry we'd forget and leave behind a possible future risk.

Or does crates.io attach them to some GitHub repo id despite the naming convention? I don't think that's a thing....

Kobzol:
It should have said "at most one", I clarified it in the PR description.

Before the endpoint for enumerating TP configs is implemented, I don't think that we can reliably remove all stale configs, right?

The TP configs are attached to a repository name (org/name).

docs/toml-schema.md
emilyalbini: They are not auto-created :(

emilyalbini: I think it would be better now to require them, and eventually integrate them in the team repo.

Kobzol:
Changed to make environment mandatory and expand the set of TPs into a set of crates in the v1 API.

docs/toml-schema.md
ehuss: What do environments do in this case? Why would they be required?

Kobzol: Environments can be used to ensure that the publish happens only from a specific branch or workflow job. That is useful to ensure that unexpected parts of CI (e.g. those that execute untrusted code at build time) are not allowed to publish.

sync-team/src/crates_io/api.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Create a new trusted publishing configuration for a given crate.
+    pub(crate) fn create_trusted_publishing_github_config(
+        &self,
+        config: &CratesIoPublishingConfig,
+    ) -> anyhow::Result<()> {
+        debug!(
+            "Creating trusted publishing config for '{}' in repo '{}/{}', workflow file '{}' and environment '{}'",
+            config.krate.0,
+            config.repo_name,
```
marcoieni:
```suggestion
            config.repo_org,
```

Kobzol: Thanks, fixed!

sync-team/src/crates_io/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    pub(crate) fn diff_all(&self) -> anyhow::Result<Diff> {
+        let mut config_diffs: Vec<ConfigDiff> = vec![];
+
+        // Note: we currently only support one trusted publishing configuration per crate
+        for (krate, desired) in &self.crates {
+            let mut configs = self
+                .crates_io_api
+                .list_trusted_publishing_github_configs(&krate.0)?;
```
marcoieni:
```suggestion
                .list_trusted_publishing_github_configs(&krate.0)
                .with_context(|| format!("Failed to list configs for crate '{}'", krate.0))?;
```

sync-team/src/crates_io/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+                        && *workflow_filename == desired.workflow_file
+                        && environment.as_deref() == Some(&desired.environment)
+                })
+                .collect::<Vec<_>>();
+
+            if !matching_configs.is_empty() {
+                // If we found a matching config, we don't need to do anything with it
+                // It shouldn't be possible to have multiple configs with the same repo, workflow
+                // and environment for a single crate.
+                assert_eq!(matching_configs.len(), 1);
```
marcoieni, line 97: I wonder if returning an error makes more sense than asserting

Kobzol: If the assert fails, then our assumptions (that there can't be more than one config with the same publish file and environment per repo) do not hold. In that case we should update the code.

Kobzol:
Applied suggestion and cleaned up the history :+1:

marcoieni (approved):

PR #2077, opened 2025-11-13
lcnr: add github-sponsors

lcnr:
"testing" #2071

r? @Kobzol

PR #2079, opened 2025-11-13
waffle: add github-sponsors, fixup name

WaffleLapkin:
Since @Kobzol [wants some test dummies](https://github.com/rust-lang/team/pull/2077#issuecomment-3526659033) :)

### DIFF
--- a/people/WaffleLapkin.toml
+++ b/people/WaffleLapkin.toml
@@ -4,3 +4,6 @@ github-id = 38225716
 email = 'waffle.lapkin@gmail.com'
 zulip-id = 273349
 discord-id = 123482303698567168
+
+[funding]
+github-sponsors = true


WaffleLapkin:
My github sponsors profile also have links to some other cool rust people who have sponsors enabled, but I'm not sure if they would like to be advertised or not..

Kobzol:
> My github sponsors profile also have links to some other cool rust people who have sponsors enabled, but I'm not sure if they would like to be advertised or not..

I think that's fine, it's transitive, and not something that would be automatically on the Rust website. You can probably also find links to Sponsors just from GH profiles that we link to from the website already.

WaffleLapkin:
To be clear I meant it in "maybe worth asking them is they want to be included in this", not "it's a reason to not include me" :p

Kobzol:
Yeah I definitely plan to do a wider sweep asking people to add their Sponsors! I first wanted to actually have something to show for it though (some integration with the Rust website), to motivate people to do it. So far I have you beta testers :P

jieyouxu (approved):

PR #2082, opened 2025-11-14
Add triagebot to glob

Kobzol:
It is supposed to be used there (https://github.com/rust-lang/glob/blob/master/triagebot.toml).


tgross35:
Fwiw I think that file could probably just be deleted, we don't really have enough dedicated reviewers to do assignments. But it doesn't hurt to have.

Kobzol:
I wanted to use it to make sure that PRs don't fall through the cracks. I think that "not having enough dedicated reviewers" is something that can be *ever so slightly* improved by making triagebot work there :)

PR #2084, opened 2025-11-17
Create a private channel that contains all team members

Kobzol:
So that we can communicate with them all at once. This should be a more reliable alternative to the `all@rlo.org` mailing list.

### DIFF
--- a/teams/all.toml
+++ b/teams/all.toml
@@ -45,7 +45,9 @@ excluded-people = [
 
 # Private channel with all team members, so that we can have an easy way of reaching them.
 [[zulip-streams]]
-name = "rust project"
+name = "all/private"
+# Exclude the following people from the Zulip stream for grandfathering purposes,
+# where previously we didn't require all project team members to have Zulip IDs.
 excluded-people = [
     "U007D",
     "andrewpollack",
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -758,16 +758,20 @@ fn validate_zulip_stream_ids(data: &Data, errors: &mut Vec<String>) {
             return Ok(());
         }
 
-        wrapper(team.members(data)?.iter(), errors, |member, _| {
-            if let Some(member) = data.person(member) {
-                if member.zulip_id().is_none() {
-                    bail!(
-                        "person `{}` in '{}' is a member of a Zulip stream but has no Zulip id",
-                        member.github(),
-                        team.name()
-                    );
+        wrapper(streams.iter(), errors, |stream, errors| {
+            wrapper(stream.members().iter(), errors, |member, _| {
+                match member {
+                    ZulipMember::MemberWithId { .. } | ZulipMember::JustId(_) => {}
+                    ZulipMember::MemberWithoutId { github } => {
+                        bail!(
+                            "person `{github}` is a member of a Zulip stream `{}` defined in team `{}`, but has no Zulip id",
+                            stream.name(),
+                            team.name()
+                        );
+                    }
                 }
-            }
+                Ok(())
+            });
             Ok(())
         });
         Ok(())


marcoieni (approved):

marcoieni:
looks good from an infra admin perspective. probably needs a team-repo-admin review, too

teams/all.toml
```diff
@@ -42,3 +42,15 @@ excluded-people = [
     "celaus",
     "opeolluwa",
 ]
+
+# Private channel with all team members, so that we can have an easy way of reaching them.
+[[zulip-streams]]
+name = "rust project"
```
jackh726: I would make this `all/private` (which matches the naming of `all@rust-lang.org`. Then you could have an `all` (if it seems helpful) in the future.

teams/all.toml
```diff
@@ -42,3 +42,15 @@ excluded-people = [
...
+excluded-people = [
+    "U007D",
+    "andrewpollack",
+    "arshiamufti",
+    "bnchi",
+    "celaus",
+    "opeolluwa",
```
jackh726, line 57: Q: why?

jieyouxu: (These members do not have a corresponding Zulip ID)

teams/all.toml
Kobzol: Fair enough. This channel does not necessarily be private for all information shared there, but we don't have an easy way of syncing public channels at the moment.

Kobzol:
Clarified why people were excluded and renamed the stream to `all/private`.

teams/all.toml
```diff
@@ -42,3 +42,17 @@ excluded-people = [
     "celaus",
     "opeolluwa",
 ]
+
+# Private channel with all team members, so that we can have an easy way of reaching them.
+[[zulip-streams]]
+name = "all/private"
```
tshepang, line 48: would be kool if this was named similar to [the private one for leads](https://github.com/rust-lang/team/blob/797965e4abf5cdba42cb4332ce60b1460e0912a1/teams/leads.toml#L17)

so we would have
- `project members (private)`
- `project leads (private)`

with this pr, we would have
- `all/private`
- `project leads (private)`

Kobzol: There are already a bunch of `XXX/private` channels, so the project leads one is an exception, I'd say, which was chosen because the channel already existed.

tshepang: oh, in that case (offtopic for this pr here), could it be renamed to `leads/private` to match others

Kobzol: I don't think we can automatically rename the channel through team, it would probably just create a new one. We'd have to do it by hand, but also sync it with team, so that it doesn't recreate a channel with the old name. So it's not trivial to rename.

Kobzol:
The funding page is now live (https://rust-lang.org/funding/), so it would be cool to have this channel so that I can tell Project members to add their GH sponsors to `team` :) (btw I don't think this needs infra-admin review?).

jackh726:
Obviously I'm a big plus on this, but want to make sure that this has at least a general nod of approval from the council.

Kobzol:
Asked [here](https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/t-all.2Fprivate.20channel/with/557962302) + we can discuss on Friday.

Kobzol:
We discussed it at today's Council meeting. No one seemed opposed to the idea, and I got some thumbs up. So I think that we can move forward here.

jackh726:
I went ahead and created the stream on zulip.

jackh726:
huh...not sure what happened, I messed up the name and permissions of the channel on zulip? Fixed, trying again

PR #2075, opened 2025-11-12
Enable syncing for Zulip #t-rustdoc/private

fmease:
Inspired by https://github.com/rust-lang/team/pull/2074, https://github.com/rust-lang/team/pull/2073 and https://github.com/rust-lang/team/pull/2066.

PR #2086, opened 2025-11-18
Enable syncing for Zulip `#t-docs-rs/private`

GuillaumeGomez:
So based on https://github.com/rust-lang/team/pull/2075, we need to invite rust-lang-owner to the stream first.

Since we also want to have the infra team to have access to this channel, I added them as well.

cc @syphar

### DIFF
--- a/teams/docs-rs.toml
+++ b/teams/docs-rs.toml
@@ -44,3 +44,12 @@ address = "help@docs.rs"
 
 [[zulip-groups]]
 name = "T-docs-rs"
+
+[[zulip-streams]]
+name = "t-docs-rs/private"
+extra-teams = [
+    "infra",
+]
+extra-people = [
+    "walterhpearce",
+]


teams/docs-rs.toml
```diff
@@ -44,3 +44,6 @@ address = "help@docs.rs"
 
 [[zulip-groups]]
 name = "T-docs-rs"
+
+[[zulip-streams]]
```
Kobzol, line 48: It works a bit differently. The channel should only be included in one file (here), but it should have:
```toml
extra-teams = [
    "infra",
]
```

GuillaumeGomez:
Seems like we can't have two teams invited into a private stream?

EDIT: `kobzol` provided the answer. :)

teams/docs-rs.toml
GuillaumeGomez: Oh I see, thanks!

marcoieni:
I saw @walterhpearce is in the stream, too. So this PR will remove him probably.

marcoieni:
we could add him with the `extra-people` array:
```
extra-people = [
    "walterhpearce",
]
```

GuillaumeGomez:
Who are they?

marcoieni:
Walter works at the rust foundation as a security engineer

GuillaumeGomez:
Oh I see. Why aren't they part of the infra team? Would make things simpler. :)

marcoieni:
Less work for me is always welcome 😁
Jokes aside, he can answer for himself when he is online, but I think his work is cross team, i.e. from crates-io vulnerabilities to compiler vulnerabilities, so he could be in almost every team in theory 😂

My hope is that the project creates a dedicated security team one day :)

GuillaumeGomez:
That would be convenient! Well, adding them in the meantime.

marcoieni:
Here is the diff:
```
    📝 Updating stream membership:
      Name: t-docs-rs/private
      ID: 356854
      Members:
        ➕ 259697
        − 125254
        − 210267
        − 310345
        − 452253
```

so consider adding back some of these zulip accounts if you think they should be there!

GuillaumeGomez:
Most of them were former members of the docs.rs team (sadness :crying_cat_face:), so I think it's as expected. They can still ask us to be brought back on the channel if needed.

PR #2101, opened 2025-11-21
Move sec-response to a regular subteam of launching pad

jamesmunns:
As per https://github.com/rust-lang/leadership-council/issues/141, we are retaining security response in the launching pad as a cross cutting team.

CC @cuviper, @pietroalbini, @Manishearth

You may also want to rename your zulip group and working repository, either as part of this PR or as a follow-up.

### DIFF
--- a/teams/security-response.toml
+++ b/teams/security-response.toml
@@ -19,7 +19,7 @@ team-name = "security"
 orgs = ["rust-lang"]
 
 [website]
-name = "Security Response Subteam"
+name = "Security Response Team"
 description = "Triaging and responding to incoming vulnerability reports"
 email = "security@rust-lang.org"
 repo = "https://github.com/rust-lang/wg-security-response"


teams/security-response.toml
```diff
@@ -20,7 +19,7 @@ team-name = "security"
 orgs = ["rust-lang"]
 
 [website]
-name = "Security Response WG"
+name = "Security Response Subteam"
```
emilyalbini: Is there a reason why this isn't Security Response Team?

jamesmunns: Fixed

teams/security-response.toml
```diff
@@ -20,7 +19,7 @@ team-name = "security"
 orgs = ["rust-lang"]
 
 [website]
-name = "Security Response WG"
+name = "Security Response Team"
 description = "Triaging and responding to incoming vulnerability reports"
 email = "security@rust-lang.org"
 repo = "https://github.com/rust-lang/wg-security-response"
```
cuviper, line 25: We should rename the repo too, but that's an infra task.

PR #2102, opened 2025-11-21
Make validation error messages more verbose

Kobzol:
The inner context is not actually printed by sync-team, so it was swallowed. Found in https://github.com/rust-lang/team/actions/runs/19570968459/job/56044145703?pr=2101.

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -1131,7 +1131,7 @@ where
 {
     for item in iter {
         if let Err(err) = func(item, errors) {
-            errors.push(err.to_string());
+            errors.push(format!("{err:?}"));
         }
     }
 }


marcoieni:
Imo the actual fix is to change this line:
https://github.com/rust-lang/team/blob/a3e926d8db55a6daa68ecb31ae27e9a705e96176/src/validate.rs#L127

to `error!("validation error: {err:?}");`

so that the error context is printed

`:?` should be used everywhere we print anyhow errors probably

Kobzol:
Well, we use anyhow errors, so the Debug output will include all kind of crap, possibly even backtraces. I kind of like the single-line error messages that we have right now. Maybe the verbose output could be optional with `-v` or something.

marcoieni:
[this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024) is how the error would look like.

It doesn't contain crap, but the context of all the errors. 

Why wouldn't we want to print the error context?

Kobzol:
Ok, I tried it and it's better than I thought :) I don't understand why there's no backtrace, anyhow should normally capture it and print it. Anyway, I pushed an alternative implementation based on what you suggested, this is how it would look like:
```
[ERROR rust_team::validate] validation error: Access for the repo rust-lang/wg-security-response is invalid
    
    Caused by:
        The file 'teams/foo-security-response.toml' does not exist.
```

marcoieni (approved):

marcoieni:
> I don't understand why there's no backtrace, anyhow should normally capture it and print it.

mmmh maybe you need to enable it as a feature? Never saw backtraces in anyhow

Kobzol:
No it's enabled by default. I just forgot that `RUST_BACKTRACE` needs to be set to see it :) Nevermind.

PR #2115, opened 2025-11-24
Fix script for finding publishable package

Kobzol:
When a crate sets `publish = false`, `cargo metadata` seems to represent it with `[]`, rather than `false`.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -446,4 +446,4 @@ environment = "deploy"
 
 > [!TIP]
 > To list the public crates of a workspace, run:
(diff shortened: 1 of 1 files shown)


weihanglo:
this sounds like a Cargo bug to me

Kobzol:
https://github.com/kobzol/cargo/blob/1921d097ca06015865096fea37d8fdb70296b4ff/src/cargo/core/package.rs#L84 it's not a bug, I think, because the `publish` field can contain only specified registries where publishing is enabled. So an empty array kind of makes sense.

weihanglo:
Yeah but in TOML an empty array means allowing any. I hate this kind of confusion 😞

PR #2126, opened 2025-11-26
feat: add `[[environments]]` in the repos directory

### LINKED ISSUE #2037: Provision environments from this repo
GitHub environments are currently setup manually by github admins.
It would be nice to set them up in this repo.

We should have a new array `[[environments]]` (similar to `[[branch-protections]]`). 

As a first step, the only field of an environment should be `name`. The goal should be to import all existing environments into the team repo. So when we merge the PR, the dry run should be empty.

> [!NOTE]
> The dry run only runs on PRs pushed directly to this repo, so once the contributor opens the PR, an infra admin (like me) can resubmit the PR to check the dry run output.

EDIT: branches and tags support was added in https://github.com/rust-lang/team/issues/2132

amustaque97:
Closes #2037

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -429,6 +429,8 @@ allowed-merge-teams = ["awesome-team"]
 # When "homu" is used, "bors" has to be in the `bots` array.
 # (optional)
 merge-bots = ["homu"]
+# Supported environments
+[[environments]]
 ```
 
 ### Crates.io trusted publishing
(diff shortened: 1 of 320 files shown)


marcoieni:
Why you added environments with empty names everywhere?
we should only add the environments block to repositories that have them configured.

So I was expecting:
1. change rust code to implement logic to detect drift in environments and to add/delete them
2. I run the team code to detect where environments are missing
3. We add the environments block where necessary
4. We verify that team doesn't sync anything (dry run is clean) and we merge the PR

amustaque97:
Closing this PR, and I will raise a clean & proper PR!

PR #2127, opened 2025-11-26
feat: add support for `[[environments]]` block

### LINKED ISSUE #2037: Provision environments from this repo

amustaque97:
closes #2037

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -431,6 +431,20 @@ allowed-merge-teams = ["awesome-team"]
 merge-bots = ["homu"]
 ```
 
+### Repository environments
+
+GitHub environments are used to configure deployment protection rules and secrets for GitHub Actions workflows. This repository can manage environment names for repositories.
+
+```toml
+# The environments in this repository (optional)
+[[environments]]
+# The name of the environment (required)
+name = "production"
+
+[[environments]]
+name = "staging"
+```
+
 ### Crates.io trusted publishing
 Configure crates.io Trusted Publishing for crates published from a given repository from GitHub Actions.
 
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -56,6 +56,7 @@ static CHECKS: &[Check<fn(&Data, &mut Vec<String>)>] = checks![
     validate_repos,
     validate_archived_repos,
     validate_branch_protections,
+    validate_environments,
     validate_trusted_publishing,
     validate_member_roles,
     validate_admin_access,
@@ -974,6 +975,48 @@ fn validate_archived_repos(data: &Data, errors: &mut Vec<String>) {
     wrapper(data.archived_repos(), errors, |repo, _| {
         if !repo.access.teams.is_empty() {
             bail!("archived repo '{}' should not have any teams", repo.name);
+        }
+        Ok(())
+    });
+}
+
+/// Validate that environments have valid names (non-empty)
+fn validate_environments(data: &Data, errors: &mut Vec<String>) {
+    wrapper(data.all_repos(), errors, |repo, _| {
+        for env in &repo.environments {
+            if env.name.is_empty() {
+                bail!(
+                    "repo {}/{} has an environment with an empty name",
+                    repo.org,
+                    repo.name
+                );
+            }
+            // Environment names should not contain special characters that could cause issues
+            // with GitHub API URLs or shell commands. Forward and backward slashes are particularly
+            // problematic as they can be interpreted as path separators
+            //in URLs (/repos/{owner}/{repo}/environments/{name}) and file systems,
+            //potentially leading to security vulnerabilities or API routing issues.
+            if env.name.contains('/') || env.name.contains('\\') {
+                bail!(
+                    "repo {}/{} has an environment '{}' with invalid characters (/, \\)",
+                    repo.org,
+                    repo.name,
+                    env.name
+                );
+            }
+        }
+
+        // Check for duplicate environment names
+        let mut seen = HashSet::new();
+        for env in &repo.environments {
+            if !seen.insert(&env.name) {
+                bail!(
+                    "repo {}/{} has duplicate environment '{}'",
+                    repo.org,
+                    repo.name,
+                    env.name
+                );
+            }
         }
         Ok(())
     });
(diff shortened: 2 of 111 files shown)


docs/toml-schema.md
```diff
@@ -431,6 +431,23 @@ allowed-merge-teams = ["awesome-team"]
...
+# The name of the environment (required)
+name = "production"
+
+[[environments]]
+name = "staging"
+```
+
+> [!NOTE]
+> Currently, only environment names are managed. Environment protection rules and secrets are not synchronized.
+
```
```suggestion
```
people would figure this out from the schema. Also we can't synchronize secrets because we can't write them in clean text in git.

sync-team/src/github/api/mod.rs
```diff
@@ -333,6 +333,11 @@ pub(crate) struct Repo {
     pub(crate) allow_auto_merge: Option<bool>,
 }
 
+#[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
+pub(crate) struct Environment {
+    pub(crate) name: String,
+}
+
```
marcoieni: this struct is unused. CI doesn't detect it because we are using an old rust version (I will update it)

src/validate.rs
```diff
@@ -926,6 +927,44 @@ fn validate_archived_repos(data: &Data, errors: &mut Vec<String>) {
...
+        for env in &repo.environments {
+            if env.name.is_empty() {
+                bail!(
+                    "repo {}/{} has an environment with an empty name",
+                    repo.org,
+                    repo.name
+                );
+            }
+            // Environment names should not contain special characters that could cause issues
+            if env.name.contains('/') || env.name.contains('\\') {
```
marcoieni, line 999: github allows to create environments with these characters. Why shouldn't we?

sync-team/src/github/api/read.rs
```diff
@@ -50,6 +50,10 @@ pub(crate) trait GithubRead {
         org: &str,
         repo: &str,
     ) -> anyhow::Result<HashMap<String, (String, BranchProtection)>>;
+
+    /// Get environments for a repository
+    /// Returns a list of environment names
+    fn repo_environments(&self, org: &str, repo: &str) -> anyhow::Result<Vec<String>>;
```
marcoieni: Imo this should return a vector of Environment. so that we can add more fields later.

sync-team/src/github/tests/test_utils.rs
```diff
@@ -601,6 +628,8 @@ struct GithubOrg {
     repo_members: HashMap<String, RepoMembers>,
     // Repo name -> Vec<(protection ID, branch protection)>
     branch_protections: HashMap<String, Vec<(String, BranchProtection)>>,
+    // Repo name -> Vec<environment name>
+    repo_environments: HashMap<String, Vec<String>>,
```
marcoieni: same here, we should have an Environment struct, so that we can add fields later

sync-team/src/github/tests/test_utils.rs
```diff
@@ -584,6 +602,15 @@ impl GithubRead for GithubMock {
 
         Ok(result)
     }
+
+    fn repo_environments(&self, org: &str, repo: &str) -> anyhow::Result<Vec<String>> {
```
marcoieni: same here, we should have an Environment struct, so that we can add fields later

marcoieni (commented):
Looks good overall. I left a few minor comments. After addressing them I will test this :)

src/validate.rs
amustaque97: Because with GitHub API URLs or shell commands. Forward and backward slashes are particularly problematic as they can be interpreted as path separators in URLs (/repos/{owner}/{repo}/environments/{name}) and file systems, potentially leading to security vulnerabilities or API routing issues. I hope this makes sense.

amustaque97:
Addressed all the review comments. Thanks for the review!
Ran a basic sanity check, and it is working locally. 
Screenshot 2025-11-27 at 6 06 42 PM

src/validate.rs
marcoieni: ok, please write this as a comment 👍

marcoieni:
The dry run worked 🥳

Can you add these environments to the toml files?

https://github.com/rust-lang/team/pull/2129#issuecomment-3586204600

Also you resolved [this](https://github.com/rust-lang/team/pull/2127/files#r2565964285) comment but you didn't address it (you should delete that note imo).

Later, you can cleanup the git history (e.g. squash all commits) and this should be ready to be merged.

sync-team/src/github/tests/test_utils.rs
```diff
@@ -601,6 +627,8 @@ struct GithubOrg {
     repo_members: HashMap<String, RepoMembers>,
     // Repo name -> Vec<(protection ID, branch protection)>
     branch_protections: HashMap<String, Vec<(String, BranchProtection)>>,
+    // Repo name -> Vec<environment name>
```
marcoieni:
```suggestion
    // Repo name -> Vec<environment>
```

marcoieni:
Please add the comment to resolve https://github.com/rust-lang/team/pull/2127#discussion_r2568147992

Also, I noticed that in the commit history there's a merge commit. Can you get rid of that by rebasing? 🙏

src/validate.rs
amustaque97: This is done.

marcoieni:
The [dry run](https://github.com/rust-lang/team/pull/2129#issuecomment-3586204600) looks empty, which means that if we merge this there will be nothing applied. Which is what we want because it means we added all existing environments 🥳
merging this

marcoieni:
I created https://github.com/rust-lang/team/issues/2132 as a follow up of this task if you are interested 😁

PR #2129, opened 2025-11-27
[test] testing the dry run of #2127

marcoieni:
PR open to test dry run of #2127


marcoieni:
GitHub UI shows this PR as merged but in reality it wasn't (luckily).
Probably it shows as merged because it has the same commits as #2127

PR #2134, opened 2025-11-29
Allow configuring "trusted publishing only" config for crates

Kobzol:
Discussed in https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Trusted.20publishing.20through.20team/with/560399101.

### DIFF
--- a/sync-team/src/crates_io/api.rs
+++ b/sync-team/src/crates_io/api.rs
@@ -1,4 +1,4 @@
-use crate::crates_io::CratesIoPublishingConfig;
+use crate::crates_io::CrateConfig;
 use crate::utils::ResponseExt;
 use anyhow::{Context, anyhow};
 use log::debug;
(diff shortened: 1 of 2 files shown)


docs/toml-schema.md
```diff
@@ -456,6 +456,9 @@ crates = ["regex"]
 workflow-filename = "ci.yml"
 # GitHub Actions environment that has to be used for the publishing (required)
 environment = "deploy"
+# Disable other mechanisms for publishing this set of crates (optional, default is false)
+# If set to `true`, the crates will only be publishable through trusted publishing
+disable-other-publish-methods = false
```
marcoieni, line 461: I think we should default to true, since it's the most secure option

Kobzol: Sure, but first I wanted to get a clean dry-run, I assume that most repos won't have it enabled right now, as it's brand new.

marcoieni:
The dry run is clean: https://github.com/rust-lang/team/pull/2138#issuecomment-3598724395

Issue #2132, opened 2025-11-28
Allow configuring deployment branches in environments


marcoieni:
https://github.com/rust-lang/team/pull/2127 added support for declaring environments.
We now want to allow people to declare the deployment branches of their environments.

TOML:
```toml
[[environments]]
name = "production"

[[environments.branches]]
name = "main"
```

I used the syntax `[[environments.branches]]` because there could be more branches per environment. Let me know if I'm missing something or you have a better syntax in mind.

UI screenshot:

<img width="1634" height="1468" alt="Image" src="https://github.com/user-attachments/assets/205c5906-5198-4989-bec2-364fb7d0a49f" />

amustaque97:
My suggestion would be to follow the format
```toml
[environments.development]
branch = ["dev_db"]
tag = ["v1.0-dev"]

[environments.production]
branch = ["prod_db"]
tag = ["v1.0"]

```

This TOML structure is better because it is:
1. Simpler to read and maintain
2. More predictable for the code that loads it
3. Expandable without breaking changes
4. Cleaner than arrays of tables
5. Accurate for real-world use (one config per environment)

_Please feel free to add if I'm missing anything_ 🙂 


<img width="784" height="520" alt="Image" src="https://github.com/user-attachments/assets/d813a4dc-a98a-4a26-91f1-793dde37934e" />

Kobzol:
Yeah we need a way to associate the branches to the environment. So either
```toml
[environments.foo]
branches = []
```

or
```toml
[[environments]]
name = "foo"
branches = []
```
the first would allow us to remove the validation for unique environment names.

marcoieni:
My idea was wrong. You are both right! I like the style of the second proposal from Jakub since it's simpler 👍

We don't need tags at the moment I think

Kobzol:
> since it's simpler

`[environments.foo]` should be a bit easier to both implement and consume, because you implicitly get the environments as a map where the keys have to be unique, rather than just a list of environments, so we wouldn't have to validate that there aren't two environments with the same name.

But it doesn't really matter much, both will do the job.

marcoieni:
Both works for me! I'm not sure which style we adopt for the rest of the config files

marcoieni:
New dry run in https://github.com/rust-lang/team/pull/2159#issuecomment-3617834360:

```
    📝 Editing repo 'rust-lang/ar_archive_writer':
      Environments:
        🔄 Update: publish
            Adding branches: *
            Removing tags: *
[...]

PR #2142, opened 2025-12-02
feat: add deployment branch support

amustaque97:
Closes: #2132
Extends environment configuration to support specifying deployment branch restrictions, allowing teams to control which branches can deploy to each environment via a new field in the TOML schema.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -433,16 +433,25 @@ merge-bots = ["homu"]
 
 ### Repository environments
 
-GitHub environments are used to configure deployment protection rules and secrets for GitHub Actions workflows. This repository can manage environment names for repositories.
+GitHub environments are used to configure deployment protection rules and secrets for GitHub Actions workflows. This repository can manage environment names and deployment branch and tag policies for repositories.
 
 ```toml
 # The environments in this repository (optional)
-[[environments]]
-# The name of the environment (required)
-name = "production"
+# Use table-of-tables syntax where the key is the environment name
+[environments.production]
+# List of branch patterns that can deploy to this environment (optional)
+# If empty or omitted, any branch can deploy
+branches = ["main", "release/*"]
+# List of tag patterns that can deploy to this environment (optional)
+tags = ["v*", "release-*"]
 
-[[environments]]
-name = "staging"
+[environments.staging]
+# Only specific branches can deploy to staging
+branches = ["develop", "staging"]
+# No tag patterns specified - no tags can deploy
+
+[environments.development]
+# No branch or tag patterns specified - any branch or tag can deploy
 ```
 
 ### Crates.io crate management
--- a/repos/rust-lang/edition-guide.toml
+++ b/repos/rust-lang/edition-guide.toml
@@ -12,6 +12,5 @@ pattern = "master"
 ci-checks = ["Success gate"]
 required-approvals = 0
 
-[[environments]]
-name = "github-pages"
+[environments.github-pages]
 
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -175,7 +175,7 @@ pub struct Repo {
     pub members: Vec<RepoMember>,
     pub branch_protections: Vec<BranchProtection>,
     pub crates: Vec<Crate>,
-    pub environments: Vec<Environment>,
+    pub environments: IndexMap<String, Environment>,
     pub archived: bool,
     // This attribute is not synced by sync-team.
     pub private: bool,
@@ -269,7 +269,10 @@ pub struct CratesIoPublishing {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Environment {
-    pub name: String,
+    #[serde(default)]
+    pub branches: Vec<String>,
+    #[serde(default)]
+    pub tags: Vec<String>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -3,7 +3,7 @@ pub(crate) use crate::permissions::Permissions;
 use anyhow::{bail, format_err, Error};
 use serde::de::{Deserialize, Deserializer};
 use serde_untagged::UntaggedEnumVisitor;
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 
 #[derive(serde_derive::Deserialize, Debug)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
@@ -814,7 +814,7 @@ pub(crate) struct Repo {
     #[serde(default)]
     pub crates_io: Vec<CratesIoConfiguration>,
     #[serde(default)]
-    pub environments: Vec<Environment>,
+    pub environments: BTreeMap<String, Environment>,
 }
 
 #[derive(serde_derive::Deserialize, Debug, Clone, PartialEq)]
@@ -891,5 +891,10 @@ pub(crate) struct CratesIoConfiguration {
 #[derive(serde_derive::Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
 pub(crate) struct Environment {
-    pub name: String,
+    /// Branch patterns that can deploy to this environment
+    #[serde(default)]
+    pub branches: Vec<String>,
+    /// Tag patterns that can deploy to this environment
+    #[serde(default)]
+    pub tags: Vec<String>,
 }
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -1044,11 +1044,11 @@ fn validate_archived_repos(data: &Data, errors: &mut Vec<String>) {
     });
 }
 
-/// Validate that environments have valid names (non-empty)
+/// Validate that environments have valid names (non-empty) and deployment patterns
 fn validate_environments(data: &Data, errors: &mut Vec<String>) {
     wrapper(data.all_repos(), errors, |repo, _| {
-        for env in &repo.environments {
-            if env.name.is_empty() {
+        for (env_name, env) in &repo.environments {
+            if env_name.is_empty() {
                 bail!(
                     "repo {}/{} has an environment with an empty name",
                     repo.org,
@@ -1060,28 +1060,63 @@ fn validate_environments(data: &Data, errors: &mut Vec<String>) {
             // problematic as they can be interpreted as path separators
             //in URLs (/repos/{owner}/{repo}/environments/{name}) and file systems,
             //potentially leading to security vulnerabilities or API routing issues.
-            if env.name.contains('/') || env.name.contains('\\') {
+            if env_name.contains('/') || env_name.contains('\\') {
                 bail!(
                     "repo {}/{} has an environment '{}' with invalid characters (/, \\)",
                     repo.org,
                     repo.name,
-                    env.name
+                    env_name
                 );
+            }
+
+            // Validate branch and tag patterns are not empty and not duplicated
+            let mut seen_branches = HashSet::new();
+            let mut seen_tags = HashSet::new();
+
+            // Validate branch patterns (new field)
+            for branch in &env.branches {
+                if branch.is_empty() {
+                    bail!(
+                        "repo {}/{} environment '{}' has an empty branch pattern",
+                        repo.org,
+                        repo.name,
+                        env_name
+                    );
+                }
+                if !seen_branches.insert(branch) {
+                    bail!(
+                        "repo {}/{} environment '{}' has duplicate branch pattern '{}'",
+                        repo.org,
+                        repo.name,
+                        env_name,
+                        branch
+                    );
+                }
+            }
+
+            // Validate tag patterns (new field)
+            for tag in &env.tags {
+                if tag.is_empty() {
+                    bail!(
+                        "repo {}/{} environment '{}' has an empty tag pattern",
+                        repo.org,
+                        repo.name,
+                        env_name
+                    );
+                }
+                if !seen_tags.insert(tag) {
+                    bail!(
+                        "repo {}/{} environment '{}' has duplicate tag pattern '{}'",
+                        repo.org,
+                        repo.name,
+                        env_name,
+                        tag
+                    );
+                }
             }
         }
 
-        // Check for duplicate environment names
-        let mut seen = HashSet::new();
-        for env in &repo.environments {
-            if !seen.insert(&env.name) {
-                bail!(
-                    "repo {}/{} has duplicate environment '{}'",
-                    repo.org,
-                    repo.name,
-                    env.name
-                );
-            }
-        }
+        // No need to check for duplicate environment names since HashMap keys are unique
         Ok(())
     });
 }
--- a/sync-team/src/github/api/read.rs
+++ b/sync-team/src/github/api/read.rs
@@ -53,8 +53,12 @@ pub(crate) trait GithubRead {
     ) -> anyhow::Result<HashMap<String, (String, BranchProtection)>>;
 
     /// Get environments for a repository
-    /// Returns a list of environment names
-    fn repo_environments(&self, org: &str, repo: &str) -> anyhow::Result<Vec<Environment>>;
+    /// Returns a map of environment names to their Environment data
+    fn repo_environments(
+        &self,
+        org: &str,
+        repo: &str,
+    ) -> anyhow::Result<HashMap<String, Environment>>;
 }
 
 pub(crate) struct GitHubApiRead {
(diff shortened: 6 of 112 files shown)


repos/rust-lang/edition-guide.toml
```diff
@@ -12,6 +12,6 @@ pattern = "master"
 ci-checks = ["Success gate"]
 required-approvals = 0
 
-[[environments]]
-name = "github-pages"
+[environments.github-pages]
+branches = []
```
Kobzol: Just to avoid a common default case, if the branches array is empty, I'd remove it from the current files, and keep the default.

amustaque97: done!

src/schema.rs
```diff
@@ -814,7 +814,7 @@ pub(crate) struct Repo {
     #[serde(default)]
     pub crates_io_publishing: Vec<CratesIoPublishing>,
     #[serde(default)]
-    pub environments: Vec<Environment>,
+    pub environments: HashMap<String, Environment>,
```
marcoieni: I prefer BTreeMap so that the order is consistent.
Useful for testing for example.

sync-team/src/github/api/read.rs
```diff
@@ -441,27 +445,75 @@ impl GithubRead for GitHubApiRead {
...
+                    branches.extend(resp.branch_policies.into_iter().map(|p| p.name));
+                    Ok(())
+                },
+            );
+
+            // If the API call fails (e.g., 404 when no policies are configured),
+            // treat it as an empty branches list
+            if result.is_err() {
+                branches.clear();
+            }
```
marcoieni: this silently fails in case of other errors. E.g. 403. 

In the response of the list environments endpoint, you have the node:

```
"protection_rules": [
        {
          "id": 20383473,
          "node_id": "GA_kwDOIbxIDM4BNwbx",
          "type": "branch_policy"
        }
      ],
```

which tells you if there are any branch policies. Maybe if that is present, we should also call the branch endpoint. Wdyt?

In this way, 404 can never happen and we do less API calls.

sync-team/src/github/api/write.rs
```diff
@@ -500,14 +507,184 @@ impl GitHubWrite {
...
             let url = GitHubUrl::repos(org, repo, &format!("environments/{}", name))?;
-            self.client
-                .send(Method::PUT, &url, &serde_json::json!({}))?;
+
+            let body = if branches.is_empty() {
+                serde_json::json!({})
+            } else {
+                serde_json::json!({
+                    "deployment_branch_policy": {
+                        "protected_branches": false,
```
marcoieni: what's this `protected_branches` field? Why do we need to set it?

sync-team/src/github/api/write.rs
```diff
@@ -500,14 +507,184 @@ impl GitHubWrite {
...
+                })
+            };
+
+            self.client.send(Method::PUT, &url, &body)?;
+
+            // This ensures old policies are cleaned up when transitioning to unrestricted deployments
+            self.set_environment_branch_policies(org, repo, name, branches)?;
+        }
+        Ok(())
+    }
```
marcoieni, line 570: this method is identical to create_environment

sync-team/src/github/api/write.rs
amustaque97: I'm using api endpoint https://docs.github.com/en/rest/deployments/environments?apiVersion=2022-11-28#create-or-update-an-environment. I assume that we will deploy branches with patterns, so setting protected_branches as false

Screenshot 2025-12-05 at 12 37 58 PM

sync-team/src/github/api/write.rs
amustaque97: I guess I messed up while doing some refactoring during development. Thanks for pointing it out.

sync-team/src/github/mod.rs
```diff
@@ -865,7 +878,7 @@ struct CreateRepoDiff {
     settings: RepoSettings,
     permissions: Vec<RepoPermissionAssignmentDiff>,
     branch_protections: Vec<(String, api::BranchProtection)>,
-    environments: Vec<String>,
+    environments: Vec<(String, Vec<String>)>,
```
marcoieni: I would substitute `Vec<String>` with a struct

marcoieni:
I ran a [dry run](https://github.com/rust-lang/team/pull/2159#issuecomment-3617834360) of this in https://github.com/rust-lang/team/pull/2142
the diff is like this:
```
 Environments:
        🔄 Update: github-pages
```

Imo the diff should show that (in this case) it's deleting the branch protection.
So please show in the diff whether you are adding/removing branches 👍

sync-team/src/github/mod.rs
amustaque97: I would like to change either whole  `Vec<(String, Vec<String>)>` as a struct, not just `Vec<String>`. IMO, it will make it unnecessarily complex. 
```rust
struct Branches {
    names: Vec<String>
}
```


I would like to keep it as it is. WDYT?

marcoieni:
From https://github.com/rust-lang/team/pull/2159#issuecomment-3617834360 you can update the toml files to import the branches 👍

amustaque97:
> From [#2159 (comment)](https://github.com/rust-lang/team/pull/2159#issuecomment-3617834360) you can update the toml files to import the branches 👍

I made the required changes 🙌🏻

marcoieni:
the dry run doesn't look clean https://github.com/rust-lang/team/pull/2159#issuecomment-3617834360
image

rust_team_data/src/v1.rs
```diff
@@ -175,7 +175,7 @@ pub struct Repo {
     pub members: Vec<RepoMember>,
     pub branch_protections: Vec<BranchProtection>,
     pub crates: Vec<Crate>,
-    pub environments: Vec<Environment>,
+    pub environments: IndexMap<String, Environment>,
```
Kobzol, line 178: This is a breaking change. I think that we can leave the environments as they were in the v1 API, and just add a `branches` field to the environment.

src/validate.rs
```diff
@@ -1058,28 +1058,39 @@ fn validate_environments(data: &Data, errors: &mut Vec<String>) {
...
+            for branch in &env.branches {
+                if branch.is_empty() {
+                    bail!(
+                        "repo {}/{} environment '{}' has an empty branch name",
+                        repo.org,
+                        repo.name,
+                        env_name
+                    );
+                }
+                if !seen_branches.insert(branch) {
```
Kobzol, line 1086: Is this really an issue? Why couldn't I have two environments on the same branch?

sync-team/src/github/mod.rs
```diff
@@ -949,7 +976,8 @@ struct UpdateRepoDiff {
 
 #[derive(Debug)]
 enum EnvironmentDiff {
-    Create(String),
+    Create(String, Vec<String>),
```
Kobzol: Could you please use named structs here? It's not clear what `String, Vec<String>, Vec<String>` represents :)

sync-team/src/github/mod.rs
```diff
@@ -1092,7 +1123,36 @@ impl std::fmt::Display for UpdateRepoDiff {
...
             for env_diff in environment_diffs {
                 match env_diff {
-                    EnvironmentDiff::Create(name) => writeln!(f, "    ➕ Create: {name}")?,
+                    EnvironmentDiff::Create(name, branches) => {
+                        writeln!(f, "    ➕ Create: {name}")?;
+                        if !branches.is_empty() {
+                            writeln!(f, "        Branches: {}", branches.join(", "))?;
+                        }
+                    }
+                    EnvironmentDiff::Update(name, old_branches, new_branches) => {
```
Kobzol: Also, we usually precompute the "what to add" and "what to remove" things before creating the diff. Could you please modify it so that the update diff already contains branches to add and branches to remove, rather than old/new branches?

marcoieni:
the dry run now is clean, thanks!

rust_team_data/src/v1.rs
amustaque97: This would not be a problem in any way. All the migrations/changes have been taken care of. Please don't hesitate to correct me.

Kobzol: To clarify, the contents of the v1 API is used by several other projects. If this PR is merged as-is, those projects will try to deserialize this field as a `Vec` (because they will still have the old `serde` definition), but the API will now expose it as a JSON dictionary. So the repository endpoint will stop working for them.

src/validate.rs
amustaque97: > Is this really an issue?

Not really, this is for better logging and checks. This will be helpful when we try to add the same branch twice or more in any environment. 

> Why couldn't I have two environments on the same branch?

It’s allowed — a single branch can belong to multiple environments. As you can see in the screenshot below, we iterate over each environment and check for duplicates within that scope.

Screenshot 2025-12-11 at 5 25 12 PM

sync-team/src/github/mod.rs
amustaque97: This change has been made.

sync-team/src/github/mod.rs
amustaque97: I got you, and it 100% makes sense. I have made the required changes.

src/validate.rs
Kobzol: Sorry :facepalm: I read the code wrong. Nevermind!

rust_team_data/src/v1.rs
amustaque97: I'm not sure. Is there any way to confirm this? As environment feature was implemented very recently https://github.com/rust-lang/team/pull/2127, if we still have time, I would like to use the current structure because IMO this is how the schema should be.

I would love to hear your thoughts @Kobzol

Kobzol: Hmm, in theory if the projects haven't updated their `team` version *yet*, it would be doable. Here is a list of some projects that use the API: https://github.com/search?type=code&q=%22rust_team_data+%3D+%22+org%3Arust-lang We'd need to check what version of `team` they use, and if they use the repos endpoint.

Kobzol:
Some repositories use tags for protection rules, not branches. This PR will (I think) set everything as branches. I wonder if that will work when a repository then tries to deploy from a tag, have you tried that? The type (branch/tag) is not mandatory when creating the branch policy, but I think that if it doesn't match, it might be an issue. So in that case we might want to explicitly model both branches and tags in `team`.

rust_team_data/src/v1.rs
amustaque97: Thanks for the nudge, I got lucky this time. I verified all the repos using the link and can confirm that it is safe to accept the proposed changes.

amustaque97:
> Some repositories use tags for protection rules, not branches. This PR will (I think) set everything as branches. I wonder if that will work when a repository then tries to deploy from a tag. Have you tried that? The type (branch/tag) is not mandatory when creating the branch policy, but I think that if it doesn't match, it might be an issue. So in that case we might want to explicitly model both branches and tags in `team`.

Whoops 😬 , I implemented changes based on the comment https://github.com/rust-lang/team/issues/2132#issuecomment-3591469765; I have added support for tags, too.  Can you please review the PR again?

rust_team_data/src/v1.rs
Kobzol: Looks like the only repo that uses the repos endpoint is `triagebot`, and that should be fine, yeah, as it doesn't have the `environments` field yet.

Kobzol:
I see, I didn't know that. It seemed in https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/mdbook.20trusted.20publishing.20broken/with/563418198 that tags would be required. I'll ask t-infra on the next meeting.

marcoieni:
> I see, I didn't know that. It seemed in https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/mdbook.20trusted.20publishing.20broken/with/563418198 that tags would be required. I'll ask t-infra on the next meeting.

I don't see why we wouldn't want to do it 🤔

docs/toml-schema.md
```diff
@@ -433,16 +433,37 @@ merge-bots = ["homu"]
...
+# No tag patterns specified - no tags can deploy
+
+[environments.development]
+# No branch or tag patterns specified - any branch or tag can deploy
+
+## Legacy syntax (still supported for backwards compatibility):
+# The old "branches" and "deployment-patterns" fields are still supported
+# and will be automatically merged with the new "branch" and "tag" fields
+[environments.legacy-example]
+branches = ["main", "stable"]  # Old style - treated as branch patterns
```
marcoieni: here you say `branches`, while above you just say `branch`

docs/toml-schema.md
```diff
@@ -433,16 +433,37 @@ merge-bots = ["homu"]
...
-
-[[environments]]
-name = "staging"
+# Use table-of-tables syntax where the key is the environment name
+[environments.production]
+# List of branch patterns that can deploy to this environment (optional)
+# If empty or omitted, any branch can deploy
+branch = ["main", "release/*"]
+# List of tag patterns that can deploy to this environment (optional)
+tag = ["v*", "release-*"]
```
marcoieni: should this be `tags`?

Kobzol: `branches` and `tags` probably makes more sense, as it's both arrays.

amustaque97:
All the comments are addressed
1. consistent naming of attributes, i.e. branches and tags
2. Updated repos directory with required tags as asked in comment https://github.com/rust-lang/team/issues/2132#issuecomment-3646201406

It is ready for review 👍🏻

docs/toml-schema.md
```diff
@@ -433,16 +433,33 @@ merge-bots = ["homu"]
...
+branches = ["develop", "staging"]
+# No tag patterns specified - no tags can deploy
+
+[environments.development]
+# No branch or tag patterns specified - any branch or tag can deploy
+
+## Legacy syntax (still supported for backwards compatibility):
+# The old "deployment-patterns" field is still supported
+[environments.legacy-patterns]
+deployment-patterns = [
```
marcoieni: is this used by any of our github repos? If yes, can we migrate these repos to the new system so that the team code is simpler?

marcoieni: I don't see it used in any of the toml files. Since it's legacy, why should we support this?

amustaque97: whooops! This was a stupid mistake :( This was another implementation approach I tried and generated docs using LLM, and forgot to remove.

amustaque97: This is fixed now. Thank you again for highlighting it.

src/schema.rs
```diff
@@ -887,5 +887,21 @@ pub(crate) struct CratesIoPublishing {
...
 pub(crate) struct Environment {
+    /// Branch patterns that can deploy to this environment
+    #[serde(default)]
+    pub branches: Vec<String>,
+    /// Tag patterns that can deploy to this environment
+    #[serde(default)]
+    pub tags: Vec<String>,
+    /// Legacy field for backwards compatibility (old deployment-patterns)
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub deployment_patterns: Option<Vec<LegacyDeploymentPattern>>,
```
marcoieni:
```suggestion
```
we said we want to delete this, right?

amustaque97: I have removed it from the codebase.

sync-team/src/github/api/write.rs
```diff
@@ -500,14 +508,243 @@ impl GitHubWrite {
...
+            self.client.send(Method::PUT, &url, &body)?;
+
+            // Always sync branch/tag policies to ensure cleanup of old policies
+            self.set_environment_deployment_patterns(org, repo, name, branches, tags)?;
+        }
+        Ok(())
+    }
+
+    /// Get existing branch policies for an environment
+    fn get_environment_branch_policies(
```
marcoieni, line 573: maybe this function needs to be in the `read.rs` file?
But it's fine I guess.

marcoieni:
there's a conflict 😭 can you fix it?

sync-team/src/github/api/write.rs
Kobzol: I was also thinking that, but it's not so easy currently to access the read stuff from the write part of the API. That being said, I think it would be better if all the state was pre-computed beforehand, and then the write part only did a bunch of modifications, rather than the update part having to figure out what needs to be changed.

marcoieni: anyway this is non blocking I think. It could be addressed in a later pr

marcoieni:
the [dry run ](https://github.com/rust-lang/team/pull/2159#issuecomment-3617834360)still shows diffs:
```
    📝 Editing repo 'rust-lang/ar_archive_writer':
      Environments:
        🔄 Update: publish
            Adding branches: *
    📝 Editing repo 'rust-lang/cargo':
      Environments:
        🔄 Update: release
            Adding branches: *
    📝 Editing repo 'rust-lang/measureme':
      Environments:
        🔄 Update: publish
            Adding branches: *
```

amustaque97:
> 📝 Editing repo 'rust-lang/ar_archive_writer':
>       Environments:
>         🔄 Update: publish
>             Adding branches: *
>     📝 Editing repo 'rust-lang/cargo':
>       Environments:
>         🔄 Update: release
>             Adding branches: *
>     📝 Editing repo 'rust-lang/measureme':
>       Environments:
>         🔄 Update: publish
>             Adding branches: *

TBH, I didn't touch this area today. I removed the "*" in the toml files. Can you please dry-run again? This is annoying, sorry! I want to ensure I don't mess it up and that changes look clean and correct.

Issue #1863, opened 2025-06-03
Detect when a repo is not tracked


marcoieni:
Recently we added many repositories in the team repo. See: https://github.com/rust-lang/team/tree/master/repos

how do we make sure that no other repositories are created manually? We could add a CI check that runs periodically and fail the job if repos from our organizations contains repositories not tracked by the `team` repo.

amustaque97:
I would like to work on the issue 🙌🏻

marcoieni:
Nice! Also, let me know if you have other ideas alternative to the approach I described

amustaque97:
> Nice! Also, let me know if you have other ideas alternative to the approach I described

How about if we follow a similar approach to _GitHub Status Page_ on Zulip chat. Set up a webhook that listens for repository creation events and posts a notification.

Kobzol:
That wouldn't be a robust solution for `team`, but it might be useful for t-infra as a general notification of something happening in rust-lang. If it's possible to set it up using some simple RSS, e.g. as with https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/github.20status/with/560727868, that oculd be nice.

amustaque97:
> That wouldn't be a robust solution for team

I would like to understand what the next steps are after CI failure.

Kobzol:
Well, t-infra would likely get a GH ping, that might be enough :) Or we can send a message to a Zulip channel, as we already do e.g. with [josh-sync](https://github.com/rust-lang/josh-sync/blob/main/.github/workflows/rustc-pull.yml#L154). And then we would have to go and open a PR on this repo. Or the CI action could actually pre-generate the TOML file based on GitHub data and send the PR itself :) But I'm not sure if it's worth the maintenance effort for such logic.

amustaque97:
Interesting! How about combining the CI check with automatic PR creation? Reviewers would still get notifications, or we could use the _codeowners_ feature to limit notifications to specific people. What do you think?

Kobzol:
We can start with just having a CI job that will fail if there are any extra/new repos, that should be enough as a first step.

PR #2149, opened 2025-12-04
feat: detect untracked repo

amustaque97:
Closes: #1863

### DIFF
--- a/.github/workflows/check-untracked-repos.yml
+++ b/.github/workflows/check-untracked-repos.yml
@@ -1,0 +1,24 @@
+name: Check Untracked Repositories
+
+on:
+  schedule:
+    # Every 6 hours
+    - cron: '0 */6 * * *'
+  workflow_dispatch:
+
+jobs:
+  check-untracked-repos:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Install Rust stable
+        uses: ./.github/actions/setup-rust
+
+      - name: Check for untracked repositories
+        shell: bash
+        run: |
+          cargo build --release
+          ./target/release/rust-team ci check-untracked-repos
--- a/src/ci.rs
+++ b/src/ci.rs
@@ -1,7 +1,8 @@
 use crate::data::Data;
 use crate::schema::RepoPermission;
-use anyhow::Context;
-use std::collections::BTreeSet;
+use anyhow::{bail, Context};
+use log::{debug, info, warn};
+use std::collections::{BTreeSet, HashSet};
 use std::path::{Path, PathBuf};
 
 /// Generates the contents of `.github/CODEOWNERS`, based on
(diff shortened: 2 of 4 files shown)


.github/workflows/check-untracked-repos.yml
```diff
@@ -0,0 +1,42 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  check-untracked-repos:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
marcoieni: please use the latest version of the github actions

.github/workflows/check-untracked-repos.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Cache Cargo dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
marcoieni:
```suggestion
```
let's delete this for simplicity

.github/workflows/check-untracked-repos.yml
```diff
@@ -0,0 +1,42 @@
...
+  schedule:
+    - cron: '0 */6 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  check-untracked-repos:
+    runs-on: ubuntu-latest
```
marcoieni, line 11:
```suggestion
    runs-on: ubuntu-24.04
```
let's fix the runner version to avoid surprises

.github/workflows/check-untracked-repos.yml
```diff
@@ -0,0 +1,42 @@
+name: Check Untracked Repositories
+
+on:
+  schedule:
+    - cron: '0 */6 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
```
marcoieni: why is this needed?

.github/workflows/check-untracked-repos.yml
```diff
@@ -0,0 +1,42 @@
...
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Check for untracked repositories
+        run: cargo run --bin rust-team ci check-untracked-repos
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
marcoieni: this is the token with write access.
Instead we want to use the token with read only access, which is the one we use in the dry run. So I'm afraid we need to do something like this:

https://github.com/rust-lang/team/blob/bbdb6ce9b660bff3b7c535f2ed6358d80ddd9cae/.github/workflows/dry-run.yml#L120

Is there a way to simplify the workflow without having to copy-paste? Not sure.

src/ci.rs
```diff
@@ -170,3 +171,150 @@ fn codeowners_path() -> PathBuf {
...
+    for repo in &untracked {
+        warn!("  - {}/{}", repo.org, repo.name);
+    }
+    warn!("");
+    warn!("Repository details:");
+    for repo in &untracked {
+        warn!("  Organization: {}", repo.org);
+        warn!("  Repository: {}", repo.name);
+        warn!("  ---");
+    }
```
marcoieni: what's the point of printing the repos twice?

src/ci.rs
```diff
@@ -170,3 +171,150 @@ fn codeowners_path() -> PathBuf {
         .join(".github")
         .join("CODEOWNERS")
 }
+
+#[derive(Debug, serde::Deserialize)]
+struct GitHubRepo {
+    name: String,
+    #[allow(dead_code)]
+    description: Option<String>,
```
marcoieni: why reading this field if we don't use it?

.github/workflows/check-untracked-repos.yml
Kobzol: I *think* that secrets can't be passed between jobs, so a reusable job probably wouldn't work. But a reusable workflow might, if we parametrized what gets executed in it.

amustaque97: I made some changes, I don't know if this will be liked or not. I wanted to give it a try.

.github/workflows/check-untracked-repos.yml
amustaque97: This is no longer needed, so I removed it.

src/ci.rs
amustaque97: Sorry, this was supposed to be a local change, pushed it accidentally. 

This has been removed.

src/ci.rs
amustaque97: My bad, this was added initially and forgot to remove 😓

.github/workflows/check-untracked-repos.yml
amustaque97: > I _think_ that secrets can't be passed between jobs, so a reusable job probably wouldn't work. But a reusable workflow might, if we parametrized what gets executed in it.

Whoops! I can see other comments now. I was reading about https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action and https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows. I came up with the action changes. It should work 🤞🏻

.github/workflows/check-untracked-repos.yml
amustaque97: Sure, upgraded to version v6

.github/actions/setup-rust-and-tokens/action.yml
marcoieni, line 1: can you please open a new PR where you only add this action, extracting it from the existing workflow? So that we know that it works.

Also, I don't like that this action does both rust stuff and github apps stuff.
I would live the rust stuff alone, or create two separate actions.

amustaque97: PR link: https://github.com/rust-lang/team/pull/2152

amustaque97:
I have resolved all the comments, synced with the latest main branch and squashed commits. PR is ready for review 🙌🏻

Kobzol:
I'm kinda confused why does this needs any permissions at all? We can just read the publicly available list of repositories without any GitHub token, right? I don't think that we need to even read private repositories; in fact this action could leak the existence of said repositories when it runs on CI (sometimes private repos can be created manually, I think, e.g. for security responses, I don't even think that `team` currently tracks all private repositories).

That would also allow this CI workflow run completely unprivileged, which would be better in general.

By the way, this will probably fail (at least) for the `gll` repository, which is still untracked.

amustaque97:
> I'm kinda confused why does this needs any permissions at all? We can just read the publicly available list of repositories without any GitHub token, right? I don't think that we need to even read private repositories; in fact this action could leak the existence of said repositories when it runs on CI (sometimes private repos can be created manually, I think, e.g. for security responses, I don't even think that `team` currently tracks all private repositories).
> 
> That would also allow this CI workflow run completely unprivileged, which would be better in general.
> 
> By the way, this will probably fail (at least) for the `gll` repository, which is still untracked.

Interesting, thank you for saving us from a leak. I never realised until now that we only manage public repos using _team_ repos. I'm glad we found it soon. I have refactored the code to use the public api. 

Can you please take a second look?

src/ci.rs
```diff
@@ -170,3 +171,140 @@ fn codeowners_path() -> PathBuf {
...
+
+fn find_untracked_repos(
+    github_repos: &[(String, GitHubRepo)],
+    tracked_repos: &HashSet<(String, String)>,
+) -> Vec<UntrackedRepo> {
+    github_repos
+        .iter()
+        .filter(|(org, repo)| {
+            // Skip archived repos
+            if repo.archived {
```
Kobzol: I don't think we need to skip archived repositories, we should also have those in `team`.

Kobzol (commented):
It looks great, thank you!

We should eventually just switch to `octocrab` for the simple stuff, or at least unify the `team` and `sync-team` GitHub APIs, it's annoying to reimplement the same GitHub endpoints again and again :D

Kobzol:
Opened https://github.com/rust-lang/team/pull/2180 to unblock this.

marcoieni:
> It looks great, thank you!
> 
> We should eventually just switch to `octocrab` for the simple stuff, or at least unify the `team` and `sync-team` GitHub APIs, it's annoying to reimplement the same GitHub endpoints again and again :D

Let's create our own common crate and not depend on octocrab, so that:
- we have full control of the http client and we can implement better error handling, retries, etc
- we minimize dependencies

At least this is my preference. Because for example we might want to use graphql for certain calls.

We could create a common crate later as a separate issue

Kobzol:
So, I agree that we should definitely have our own *interface* (and use that in both `team` and `sync-team`). But inside that interface, we can just offload the simpler things to `octocrab`, even using the same shared HTTP client. The beauty of `octocrab` is that it allows you to do even fully custom REST API calls, and even GraphQL calls (we do that in `bors`), using the same authenticated client. So you can use `octocrab`'s code for the simple stuff, but then still do more advanced stuff manually. And you can also reuse (but also easily extend!) its `serde` types for GitHub REST API datatypes. It's a win-win, IMO.

marcoieni:
Can you customize it with middlewares? E.g. retry automatically on every 500?

Kobzol:
I think that you can configure it using any `tower` middleware (e.g. https://docs.rs/octocrab/latest/octocrab/struct.OctocrabBuilder.html#method.with_service). I'll try it.

EDIT: I forgot that `octocrab` is async. That's not worth the rewrite, IMO. Nevermind.

.github/workflows/dry-run.yml
```diff
@@ -112,4 +112,4 @@ jobs:
           echo "Pull request ${PR}"
           gh pr comment ${PR} --repo rust-lang/team --body-file comment.txt \
             --edit-last \
-            --create-if-none
+            --create-if-none
```
marcoieni: please revert this change

.github/workflows/check-untracked-repos.yml
```diff
@@ -0,0 +1,23 @@
+name: Check Untracked Repositories
+
+on:
+  schedule:
+    - cron: '0 */6 * * *'
```
marcoieni, line 6:
```suggestion
    # Every 6 hours
    - cron: '0 */6 * * *'
```

.github/workflows/dry-run.yml
amustaque97: done!

marcoieni (approved):
Thanks!

PR #2151, opened 2025-12-04
Sync crate owners on crates.io

Kobzol:
This PR implements sync of crate ownership on crates.io (https://github.com/rust-lang/team/pull/2134#pullrequestreview-3527615781). It allows configuring Rust teams that own a given crate on crates.io, and synces the team ownership on crates.io.

It does not yet remove user ownership for other accounts than `rust-lang-owner`; we first need to transition some crates to team ownership before that.

Best reviewed commit-by-commit.

CC @Mark-Simulacrum

### DIFF
--- a/repos/rust-lang/ar_archive_writer.toml
+++ b/repos/rust-lang/ar_archive_writer.toml
@@ -9,7 +9,7 @@ compiler = "write"
 [[environments]]
 name = "publish"
 
-[[crates-io-publishing]]
+[[crates-io]]
 crates = ["ar_archive_writer"]
-workflow-filename = "publish.yml"
-environment = "publish"
+publish-workflow = "publish.yml"
+publish-environment = "publish"
--- a/repos/rust-lang/crates_io_og_image.toml
+++ b/repos/rust-lang/crates_io_og_image.toml
@@ -15,7 +15,7 @@ pr-required = false
 [[environments]]
 name = "release"
 
-[[crates-io-publishing]]
+[[crates-io]]
 crates = ["crates_io_og_image"]
-workflow-filename = "release.yml"
-environment = "release"
+publish-workflow = "release.yml"
+publish-environment = "release"


marcoieni:
merging this PR will remove other owners, right?
Such as the ones in https://crates.io/crates/annotate-snippets
Or am I misunderstanding?

Kobzol:
I see, I didn't realize we still have crates with individual owners. Ok, then *that* logic will have to wait, and I'll rework this PR slightly.

Kobzol:
Ok, new version, now it should do complete sync of team owners.

src/validate.rs
```diff
@@ -1182,6 +1182,31 @@ fn validate_trusted_publishing(data: &Data, errors: &mut Vec<String>) {
...
+            if !publishing.teams.is_empty() && !publishing.disable_other_publish_methods {
+                return Err(anyhow::anyhow!(
+                    "Repository `{repo_name}` configures crates.io access for team(s) `{}` while setting `disable-other-publish-methods = false`. Either remove the team access or set it to `true`.",
+                    publishing.teams.join(", ")
+                ));
+            }
+            for team in &publishing.teams {
+                let Some(team) = data.team(team) else {
+                    return Err(anyhow::anyhow!(
+                        "Repository `{repo_name}` configures crates.io access for team `{team}`, which does not exist.",
```
Mark-Simulacrum, line 1197: Does this mean we can't create a team and add it to a crate in the same PR?

Mark-Simulacrum (dismissed):
We should probably contact crate owners that would get removed with this to let them know, maybe confirm they're ok with it.

src/validate.rs
Kobzol: This just checks if the team exists as a TOML file on disk, so that should be fine.

Kobzol:
In the current state, this would only remove team, not user, owners, and it should not actually remove anyone, as long as I backfilled the teams currently. We only have 5 repos with crates tracked in team currently.

Kobzol:
Rebased.

Kobzol:
Removed invitation of `rust-lang-owner`, as I realized that it doesn't make sense, it already has to own these crates, otherwise it couldn't access them in the first place. That also saves us from hardcoding `rust-lang-owner` in the source code.

Kobzol:
Rebased over the recent changes to the crates.io syncing logic. Unlike trusted publishing, the dry-run should actually work here (still has to be done in a separate PR though).

sync-team/src/crates_io/mod.rs
```diff
@@ -153,6 +173,7 @@ impl SyncCratesIo {
                 config_diffs.extend(configs.iter_mut().map(|c| ConfigDiff::Delete(c.clone())));
             }
 
+            // Sync "trusted publishing only" crate option
```
marcoieni, line 176: comments like this indicate that we should split this function into multiple ones. Maybe as a refactor in another PR :)

sync-team/src/crates_io/mod.rs
```diff
@@ -165,6 +186,48 @@ impl SyncCratesIo {
...
+                .list_crate_owners(&krate.0)
+                .with_context(|| anyhow::anyhow!("Cannot list crate owners of {krate}"))?;
+
+            // Sync team owners
+            let existing_teams: HashSet<CratesIoOwner> = owners
+                .iter()
+                .filter(|owner| match owner.kind() {
+                    OwnerKind::User => false,
+                    OwnerKind::Team => true,
+                })
```
marcoieni, line 202:
```suggestion
                .filter(|owner| owner.kind() == OwnerKind::Team)
```
or maybe you prefer the match to get compiler errors when new kinds are added?
Anyway, we can also change this in another PR

marcoieni (commented):
Can you fix the CI? https://github.com/rust-lang/team/pull/2185#issuecomment-3655130466

sync-team/src/crates_io/mod.rs
Kobzol: I don't think it's probable that new kinds would be added here, but yeah, that was the idea.

Kobzol:
Fixed CI.

PR #2160, opened 2025-12-05
Batch load trusted publishing configs and remove unused ones

Kobzol:
It would be nice to do a dry-run of this, to ensure that we won't be deleting TP configs unexpectedly.

### DIFF
--- a/sync-team/src/crates_io/api.rs
+++ b/sync-team/src/crates_io/api.rs
@@ -166,25 +166,24 @@ impl CratesIoApi {
         Ok(())
     }
 
-    /// Get information about a crate.
-    pub(crate) fn get_crate(&self, krate: &str) -> anyhow::Result<CratesIoCrate> {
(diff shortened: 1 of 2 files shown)


Kobzol:
Marking as draft, I should implement support for paging first.

Kobzol:
Ok, implemented now.

marcoieni:
the dry run failed: https://github.com/rust-lang/team/actions/runs/20100566092/job/57670041118

Kobzol:
That is because the dry run is always using the workflow from `main` (and that doesn't have the env. var. set). So it cannot really work. Do you want me to land setting the env. var. in the dry run in a separate PR, so we can merge it first to make the dry run work?

Kobzol:
https://github.com/rust-lang/team/pull/2172

Kobzol:
image

Looks like two TP configs weren't in `team`, I'll add them back.

PR #2167, opened 2025-12-09
Remove myself from the crates.io team

0xPoe:
I haven't participated in discussions or code reviews recently, so I believe it would be appropriate to mark myself as inactive. cc: @jtgeibel

### DIFF
--- a/teams/crates-io.toml
+++ b/teams/crates-io.toml
@@ -9,7 +9,6 @@ members = [
     "Turbo87",
     "LawnGnome",
     "mdtro",
-    "0xPoe",
     "eth3lbert",
 ]
 alumni = [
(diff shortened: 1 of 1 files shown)


fmease:
Meta: Is Copilot being enabled here intentional? That'd be surprising

Kobzol:
AFAIK it cannot be turned off per repo/org, and it is triggered by the user having it enabled (which they might not even know about).

0xPoe:
> it is triggered by the user having it enabled (which they might not even know about).

Yes, I didn't know that this was triggered by me. I have turned off this feature. Thank you.

Mahir101:
Sir, you are an alma mater now

PR #2171, opened 2025-12-10
Add a cargo-team repo

ehuss:
The Cargo team would like to have a separate repo for storing things that are more directed towards the team, and not directly related to the cargo tool. Primarily this will start with meeting notes, but we may expand to other things we want to capture.

cc @rust-lang/cargo


weihanglo (approved):
Thanks!

jieyouxu (approved):

PR #2175, opened 2025-12-10
Print crate name in deleting TP diff

Kobzol:
This piece of information is quite important.. :sweat_smile:


jieyouxu (approved):

jieyouxu:
(not infra-admin but obvious enough, so merging.)

PR #2178, opened 2025-12-11
Create a GitHub team and a Zulip group for the RFMF design committee

Kobzol:
So that we're pingable on GitHub and Zulip. Also adds a private repository (will need an infra-admin to actually make it private) to share private data, like some details about contributor's current job status, call transcripts and similar.

### DIFF
--- a/repos/rust-lang/funding-private.toml
+++ b/repos/rust-lang/funding-private.toml
@@ -1,0 +1,9 @@
+org = "rust-lang"
+name = "funding-private"
+description = "Private data related to funding Rust Project contributors"
+bots = []
+private-non-synced = true
+
+[access.teams]
+rfmf-design-committee = "write"
+leadership-council = "write"


repos/rust-lang/funding-private.toml
```diff
@@ -0,0 +1,9 @@
+org = "rust-lang"
+name = "funding-private"
+description = "Private data related to funding Rust Project contributors"
```
Mark-Simulacrum, line 3: Q: Do we actually need a fully private repository? I'm seeing that there's an option for Internal repositories now:

Image

Maybe internal would make sense? I wonder if we could use that elsewhere too...

Kobzol: I think this one should indeed be private, as there might be e.g. employment data that should only be shared with the committee and LC/PD/RF.

PR #2179, opened 2025-12-11
Do not panic when a person is not found

Kobzol:
When a team contains a missing person, we used to have a nice error, but after the recent Zulip ID requirement change, the code now panics. This PR fixes that.


marcoieni (approved):

apiraino:
I love the title of this PR

PR #2183, opened 2025-12-13
jyn can use cloud desktops

jyn514:
r? @Kobzol


Kobzol:
jyn wants to have access to Linux aarch64 machines to debug target-specific issues. jyn is ofc a long-time contributor, so we should allow that :)

PR #2168, opened 2025-12-10
feat: implement rulesets for branch protections

amustaque97:
Related to #1735

PR #2188, opened 2025-12-16
refactor: improve environment diffing using BTreeSet

amustaque97:
**Summary**

- Replace HashSet with BTreeSet for deterministic ordering
- Simplify diff_environments logic using set operations directly
- Simplify EnvironmentDiff enum by using tuple variants
- Remove unused patterns_equal function
- Update test snapshots to match the new enum structure

Separate PR is asked in one of the comments https://github.com/rust-lang/team/pull/2168#discussion_r2622187439, and it makes sense. This is done as part of refactoring existing code; there are no functionality changes. 

cc @marcoieni

### DIFF
--- a/sync-team/src/github/mod.rs
+++ b/sync-team/src/github/mod.rs
@@ -7,7 +7,7 @@ use crate::Config;
 use crate::github::api::{GithubRead, Login, PushAllowanceActor, RepoPermission, RepoSettings};
 use log::debug;
 use rust_team_data::v1::{Bot, BranchProtectionMode, MergeBot};
-use std::collections::{BTreeMap, HashMap, HashSet};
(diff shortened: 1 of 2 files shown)


sync-team/src/github/mod.rs
```diff
@@ -1027,9 +1005,7 @@ enum EnvironmentDiff {
         new_branches: Vec<String>,
```
Kobzol, line 1005: I think that it would be better if the sync code directly operated on branches/tags to add or remove, rather than having generic code that then tries to load the current state from GitHub again when doing the sync.

In other words, I would remove `new_branches` and `new_tags`, and keep only add/remove.

marcoieni: I approved this PR because it's an improvement with respect to the previous code.
This improvement can be done in another PR 👍

PR #2195, opened 2025-12-19
Move jsha to alumnis for other groups/teams as well

GuillaumeGomez:
Follow-up of https://github.com/rust-lang/team/pull/2083.

They forgot to remove themselves from rustdoc front-end ping group and from docs.rs.


GuillaumeGomez:
HTTP error.

Issue #2192, opened 2025-12-18
Implement bypass actors for rulesets based on allowed_merge_teams and merge_bots


amustaque97:
## Summary
When creating rulesets from branch protection configurations, we need to add bypass actors based on `allowed_merge_teams` and `merge_bots` to allow specific teams and bots to bypass the ruleset restrictions.

## Current Behavior
The `construct_ruleset` function creates rulesets with merge queue enabled but doesn't populate the `bypass_actors` field. This means that teams and bots that were previously allowed to push directly via branch protections won't have equivalent bypass permissions in rulesets.

## Expected Behavior
Rulesets should include bypass actors that match the behavior of branch protections:
- Teams listed in `allowed_merge_teams` should be added as bypass actors
- Bots listed in `merge_bots` (e.g., Homu, rust-timer) should be added as bypass actors

## Technical Details
Location: `sync-team/src/github/mod.rs`, function `construct_ruleset` (around line 951-952)

The challenge is that bypass actors require:
1. **Team IDs** (not just team names) - requires fetching team information from GitHub API
2. **User IDs** for bots - requires resolving bot usernames to GitHub user IDs

## Implementation Notes
- Need to add API calls to fetch team IDs from team names
- Need to map bot names to their GitHub user IDs
- May need to pass `SyncGitHub` context to `construct_ruleset` to access the GitHub API client

## Related Code
The branch protection equivalent uses `PushAllowanceActor`:
```rust
let mut push_allowances: Vec<PushAllowanceActor> = branch_protection
    .allowed_merge_teams
    .iter()
    .map(|team| api::PushAllowanceActor::Team(...))
    .collect();

amustaque97:
I will pick this now :)

marcoieni:
Closed by https://github.com/rust-lang/team/pull/2304
We will implement missing features if some repos need them.

PR #2199, opened 2025-12-22
feat: implement ruleset bypass actor

amustaque97:
Closes: #2192

### DIFF
--- a/sync-team/src/github/api/mod.rs
+++ b/sync-team/src/github/api/mod.rs
@@ -544,7 +544,6 @@ pub(crate) enum RulesetActorType {
 #[serde(rename_all = "snake_case")]
 pub(crate) enum RulesetBypassMode {
     Always,
-    #[serde(rename = "pull_request")]
     PullRequest,
     Exempt,
 }


sync-team/src/github/mod.rs
```diff
@@ -961,16 +1101,19 @@ pub fn construct_ruleset(
...
+    // - branch_protection.merge_bots (bot user logins)
 
     api::Ruleset {
         id: None,
         name: format!("Branch protection for {}", branch_protection.pattern),
         target: RulesetTarget::Branch,
         source_type: RulesetSourceType::Repository,
         enforcement: RulesetEnforcement::Active,
-        bypass_actors: None,
+        bypass_actors: None, // Will be populated when applying the diff
```
amustaque97, line 1130: We can discuss if this needs to be changed. I can think of making ID resolution methods available on the read client. 

I'm okay with this comment because the current implementation is read-only

sync-team/src/github/api/mod.rs
```diff
@@ -521,25 +521,110 @@ pub(crate) enum RulesetEnforcement {
...
+    /// The bypass mode for the actor. Defaults to "always" per GitHub API.
+    /// - always: Actor can always bypass
+    /// - pull_request: Actor can only bypass on pull requests (not applicable for DeployKey)
+    /// - exempt: Rules won't run for actor, no audit entry created
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) bypass_mode: Option<RulesetBypassMode>,
 }
 
+/// Custom deserializer for actor_id that distinguishes between null and missing
+fn deserialize_actor_id<'de, D>(deserializer: D) -> Result<Option<ActorId>, D::Error>
```
amustaque97: I'm not sure if this can be written better. I would love to learn. This is generated code but it works. Here is the video reference. 

https://github.com/user-attachments/assets/e5ee56f5-4d83-4312-8cb5-a42c42fd82d4
https://github.com/user-attachments/assets/b16f1d52-746b-4afa-9217-7b2a49720f90

marcoieni: Maybe we can simplify things by only implementing what we need.
For example, we only need to allow github apps and github teams for now I think.
For the crates-io repo we only need github apps for sure.
We don't use deploy keys anywhere if I remember correctly. So we don't need to implement them.
I think we should have a pragmatic approach where we implement stuff only when we find we use it in some repos 👍
Maybe I should have catch this earlier so that you saved time implementing this. If that's the case, sorry! D:

amustaque97: I made the required changes and removed the unused code.

sync-team/src/github/api/mod.rs
```diff
@@ -521,25 +521,32 @@ pub(crate) enum RulesetEnforcement {
...
-    RepositoryRole,
+    /// GitHub Team
     Team,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub(crate) enum RulesetBypassMode {
     Always,
+    #[serde(rename = "pull_request")]
```
marcoieni: isn't `rename_all = "snake_case"` already doing this rename?

amustaque97:
> Can you edit a toml file that can take advantage of this so that we can test this PR later? (I still need to review it before testing it)

We don't have to do it for time being until this is stable enough as we've added an undocumented config https://github.com/rust-lang/team/blob/main/config.toml#L84-L86

So it will read the values from `repos/rust-lang/bors.toml`.

Yes we need to test it as i was not able to setup and test bots or allowed-merge-teams 😅

sync-team/src/github/mod.rs
```diff
@@ -904,6 +957,98 @@ pub(crate) fn convert_branch_pattern_to_ref_pattern(pattern: &str) -> String {
...
+    branch_protection: &rust_team_data::v1::BranchProtection,
+    github_write: &api::GitHubWrite,
+) -> anyhow::Result<Vec<api::RulesetBypassActor>> {
+    let mut bypass_actors = Vec::new();
+
+    // Resolve team bypass actors
+    for team_name in &branch_protection.allowed_merge_teams {
+        match github_write.resolve_team_database_id(org, team_name) {
+            Ok(Some(actor_id)) => {
+                bypass_actors.push(api::RulesetBypassActor {
```
amustaque97, line 992: Image

There is no concerete example or an easy way that i can figure out if bors, homu or merge bots are user accounts or a team. So I would need your attention here so i can move accordingly.

sync-team/src/github/mod.rs
```diff
@@ -904,6 +957,98 @@ pub(crate) fn convert_branch_pattern_to_ref_pattern(pattern: &str) -> String {
...
+        let user_login = match merge_bot {
+            MergeBot::Homu => "bors",
+            MergeBot::RustTimer => "rust-timer",
+            MergeBot::Bors => {
+                // Bors uses a GitHub app, which is not configured through team (it is set manually)
+                // Its bypass actor will be roundtripped by sync-team.
+                continue;
+            }
+        };
+        match github_write.resolve_user_database_id(user_login, org) {
```
amustaque97, line 1025: same here as https://github.com/rust-lang/team/pull/2199/changes#r2713998682

sync-team/src/github/api/mod.rs
amustaque97: This was stupid, I missed it while doing self review. This has been removed. Thanks for highlighting it.

marcoieni:
closing in favor of https://github.com/rust-lang/team/pull/2304
If that PR is missing some features, please open a new PR if you want 🙏

PR #2207, opened 2025-12-29
Add myself to the cloud-compute team

samueltardieu:


samueltardieu:
I sometimes have to work on Clippy or rustc from an older laptop with limited CPU and disk space. Also, I use this laptop to review Clippy pull requests, which means I need to compile Clippy and run the testsuite. Using the dev machines would help in those cases.

PR #2216, opened 2026-01-07
Remove `try`/`auto` homu branch protections from `rust-lang/rust`

Kobzol:
~~To be merged once we get rid of homu.~~ We should double-check that `homu` no longer has access to push to these branches (including `main`) after this is merged. We can also delete the `try` and `auto` branches after this is merged.

### DIFF
--- a/repos/rust-lang/rust.toml
+++ b/repos/rust-lang/rust.toml
@@ -29,19 +29,18 @@ triage = "write"
 
 [[branch-protections]]
 pattern = "main"
-merge-bots = ["homu", "bors"]
+merge-bots = ["bors"]
 
 [[branch-protections]]
 pattern = "stable"
-merge-bots = ["homu", "bors"]
+merge-bots = ["bors"]
 
 [[branch-protections]]
 pattern = "beta"
-merge-bots = ["homu", "bors"]
+merge-bots = ["bors"]
 
 [[branch-protections]]
 pattern = "*"
-merge-bots = ["homu"]
 
 [[branch-protections]]
 pattern = "*/**/*"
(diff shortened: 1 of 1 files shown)


Kobzol:
I don't think that there is any point in returning to homu now, so we can merge this.

Mark-Simulacrum:
I think we want to delete the branches before/after merging this, right? Then they will no longer be used at all because we have the "cannot create branch" protection as well.

I think before doing that we should try to double check that there's no AWS and/or GitHub state that references the branch name in OIDC or environment rules or similar, to avoid accidentally exposing credentials to some future creator of the branch name. I'm not sure if there's an authoritative way to do that unfortunately, but at minimum they should be removed from the bors environment (https://github.com/rust-lang/rust/settings/environments/5453059056/edit).

Kobzol:
This PR actually removes them from the bors environment, we can do that through team now :) Not sure about the OIDC stuff though, will check simpleinfra.

Kobzol:
Haven't found any references in simpleinfra, I think/hope we migrated everything to the `bors` environment when we switched off `rust-lang-ci` last year.

marcoieni:
[ERROR rust_team::validate] validation error: GitHub user `taylors0001` changed username to `iownrena`

Can anyone open a PR to fix this and I can approve and merge?

jieyouxu:
Yes, I can do it

jieyouxu:
Oh... it's the reverse of https://github.com/rust-lang/team/pull/2209

marcoieni:
deleted both `auto` and `try` branches
image

PR #2234, opened 2026-01-29
Add foundation staff marker team and include in all/private

jackh726:
I wanted to make this a marker team, but I didn't see an obvious way to concat private zulip stream membership.

(This means they will end up in `@all`, but not sure if that's *actually* what we want. It's a fine stopgap for now.)

### DIFF
--- a/teams/all.toml
+++ b/teams/all.toml
@@ -33,5 +33,9 @@ extra-emails = [
 name = "T-all"
 
 # Private channel with all team members, so that we can have an easy way of reaching them.
+# Also includes Rust Foundation staff, for any relevant non-public communication.
 [[zulip-streams]]
 name = "all/private"
+extra-teams = [
+    "foundation-staff",
+]
--- a/config.toml
+++ b/config.toml
@@ -78,7 +78,8 @@ members-without-zulip-id = [
     "ryankurte",
     "Stammark",
     "therealprof",
-    "zeenix"
+    "zeenix",
+    "abibroom",
 ]
 
 enable-rulesets-repos = [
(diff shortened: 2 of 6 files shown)


Kobzol:
Created an alternative: https://github.com/rust-lang/team/pull/2236

jackh726:
Updated to include remaining foundation staff members

Kobzol (approved):

teams/all.toml
```diff
@@ -35,3 +35,6 @@ name = "T-all"
 # Private channel with all team members, so that we can have an easy way of reaching them.
```
jieyouxu, line 35: Suggestion: can you please update this comment to say "and foundation staff" to reflect the actual configuration?

jieyouxu:
Q: What's the consensus on [#council > Include Rust Foundation staff as part of the Project?](https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/Include.20Rust.20Foundation.20staff.20as.20part.20of.20the.20Project.3F/with/571174233)? I feel like I'm reading "we're generally in favor of this" but I have a hard time figuring out what the actual decision is based on that thread.

teams/all.toml
jackh726: Updated.

jackh726:
> Q: What's the consensus on [#council > Include Rust Foundation staff as part of the Project?](https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/Include.20Rust.20Foundation.20staff.20as.20part.20of.20the.20Project.3F/with/571174233)? I feel like I'm reading "we're generally in favor of this" but I have a hard time figuring out what the actual decision is based on that thread.

I don't think we need the entire council to +1 this, and @Kobzol (who was the only one with a concern) seems on board with this.

PR #2235, opened 2026-01-29
Add rustbot to the `portable-simd` repository

Urgau:
We need to add `rustbot` to the `portable-simd` repository, otherwise triagebot won't have the necessary permissions for all the enabled features.

Following https://github.com/rust-lang/portable-simd/pull/504

cc @calebzulawski @programmerjake


programmerjake:
also i think you meant to mention @calebzulawski

Urgau:
indeed, sorry for the wrong ping

PR #2238, opened 2026-02-03
Give more people access to beyond-refs

tmandry:


tmandry:
Since CI errors on rust-for-linux I added a couple more. Zulip thread here: https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/team.20does.20not.20have.20any.20github.20teams/with/571739380

repos/rust-lang/beyond-refs.toml
```diff
@@ -20,6 +20,12 @@ types = "maintain"
 wg-async = "maintain"
 wg-safe-transmute = "maintain"
 
+[access.individuals]
+BennoLossin = "maintain"
+dingxiangfei2009 = "maintain"
+joshlf = "maintain"
+jswrenn = "maintain"
```
jieyouxu: Question: can this instead be made a proper subteam like `beyond-refs-wiki`? AFAIK we've been trying to have all permissions managed by teams, and not grant individual accesses.

tmandry: Ok, done.

PR #2239, opened 2026-02-04
Add github team for rust-for-linux

tmandry:
This makes it possible to give the team repo permissions, like I was attempting to do in #2238. It also makes it easier to @-mention the group.

cc @ojeda

### DIFF
--- a/teams/rust-for-linux.toml
+++ b/teams/rust-for-linux.toml
@@ -21,3 +21,6 @@ members = [
 
 [permissions]
 bors.rust.try = true
+
+[[github]]
+orgs = ["rust-lang"]


ojeda:
Sounds good to me, thanks!

If there is a better way to ping the group now, then should the pinging notes at https://rustc-dev-guide.rust-lang.org/tests/ecosystem-test-jobs/rust-for-linux.html and/or https://rustc-dev-guide.rust-lang.org/notification-groups/rust-for-linux.html be updated or is that automatic?

tmandry:
~~It can be updated once this merges.~~ The benefit (and drawback) of using the github group is that people can filter group mentions separately on github, whereas `@rustbot ping` does @-mentions of every individual.

jieyouxu:
> It can be updated once this merges. The benefit (and drawback) of using the github group is that people can filter group mentions separately on github, whereas `@rustbot ping` does @-mentions of every individual.

Note that you can only use the github group `@`-ping if you are an rust-lang organization member.

ojeda:
I see, thanks @tmandry @jieyouxu!

PR #2244, opened 2026-02-06
Add another e-mail address for t-survey

Kobzol:
We want to add some people the permissions to create and access a limited set of surveys, without necessarily inviting them to t-survey yet.


apiraino:
(we discussed that so :+1: from me)

jieyouxu (approved):

PR #2248, opened 2026-02-11
feat: handle email encryption via x25519 asymmetric keys

### LINKED ISSUE #2245: Improve email encryption
Discussed in [#t-infra > meeting 2026-02-09](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/meeting.202026-02-09/with/572805624)

Right now we use a symmetric key to encrypt emails, this means that an infra admin needs to encrypt the email for the contributors.
We want to use public key to allow people to self-serve and improve transparency in the process.

Tasks:

- [ ] Create a pair of public/private key
- [ ] hardcode the public key in the sync-team program
- [ ] Save the private key in the github actions secrets
- [ ] create a team command to encrypt the e-mail
- [ ] An infra admin should decrypt all emails encrypted with the symmetric key and re-encrypt them using the public key
- [ ] Decrypt the emails in CI

Carbonhell:
Closes #2245.

**Note: The PR requires two changes by an infra-team admin before being ready to be merged** (cc @marcoieni):

- [x] 1. a new private/public key pair must be generated via the new CLI command `generate-keys`;
- [x] 2. the public key inside `email_encryption,rs` must be replaced with the newly generated one;
- [x] 3. the private key must be configured as a GitHub secret named `EMAIL_PRIVATE_KEY`, which is passed as an environment variable in the same way as the now deprecated `EMAIL_ENCRYPTION_KEY` secret/env var;
- [x] 4. the currently encrypted email addresses must be migrated by decrypting them (via the currently existing `decrypt-email` command on the mai branch), encrypting them via this branch's `encrypt-email` utility, and adding them as changes to this PR;

There's a section at the end of the README which refers to a SSM parameter - for simplicity, I haven't changed its syntax (even though it'd be more correct to rename it to "email-private-key" instead of encryption key).

The encryption scheme uses [x25519-dalek](https://github.com/dalek-cryptography/curve25519-dalek/tree/main/x25519-dalek) to generate a shared secret via Diffie-Hellman, with blake3 (recommended in [RustCrypto](https://github.com/RustCrypto/hashes)) used as a key derivation function to get the key to use with XChaCha20Poly1305 (which was the library used previously for symmetric encryption/decryption).

### DIFF
--- a/rust_team_data/src/email_encryption.rs
+++ b/rust_team_data/src/email_encryption.rs
@@ -19,8 +19,7 @@ const PREFIX: &str = "encrypted+";
 const SUFFIX: &str = "@rust-lang.invalid";
 const KEY_LENGTH: usize = 32;
 const NONCE_LENGTH: usize = 24;
-// TODO ask an infra admin to generate one
-const PUBLIC_KEY: &str = "d1734021de0af5cfeca64482f3c38b3350a38fd4be2e6a88b2c150be4416b261";
+const PUBLIC_KEY: &str = "a5f80676b1368fe1b2e903058810a6de86ecf2be1225a64e6d79c8d747557b7b";
 // Globally unique context (see here for details: https://docs.rs/blake3/latest/blake3/fn.derive_key.html)
 const KDF_CONTEXT: &str = "rust-team 2026-02-14 email-encryption";
 
--- a/people/camelid.toml
+++ b/people/camelid.toml
@@ -2,4 +2,4 @@ name = 'Noah Lev'
 github = 'camelid'
 github-id = 37223377
 zulip-id = 307537
-email = "encrypted+ebe6f3cec2ee373b57408f88ad0f14dc86c1f1bfaf7829aa5628073ae74e78ae52bc02b1b3ae221ed92284b923258b53a2572142cd62de3f92244cb7045d9058@rust-lang.invalid"
+email = "encrypted+02728fc388dadf662dc77fcec3e7c6d5ed7ae47a92026fc9c75744c45b9427564b4123afcacfdc46a5b0a8a8ebc414adecdd37d617e0a6cea3678488ea95fc84634266bebba24e96d44fd7b7f9fc5c5545eeb8650d5c19582797c70a39c30507@rust-lang.invalid"
--- a/teams/archive/project-foundation.toml
+++ b/teams/archive/project-foundation.toml
@@ -24,16 +24,10 @@ email = "foundation@rust-lang.org"
 [[lists]]
 address = "foundation@rust-lang.org"
 include-team-members = false
-extra-emails = [
-    "encrypted+f1dde3df4bdeb7f371978faf7a9310b5e247fcb70c8b7fbd46979965a4241f9d82d353116ddf82e755344cbb04038159513b23c7cd3ce11a5a90773bb707bca54c189ae6cd@rust-lang.invalid",
-]
 
 [[lists]]
 address = "foundation-moderation@rust-lang.org"
 include-team-members = false
-extra-emails = [
-    "encrypted+c973fe12f465d22c7391b4c666f86d98214a224d354a74631b423fadce1c2bd40cd2da1c847cf3c6ff290db7368049b58876e3655ee1db4efb1fa5185f6c07d722c921a802@rust-lang.invalid",
-]
 
 [[github]]
 orgs = ["rust-lang"]


rust_team_data/src/email_encryption.rs
```diff
@@ -2,70 +2,126 @@
...
     )
     .map_err(|_| Error::InvalidUtf8)
 }
 
-fn init_cipher(key: &str) -> Result<XChaCha20Poly1305, Error> {
-    if key.len() != KEY_LENGTH {
-        return Err(Error::WrongKeyLength);
-    }
-    let key = Key::from_slice(key.as_bytes());
+fn init_cipher(key: &[u8]) -> Result<XChaCha20Poly1305, Error> {
```
ubiratansoares: Why the original validation is no longer needed?

README.md
```diff
@@ -187,15 +187,15 @@ it. Encrypted email addresses look like this:
...
 ```
 
-The production key is accessible to select Infrastructure Team members, so if
-you need to add an encrypted email address you'll need to reach out to that
-team. The key is stored in the following parameter on AWS SSM Parameter Store:
+The `cargo run encrypt-email` CLI command can be used to encrypt an email address in a self-service manner.
+Decryption is automatically done as part of the sync process executed in the CI.
+
+The production private key is accessible to select Infrastructure Team members, and
+it is stored in the following parameter on AWS SSM Parameter Store:
```
ubiratansoares, line 204: Considering the change we are proposing here, I think we could store the private key somewhere else. Ofc we'll have it as CI secret, but eventually 1password could be an option for infra admins as well, just in case we need to decrypt existing emails and get in touch with people for whatever reason. 

Hence, I do think we could simplify a bit this approach. Not sure what other infra admins think about this 👀

rust_team_data/src/email_encryption.rs
Carbonhell: Because the shared key generated via the KDF is already guaranteed to be of 32 bytes as per the docs: https://docs.rs/blake3/latest/blake3/struct.Hash.html
Although this can also easily be enforced by specifically accepting slices of 32 bytes instead of generic u8 slices. Will do that

ubiratansoares: Ah ok! Got it!

Asked mostly to understand whether should we return an instance of `XChaCha20Poly1305` directly here

Carbonhell:
> Hi @Carbonhell, thanks for rising this PR! Great to have this contribution 🚀
> 
> I'll leave two considerations regarding your proposal
> ## Encryption scheme
> 
> I'm by no means a PKI expert, but when I first evaluated what I'd propose myself to implement this feature, I definetly recollected [rage](https://crates.io/crates/age/0.11.2), the Rust implementation for [age encryption](https://github.com/FiloSottile/age). I use `age` (the Go CLI) to encrypt some of my personal stuff 😄
> 
> `rage` [builds on top of the some dependencies you mentioned](https://docs.rs/crate/age/0.11.2/features), but the implementation for small recipients using the very same x25519 looks super straightforward as per [available examples](https://docs.rs/age/0.11.2/age/#streamlined-apis), and it also seems the right fit for our use case.
> 
> My initial take is that having less code attached to this feature make things simpler to grasp and maitain for the long run. That said : would you mind giving a try on rage and bring some feedback to us, without commiting to it all?
> 
> I _think_ the implementation will become shorter in terms of code, but we could definetly learn from a comparison on binary sizes, private/public key sizes and total dependencies we'd bring to our dep graph.
> 
> After that, perhaps we can make a decision regarding which crypto approach we want to use, perhaps gathering input from other infra folks too 🙂
> ## Migrating existing encrypted emails and setting credentials
> 
> As soon as we have an agreement on the PKI approach - either yours, `rage` or something else - I'd be happy to generate the key, add it to CI, decrypt existing emails and leave a comment to additional changes you can add in a single additional commit to finish this PR 🙂
> 
> Just wanted to let you know that this particular feature is not a priority for infra right now, hence it may take some time between reviews, feedback and action from our side.
> 
> I'll leave some additional comments in the code itself, just to highlight a few things that came to mind

Thank you for the feedback! To be fair, I wasn't aware of `rage`. The rationale for using x25519-dalek and blake3 was due to them being on https://blessed.rs/crates#section-cryptography, but I agree that the steps required to actually encrypt contents aren't ideal / a bit too low level for our needs here, which are harder to maintain and with footguns (e.g., requiring a KDF instead of directly using the shared secret as key). I'll definitely give rage a try and report back about the tradeoffs between each solution! I should be able to before the end of this week :)

Edit: technically, x25519-dalek isn't on blessed.rs, but the ed25519-dalek is, and they're both based on the same core from what I gathered

rust_team_data/src/email_encryption.rs
Carbonhell: Oh, my bad, you're right. There's no need for the Result enum here (it's a leftover from the previous signature of this function) - will also remove it. Thanks for spotting this issue!

README.md
Carbonhell: I'm not familiar with 1password and how you use it, but I think a relevant point is that I can't seem to find the Terraform definition of this parameter on [simpleinfra](https://github.com/search?q=repo%3Arust-lang%2Fsimpleinfra+sync-team&type=code) - I assume it was created directly via the AWS dashboard. This might be an extra point in favor of using 1password to remove this parameter that isn't currently managed under IaC

Carbonhell:
I tried out `rage`, here are the results:
# Pros
- the implementation, as expected, is considerably simpler: you can see how it'd look like [in this branch](https://github.com/Carbonhell/team/blob/2303fdc8097097960ca234f8ba138cc2d244ce4e/rust_team_data/src/email_encryption.rs). The crate is quite ergonomic, allowing parsing a `&str` as Identity (private key) or Recipient (public key) to use it with `decrypt(identity: &impl Identity, ciphertext: &[u8])` and  `age::encrypt(recipient: &impl Recipient, plaintext: &[u8])`, without having to deal with nonces and shared secrets (which is a bit too close to rolling your own crypto to be honest);
- with the `ssh` feature, the key generation command could be removed in favor of using the [ssh feature of age](https://docs.rs/age/latest/i686-pc-windows-msvc/age/ssh/index.html) keys generated via `ssh-keygen`

# Cons
- the size of the resulting target directory is considerably larger (size calculated after `cargo build -p rust_team_data --release --features email-encryption` with `du -sh target/release/`):
    - 37 dependencies with size 55MB as baseline
    - 54 dependencies with size 81MB with the implementation of this PR
    - 186 dependencies with size 209MB with rage
- the resulting encrypted email is considerably bigger:
    - rage: `encrypted+6167652d656e6372797074696f6e2e6f72672f76310a2d3e2058323535313920684832542b487a5a4e5a704552612f4634656632776272313158543477576c667a653061524e57504158300a6c5946664168774669566779327532726e50796468464d68787665703434494677786955423434497749410a2d3e202e5b2d677265617365207e30763621204661397c4f4e730a3333746337476c414233476163374e414a356f31416d4c644d763347647a346e51582f4d7436332b2b59314d64584948325965466f39596f796b59735a66692b0a7a734373476874786974447a487741516a4b456f73477954624a725278485531506b67726c6f51487947490a2d2d2d20566e65754358563777767950316d7546336a41654e7a56726c646970477a576f42673452436863547a38730af607ec84912c7544ba326c066f91ecb3538dfe2f48cadfdba0d2197522233e071b34c9b401a9d72648a9ae52badd5581ff@rust-lang.invalid`
    - this PR: `encrypted+4d85908caaa32211b5c4ec28bfe3ead598f53c539aa1d5b2666320897cc206315098adc459e0ebd89b7bdd35d4231f609d715c8811e597f0c834ddb70c61577bdf55642f02a1f492a8e3232a715cb36254d4fb64e36d9a1c@rust-lang.invalid`

---

Considering the analysis above, I don't think the current implementation is complex enough to warrant bringing in all the dependencies of rage, but I recognize the higher maintenance effort with the current implementation. I'll leave the decision to you and adapt this PR as necessary :)

I also cleaned the signature of `init_cipher` and replaced the `get_kdf` function with [blake3's own KDF](https://docs.rs/blake3/latest/blake3/fn.derive_key.html), which slightly simplifies the implementation and avoids reusing the nonce for two different operations.

Kobzol:
Just wanted to note that a bunch of other infra crates and tools directly depend on `rust_team_data`, so it would be great if it didn't grow so much. Although we don't enable the email feature on most places, I think, so if it's not the default, it shouldn't be such a big deal.

That being said, it seems weird why e-mail encryption should even play a role in `ruat_team_data`, that's just a bunch of data structures, right? Wouldn't it be possible to represent the encrypted e-mail as a bunch of bytes or something, and move the feature into team/sync-team?

Carbonhell:
> Just wanted to note that a bunch of other infra crates and tools directly depend on `rust_team_data`, so it would be great if it didn't grow so much. Although we don't enable the email feature on most places, I think, so if it's not the default, it shouldn't be such a big deal.
> 
> That being said, it seems weird why e-mail encryption should even play a role in `ruat_team_data`, that's just a bunch of data structures, right? Wouldn't it be possible to represent the encrypted e-mail as a bunch of bytes or something, and move the feature into team/sync-team?

Based on a [quick search](https://github.com/search?q=org%3Arust-lang+rust_team_data&type=code&p=1), the `email-encryption` feature isn't used anywhere else other than in this repository. I *think* the rationale behind putting it in `rust_team_data` was to allow sharing the encryption/decryption functions both in the CLI app (under `src`) and inside `sync-team`, to decrypt the emails in [the mailgun module](https://github.com/rust-lang/team/blob/main/sync-team/src/mailgun/mod.rs#L25). But I don't see any blocker in moving them inside `sync-team` and using it from the CLI (which already depends on `sync-team`). It does seem clearer to me as right now I believe they're only meant to be used in this repository.

If you want, I can move the module as needed, removing the email-encryption feature which wouldn't be needed anymore, either in this PR or in a new one. Let me know if you'd appreciate that :)

Edit: I was thinking that it might be worth to document, or enforce via a specific type, that the `email` field under the People struct ([here](https://github.com/rust-lang/team/blob/main/rust_team_data/src/v1.rs#L282) (and I think the [TeamWebsite struct](https://github.com/rust-lang/team/blob/main/rust_team_data/src/v1.rs#L58) too?) may contain an encrypted email that can not be used by downstream consumers of the `rust_team_data` crate. Just to avoid other people writing automations or displaying those emails, ignoring the existence of encrypted ones. Not sure if you agree.

Kobzol:
I think that encrypted e-mails in the public API should just be set to `None`; they are essentially unavailable.

I checked the code again and I really don't see any reason for the encryption code and dependencies to live in `rust_team_data`. @emilyalbini Do you remember why did you add the encryption code to this crate directly, rather to e.g. `sync-team`? Maybe because back then `team` and `sync-team` didn't live in the same repo?

rust_team_data/src/email_encryption.rs
```diff
@@ -2,68 +2,120 @@
...
+use hex::{FromHex, ToHex};
+use std::convert::TryInto;
+use x25519_dalek::{EphemeralSecret, PublicKey, StaticSecret};
 
 const PREFIX: &str = "encrypted+";
 const SUFFIX: &str = "@rust-lang.invalid";
 const KEY_LENGTH: usize = 32;
 const NONCE_LENGTH: usize = 24;
+// TODO ask an infra admin to generate one
+const PUBLIC_KEY: &str = "d1734021de0af5cfeca64482f3c38b3350a38fd4be2e6a88b2c150be4416b261";
```
ubiratansoares: The new value will be 

`a5f80676b1368fe1b2e903058810a6de86ecf2be1225a64e6d79c8d747557b7b`

README.md
ubiratansoares: We decided to stick with AWS SSM rather then using 1password to keep a copy of the private key.  I think it'd be fair to mention the new CI secret we'll introduce with thie PR alongside with the updated parameter as part of this doc.

I'll share the new parameter here as soon as I have it. We'll manage this secret using Terraform

rust_team_data/src/email_encryption.rs
Carbonhell: Done!

Carbonhell:
@ubiratansoares All done!

Carbonhell:
The dry-run workflow seems to have failed, due to what seems to be a spurious network error: https://github.com/rust-lang/team/actions/runs/22410799610/job/64883427085

Carbonhell:
Hey! Had some busy times at work in the last weeks - just want to say I saw the conflicts and I'll resolve them this weekend (or earlier, if possible)

ubiratansoares:
Hey @Carbonhell, thanks for the heads up, and apologies for letting this PR taking so long for another review. It was not intended, rest assured.

I added the `EMAIL_PRIVATE_KEY` secret on Github for the `deploy` environment, as we do with other secrets we use. I did it manually for now to unblock this PR (cc @marcoieni). Let me know if you need anything else from my side.

Carbonhell:
> Hey @Carbonhell, thanks for the heads up, and apologies for letting this PR taking so long for another review. It was not intended, rest assured.
> 
> I added the `EMAIL_PRIVATE_KEY` secret on Github for the `deploy` environment, as we do with other secrets we use. I did it manually for now to unblock this PR (cc @marcoieni). Let me know if you need anything else from my side.

Np for the long review! I'm always lurking on Zulip and I saw the amazing work you're doing with the introduction of SSO - I know you have a lot on your plate 😄 
I've updated the PR to resolve the conflicts, which were about updating the Cargo.toml(s), a minor update for chacha20poly1305 (`AeadInit` -> `KeyInit` in import), moving the `EMAIL_PRIVATE_KEY` README entry in the root one, and updating the match statement in main.rs (`Cli` -> `RootOpts`). I ran 
I've also updated the PR description to reflect that nothing else should be required to proceed with the merge. Thank you!

Issue #1915, opened 2025-07-25
Try to install GitHub apps from this repo again


marcoieni:
We noticed that you can now enable enterprise permissions for GitHub Apps:

<img width="2960" height="2048" alt="Image" src="https://github.com/user-attachments/assets/072f2ef3-9dba-4c2e-9646-877ecbe6e2da" />

This could enable `team` to install github apps from CI again.
At the moment, we install github apps manually.

E.g. `bots = ["renovate"]` doesn't do anything.

The ability to sync github apps was removed in https://github.com/rust-lang/sync-team/pull/128 to run sync-team with a token generated by a GitHub app instead of a personal access token.

## Pay attention to current installations

Some github apps were installed manually. For example:
- I installed renovate in https://github.com/rust-lang/crates-io-auth-action/
- I installed forking-renovate in https://github.com/rust-lang/crates_io_og_image

We need to make sure that these apps are not uninstalled after the change.

Kobzol:
I tried this again in https://github.com/rust-lang/team/pull/2255, but it still does not seem to work.

The problem that we're trying to solve is that we want to figure out in which repositories is a GitHub App A installed, while using GitHub App installation token from GH app B as auth. We use GH app installation token for auth because we want to have fine-grained permissions, as we run this as a "dry-run" in CI, and we only want to have read-only access.

But there does not seem to be a REST/GraphQL endpoint that would tell us "This specific GH app installation is installed in repositories XYZ". [This endpoint](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-repositories-accessible-to-the-app-installation) does not work, because it only shows repositories where the *currently authenticated* app is installed in, and [this endpoint](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-repositories-accessible-to-the-user-access-token) does not seem to be usable with GH app installation auth.

marcoieni:
update on this: Kobzol found a way to make this work, we just need the time to implement it.

PR #2255, opened 2026-02-16
Re-add functionality to sync GitHub apps

Kobzol:
This essentially reverts https://github.com/rust-lang/sync-team/pull/128, plus adapts to code to changes that happened in the meantime.

This is one of the most complicated (and frankly stupid) APIs I had to deal with... It almost seems like they designed each related endpoint to return *slightly* different data, and require different tokens/permissions/authentication type in order to get all the information that we need.

The following environment variables have to be configured for the dry-run workflow:
- `ENTERPRISE_NAME`: just `rust-lang`
- `ENTERPRISE_APP_ID`: app ID of `sync-team-app-read` (`2885328`)
- `ENTERPRISE_APP_INSTALLATION_ID`: installation ID of `sync-team-app-read` in the `rust-lang` enterprise (`110701681`)
- `ENTERPRISE_APP_SECRET_KEY`: secret key of the GH app (not gonna paste that in here :) )

Fixes: https://github.com/rust-lang/team/issues/1915

### DIFF
--- a/src/sync/github/api/mod.rs
+++ b/src/sync/github/api/mod.rs
@@ -223,7 +223,7 @@ impl HttpClient {
                         .map(|r| r.contains(&RelationType::Next))
                         .unwrap_or(false)
                     {
-                        next = Some(GitHubUrl::new(link.link(), next_url.org()));
+                        next = Some(next_url.clone().with_url(link.link()));
                         break;
                     }
                 }
--- a/src/sync/github/api/url.rs
+++ b/src/sync/github/api/url.rs
@@ -40,6 +40,11 @@ impl GitHubUrl {
         self
     }
 
+    pub fn with_url(mut self, url: &str) -> GitHubUrl {
+        self.url = url.to_owned();
+        self
+    }
+
     pub fn repos(org: &str, repo: &str, remaining_endpoint: &str) -> anyhow::Result<Self> {
         let remaining_endpoint = if remaining_endpoint.is_empty() {
             "".to_string()


marcoieni:
To recap, we are still using a PAT to deploy, but we are using a GitHub App for the dry run.

I will give read access to the following resources to the [sync-team-app](https://github.com/organizations/rust-lang/settings/apps/sync-team-app) github app:

image

marcoieni:
From the dry run it looks like we are about to remove all apps:
https://github.com/rust-lang/team/pull/2257#issuecomment-3909328529

Maybe it's because the token doesn't have enough permission even after enabling those other fields?

marcoieni:
maybe if you update the branch the dry run will run? 🤔

marcoieni:
But first, can you edit the github actions workflows to add the environments and secrets you mentioned? After, I will set them 👍

Kobzol:
The dry run executed, but it failed, because the environment variables weren't set.

marcoieni:
I think stuff like `ENTERPRISE_NAME: rust-lang` can be set from the workflow file directly, no? 🤔

src/sync/github/api/tokens.rs
```diff
@@ -30,22 +67,123 @@ impl GitHubTokens {
...
+                TokenType::Enterprise => {
+                    Ok(&enterprise_client_ctx.enterprise_token)
+                }
+            },
             GitHubTokens::Pat(pat) => Ok(pat),
         }
     }
+
+    /// Return the name of the enterprise, if present.
+    pub fn get_enterprise_name(&self) -> anyhow::Result<String> {
```
marcoieni:
```suggestion
    pub fn get_enterprise_name(&self) -> anyhow::Result<&str> {
```

src/sync/github/api/tokens.rs
```diff
@@ -30,22 +67,123 @@ impl GitHubTokens {
...
+        match self {
+            GitHubTokens::App {
+                enterprise_client_ctx,
+                ..
+            } => Ok(enterprise_client_ctx.enterprise_name.clone()),
```
marcoieni:
```suggestion
            } => Ok(&enterprise_client_ctx.enterprise_name),
```

marcoieni:
I set the secret `ENTERPRISE_APP_SECRET_KEY`. 👍

marcoieni:
Testing this PR in https://github.com/rust-lang/team/pull/2379

marcoieni:
In https://github.com/rust-lang/team/pull/2379 I got the error

> Cannot get organization installation for `bors-rs` of the enterprise app

You should filter orgs that are not part of the rust enterprise.

marcoieni:
or maybe https://github.com/rust-lang/team/pull/2381 is the fix for this 🤔

src/sync/github/api/read.rs
```diff
@@ -183,6 +194,82 @@ impl GithubRead for GitHubApiRead {
...
+        } else {
+            let url = GitHubUrl::new(
+                &format!("enterprises/{}/apps/organizations/{org}/installations/{installation_id}/repositories",
+                    self.client.github_tokens.get_enterprise_name()?
+                ),
+                org,
+            )
+            .with_token_type(TokenType::Enterprise);
+            self.client
+                .rest_paginated(
```
marcoieni, line 256: we need to change rest_paginated function here because it resets the token type that you previously set as `Enterprise`:

https://github.com/Kobzol/team/blob/b5e8496cce15f97418167202cc1583387bf51895/src/sync/github/api/mod.rs#L226


One solution is adding a method to GitHubUrl to edit only the url, leaving `token type` and `org` the same.

Eg something like this (keep it as pseudo code, I'm not sure what kind of methods make sense in the sense of mutability and lifetimes):
```
impl GitHubUrl {
    pub fn with_url(&mut self, url: String) -> &mut Self {
        self.url = url;
        self
    }
}
```

Detected with GPT5.4 code review 🤖

src/sync/github/tests/test_utils.rs
```diff
@@ -604,6 +605,18 @@ impl GithubRead for GithubMock {
         Ok(self.get_org(org).members.iter().cloned().collect())
     }
 
+    async fn org_app_installations(&self, _org: &str) -> anyhow::Result<Vec<OrgAppInstallation>> {
+        Ok(vec![])
```
marcoieni, line 609: I wonder if these should be a `todo!`, so that if they are called in tests we get a failure

Kobzol: That would brick all tests, because the function is called in them, the tests use `SyncGithub` in an end-to-end way.

src/sync/github/api/read.rs
Kobzol: Good catch! Fixed.

PR #2256, opened 2026-02-16
add lqd to arm-maintainers and remove outdated entries

davidtwco:
cc @lqd


jieyouxu:
You'll also need to remove the `people/` entry of the removed arm maintainer. (Can you also reflect that removal in the PR title/desc if this is intended?)

davidtwco:
Ah, oops, should be fixed now

PR #2266, opened 2026-02-25
Add @ojeda to goal-owners

tomassedovic:
Miguel Ojeda is the lead for the Rust for Linux project and while I'm the point of contact, he should be able to write send updates or make corrections.


ojeda:
Thanks @tomassedovic!

PR #2267, opened 2026-02-25
Give myself permissions on rust-artwork repo.

m-ou-se:
Nobody currently has access to rust-artwork.

The most relevant thing in this repo is the logo svg, the current version of which I authored five years ago. (https://github.com/rust-lang/rust-artwork/pull/9) I'd like to add the logo in a few more formats and tidy up the repository a bit.

### DIFF
--- a/repos/rust-lang/rust-artwork.toml
+++ b/repos/rust-lang/rust-artwork.toml
@@ -3,5 +3,7 @@ name = "rust-artwork"
 description = "Official artwork for the Rust project."
 bots = []
 
-# There is no current clear owner of this repository, so the access is left empty for now.
 [access.teams]
+
+[access.individuals]
+m-ou-se = "write"


steffahn:
AFAICT, `"write"` is the more common choice here for configuring access these days, rather than `"maintain"` (though I forgot the exact differences since since last time I checked).

Also I guess the “access is left empty for now” comment becomes outdated with the PR.

m-ou-se:
> AFAICT, `"write"` is the more common choice here for configuring access these days, rather than `"maintain"` (though I forgot the exact differences since since last time I checked).

They are almost the same, but "maintain" allows me to adjust the repo description.

> Also I guess the “access is left empty for now” comment becomes outdated with the PR.

Updated.

Kobzol:
The description should be set through `team` :sweat_smile: Or the automation will just overwrite it on the next sync.

```toml
description = "Cool artworkz"
```
in the repo TOML file (it's already set there).

m-ou-se:
Oh cool, didn't realize how powerful the team repository had become.

Changed it to `"write"`.

ehuss:
@jieyouxu I don't have any objection to giving @m-ou-se access, but our policy is to never give individual access to repositories. This probably should have been given to a team (social-media or council?, idk).

jieyouxu:
> @jieyouxu I don't have any objection to giving @m-ou-se access, but our policy is to never give individual access to repositories. This probably should have been given to a team (social-media or council?, idk).

Probably. We can change it to just council? Also a question: was that actually written down somewhere? I tried to cite that policy but couldn't find it last time

m-ou-se:
`social-media` probably makes more sense than `council`. The council is supposed to have power as a group, but not necessarily as indivivduals. Giving power to members of the council individually isn't super useful, because membership rotates. It would be annoying if I lose my access to rust-artwork if someone else starts representing libs on the council.

Happy to send a PR that changes it to `social-media` for now. Adding a T-logo team seems a bit overkill. ^^

PR #2270, opened 2026-02-26
Give `leadership-council` team write access to `rust-artwork` repo instead of individual access

jieyouxu:
As discussed in https://github.com/rust-lang/team/pull/2267#issuecomment-3967067482.

Alternatively, maybe `social-media` team? Not really sure tbh.

cc @ehuss @m-ou-se (if you have preferences instead)

### DIFF
--- a/repos/rust-lang/rust-artwork.toml
+++ b/repos/rust-lang/rust-artwork.toml
@@ -4,6 +4,4 @@ description = "Official artwork for the Rust project."
 bots = []
 
 [access.teams]
-
-[access.individuals]
-m-ou-se = "write"
+leadership-council = "write"


steffahn:
I feel like this is oddly specific for leadership council, and rather something to be delegated to another team. Especially given the lower frequency of interaction, and the factor of rotating council representatives.

I would suggest the repo rename `master` into `main` and turn on branch protection, with the latter turned on there’s not really anything that can go wrong anyway, so quite low responsibility.

I find `social-media` sound like a reasonable start, and if that’s still too few people other teams could be added later without much concern if seen fit.

steffahn:
I pressed the remove button in case any change is desired after my comment. Feel free to re-add to queue if you want to merge it now anyway. @jackh726

jackh726:
I would say that for now, the council having access is a reasonable default for now. The council can delegate this to a different team if they so chose. Otherwise, there's no clear owner and the council is the group responsible for assigning one.

Nonetheless, let's switch from individual to team access here.

steffahn:
Do you know if branch renames and protection work through the team repo, or is it a more manual process? @jackh726

jieyouxu:
Those need infra-admin coordination as far as I am aware (the default branch rename at least)

PR #2273, opened 2026-02-27
Add Program marker team and repo

traviscross:
Now that we have more than one program manager on staff, we need some infrastructure for coordination.  Let's create a marker team for this. It will include those staff hired by the Project (via the Foundation) in support of program management and those Project members who (via the delegation by the council to the Edition and Goals teams) provide oversight and guidance for this program.

We'll create a repository for this marker team so that the PMs can create project boards and store documentation specific to this program.

cc @tomassedovic @nxsaken @ehuss @nikomatsakis

### DIFF
--- a/repos/rust-lang/program-team.toml
+++ b/repos/rust-lang/program-team.toml
@@ -1,0 +1,9 @@
+org = "rust-lang"
+name = "program-team"
+description = "Home of the Rust program team"
+bots = ["rustbot", "rfcbot"]
+
+[access.teams]
+edition = "maintain"
+goals = "maintain"
+program = "maintain"
--- a/teams/program.toml
+++ b/teams/program.toml
@@ -1,0 +1,34 @@
+# The Program team (AKA Program Management, PM) lists the staff hired
+# by the Project (via the Foundation) in support of program management
+# and those Project members who (via the delegation by the council to
+# the Edition and Goals teams) provide oversight and guidance for this
+# program.
+
+name = "program"
+kind = "marker-team"
+subteam-of = "edition" # And of goals.
+
+[people]
+leads = []
+members = [
+    "traviscross",
+    "ehuss",
+    "nikomatsakis",
+    "tomassedovic",
+    "nxsaken",
+]
+
+[[github]]
+orgs = ["rust-lang"]
+
+[permissions]
+bors.rust.review = true
+crater = true
+dev-desktop = true
+perf = true
+
+[[lists]]
+address = "program@rust-lang.org"
+
+[[zulip-groups]]
+name = "T-program"


repos/rust-lang/program-team.toml
```diff
@@ -0,0 +1,10 @@
+org = "rust-lang"
+name = "program-team"
+subteam-of = "edition" # And of goals.
```
nxsaken: Should it be a subteam? Seems cross-cutting enough to be a separate entity

tomassedovic:
This will be really useful to the PM team.

repos/rust-lang/program-team.toml
ehuss:
```suggestion
```

teams/program.toml
```diff
@@ -0,0 +1,33 @@
...
+]
+
+[[github]]
+orgs = ["rust-lang"]
+
+[permissions]
+bors.rust.review = true
+crater = true
+dev-desktop = true
+perf = true
```
jieyouxu, line 28: Question: what's our policy on giving permission to marker teams? Last I recall we seem to be trying to stay away from giving permissions to marker teams.

Should this instead of a proper subteam of LC? Or launching-pad?

I ask because the set of permissions { rust-lang/rust r+, perf } are particularly sensitive.

jieyouxu: (I will also update the forge docs if I get an answer from LC here, this is not documented either AFAIK)

repos/rust-lang/program-team.toml
traviscross: This is matching the delegation structure of https://github.com/rust-lang/leadership-council/issues/151 and https://github.com/rust-lang/leadership-council/issues/255.

teams/program.toml
traviscross: There are other marker teams that have permissions.  The only people who are going to be on this marker team are existing Project members who already have these and staff that is under contract with the Foundation (to work in support of the Project).

Given that, this seems OK and desirable to me.  Someone that we're paying to work on Rust and have under contract should have these perms.

Being an actual team implies other things such as chartered decision-making power within our Project structure.  We may or may not later want to do that with this team, but that seems separable from the question of ensuring permissions to hired staff.

repos/rust-lang/program-team.toml
traviscross: Thanks.  Had put that in the wrong file.

jieyouxu:
Self-note: update `#infra/announcements` need-to-sync zulip groups to include `T-program` once this merges.

teams/program.toml
Mark-Simulacrum: I'm not convinced there's good rationale for r+ rights. The rest seem reasonable and try rights would be needed to make them reasonable, but what is the artifact under rust-lang/rust that these folks are approving? (Many of those in this list have it for other hats).

That said, I don't think that's strictly blocking, those rights are fairly widely distributed and misuse would get some scrutiny if it gets noticed.

jieyouxu:
Since https://github.com/rust-lang/team/pull/2273#discussion_r2881355224 is non-blocking, let's merge for now. We can revisit if `r+` is actually needed.

teams/program.toml
jackh726: I tend to agree with @Mark-Simulacrum

PR #2212, opened 2026-01-06
fix: match rulesets by branch pattern

amustaque97:
Closes #2211 

**Summary**
Improves ruleset diffing to match by branch pattern instead of name (enabling in-place updates rather than delete/recreate)

Issue #2211, opened 2026-01-06
Detect ruleset diff based on the branch pattern


amustaque97:
As encountered in the PR  https://github.com/rust-lang/team/pull/2200#issuecomment-3715211184

The current logic matches rulesets by name, so when the name format changed from "Branch protection for main" to "Ruleset for main", it can't find a match and ends up deleting + recreating instead of updating. 

Expectation is 
- The ruleset will be updated (including renaming from "Branch protection for main" to "Ruleset for main") instead of being deleted and recreated
- Only if the branch patterns don't match will a new ruleset be created
- If an existing ruleset on GitHub protects `refs/heads/main`, and the expected ruleset also protects `refs/heads/main`, they'll be matched together

PR #2298, opened 2026-03-06
Fix synchronization of rulesets

Kobzol:
The [get all rulesets endpoint](https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#get-all-repository-rulesets) does not actually return all fields of each ruleset. So we were loading data from GitHub that did not match what we uploaded there, which caused unclean diffs. That is probably also the reason why some of the fields were wrapped in `Option`s (which should not be necessary based on my testing).

I also changed how the diff is outputted, to provide a much shorter diff based on only the things that are actually being changed.

I did not swap the implementation to check rulesets by patterns. I think that it is easier to just check them by name, and keep a convention that the name is always based on the branch pattern. If we wanted to use a single ruleset per multiple branches (which sounds like a reasonable extension in the future), we could do that, but as long as we only support one branch protection/ruleset per branch, we can use this simpler scheme. I added validation to check that the patterns between different branch protections on a single repo do not overlap.

I noticed that the current implementation synchronizes everything available on GitHub for each ruleset, which doesn't match how we normally do things in `team`. That being said, since it's already implemented, I'd keep it, at least we will be sure that the rulesets weren't tampered with by hand on GitHub.

Fixes: https://github.com/rust-lang/team/issues/2211
Closes: https://github.com/rust-lang/team/pull/2212

### DIFF
--- a/sync-team/src/github/mod.rs
+++ b/sync-team/src/github/mod.rs
@@ -677,11 +677,19 @@ impl SyncGitHub {
                     rules: expected_rules,
                 } = &expected_ruleset;
 
+                // With a read-only GitHub App token, GitHub does not actually return bypass actors
+                // from the API. So we should not check it, otherwise the diff will not be clean.
+                let bypass_actors_differ = if !self.github.uses_pat() {
+                    false
+                } else {
+                    bypass_actors != expected_bypass_actors
+                };
+
                 // Ruleset exists for this branch name, check if it needs updating
                 if target != expected_target
                     || source_type != expected_source_type
                     || enforcement != expected_enforcement
-                    || bypass_actors != expected_bypass_actors
+                    || bypass_actors_differ
                     || conditions != expected_conditions
                     || rules != expected_rules
                 {
@@ -1017,7 +1025,7 @@ pub fn construct_ruleset(branch_protection: &rust_team_data::v1::BranchProtectio
         target: RulesetTarget::Branch,
         source_type: RulesetSourceType::Repository,
         enforcement: RulesetEnforcement::Active,
-        bypass_actors: vec![],
+        bypass_actors: Some(vec![]),
         conditions: RulesetConditions {
             ref_name: RulesetRefNameCondition {
                 include: vec![convert_branch_pattern_to_ref_pattern(
--- a/sync-team/src/github/api/mod.rs
+++ b/sync-team/src/github/api/mod.rs
@@ -523,7 +523,11 @@ pub(crate) struct Ruleset {
     pub(crate) target: RulesetTarget,
     pub(crate) source_type: RulesetSourceType,
     pub(crate) enforcement: RulesetEnforcement,
-    pub(crate) bypass_actors: Vec<RulesetBypassActor>,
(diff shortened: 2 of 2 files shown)


sync-team/src/github/api/read.rs
```diff
@@ -544,20 +546,39 @@ impl GithubRead for GitHubApiRead {
...
+            id: u64,
+        }
+
+        let mut ruleset_ids = vec![];
 
         // REST API endpoint for rulesets
         // https://docs.github.com/en/rest/repos/rules#get-all-repository-rulesets
-        // The API returns an array of rulesets directly, not wrapped in an object
+        // The API returns only a subset of data for each ruleset :/
+        // So we then have to fetch the rulesets individuall to get the full data.
```
marcoieni:
```suggestion
        // So we then have to fetch the rulesets individually to get the full data.
```

sync-team/src/github/mod.rs
```diff
@@ -1656,21 +1674,19 @@ fn log_ruleset(
...
-                writeln!(result, "        {}: {:?}", $str, old)?;
-                logged = true;
-            } else if Some(old) != new_val {
-                if let Some(n) = new_val {
-                    writeln!(result, "        {}: {:?} => {:?}", $str, old, n)?;
-                    logged = true;
-                }
+
+            // If there is a new value and it does not match the old, print a diff
+            if let Some(new_value) = $new_field
```
marcoieni:
```suggestion
            if let Some(new_value) = new_val
```

sync-team/src/github/mod.rs
```diff
@@ -1656,21 +1674,19 @@ fn log_ruleset(
...
+                && new_value != old
+            {
+                writeln!(result, "        {}: {old:?} => {new_value:?}", $str)?;
+            } else if new_val.is_none() {
```
marcoieni, line 1695:
```suggestion
            } else {
```

sync-team/src/github/mod.rs
marcoieni: it should be equivalent, right?

sync-team/src/github/mod.rs
marcoieni: I think we are already in this case, so it's not useful to repeat the if, right?

Kobzol: We don't want to trigger the second branch if there is a new value, but it is the same as the old one.

sync-team/src/github/mod.rs
Kobzol: Yeah, I was struggling with moved value due to the way it was previously written, so this was a temporary change that I forgot to revert, thanks.

marcoieni:
the dry run on my test failed: https://github.com/rust-lang/team/actions/runs/22776512996/job/66071152519

Maybe the API is returning 404? similar to the environments issues we had. Or maybe the field needs to be optional.

marcoieni:
started another dry run: https://github.com/rust-lang/team/pull/2300

Kobzol:
Looks good! Thanks.

PR #2299, opened 2026-03-06
Create linker working area

mati865:
As per https://github.com/rust-lang/compiler-team/issues/964

### DIFF
--- a/teams/wg-linker.toml
+++ b/teams/wg-linker.toml
@@ -1,0 +1,22 @@
+name = "wg-linker"
+subteam-of = "compiler"
+kind = "working-group"
+
+[people]
+leads = []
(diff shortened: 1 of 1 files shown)


teams/wg-linker.toml
```diff
@@ -0,0 +1,22 @@
...
+subteam-of = "compiler"
+kind = "working-group"
+
+[people]
+leads = []
+members = [
+    "davidlattimore",
+    "jyn514",
+    "madsmtm",
+    "mati865"
```
tshepang:
```suggestion
    "mati865",
```

mati865: Thanks

PR #2302, opened 2026-03-09
refactor code to remove macro_rules

marcoieni:
this code should be easier to understand and maintain.


ubiratansoares (approved):
The function reads simpler that the macro. I'm totally fine with a bit more of repetition in this case, but feel free to wait for additional input from others :)

Kobzol (approved):
I didn't like the solution with the slice of fields to diff, but that has since been removed. I'm fine with this, always good to remove a macro.

PR #2046, opened 2025-10-29
Remove branch protections from `crates.io` repository

Turbo87:
The `team` repo currently only supports creating "classic" branch protections, not the more modern "rulesets". Unfortunately rulesets are required to allow GitHub Apps to bypass branch protections (e.g. to push to protected branches without PRs). For the crates.io repository, Marco manually created a ruleset based on the current classic branch protection. Once the `team` repo supports rulesets we will bring them back into the repo.

Context: https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/crates.2Eio.20GitHub.20app/with/547767243

PR #2304, opened 2026-03-09
rulesets: implement allowed-merge-apps

### LINKED ISSUE #2198: Configure the crates-io ruleset
"Revert" PR https://github.com/rust-lang/team/pull/2046 and configure the following ruleset that was created manually:

Image

Note that we would need to add support for the bypass list. (see the allowed workflows-crates-io github app in the screenshot)

marcoieni:
Close https://github.com/rust-lang/team/issues/2198

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -395,6 +395,10 @@ Admins cannot override these branch protections. If an admin needs to do that, t
 [[branch-protections]]
 # The pattern matching the branches to be protected (required)
 pattern = "main"
+# Custom name for the GitHub ruleset created from this branch protection.
+# If not specified, defaults to "Ruleset for <pattern>".
+# (optional)
+name = "My custom ruleset name"
 # Which CI checks to are required for merging (optional)
 # Cannot be set if `pr-required` is `false`.
 #
@@ -421,14 +425,22 @@ required-approvals = 1
 # in [access.teams].
 # (optional)
 allowed-merge-teams = ["awesome-team"]
-# Determines the merge queue bot(s) that manage pushes to this branch.
+# Determines the GitHub Apps that manage pushes to this branch.
 # When a bot manages the queue, some other options, like
 # `required-approvals` and `pr-required` options are not valid.
 #
-# Currently, only the "homu" option is supported.
-# When "homu" is used, "bors" has to be in the `bots` array.
+# Currently supported values: "rust-timer", "bors", "workflows-crates-io".
 # (optional)
-merge-bots = ["homu"]
+allowed-merge-apps = ["bors"]
+# Whether to enable a merge queue for this branch.
+# (optional - default `false`)
+merge-queue = false
+# Whether to prevent branch deletion.
+# (optional - default `false`)
+prevent-deletion = false
+# Whether to prevent force pushes to the branch.
+# (optional - default `false`)
+prevent-force-push = false
 ```
 
 ### Repository environments
--- a/repos/rust-lang/crates.io.toml
+++ b/repos/rust-lang/crates.io.toml
@@ -7,7 +7,23 @@ bots = ["renovate", "rustbot", "heroku-deploy-access"]
 [access.teams]
 crates-io = "write"
 
+# TODO: enable after updating triagebot
+# [[branch-protections]]
+# name = "main"
+# pattern = "main"
+# ci-checks = [
+#     "Backend / Lint",
+#     "Backend / Test",
+#     "Backend / dependencies",
+#     "Frontend / Lint",
+#     "Frontend / Test",
+# ]
+# required-approvals = 0
+# merge-queue = false
+# prevent-deletion = true
+# prevent-force-push = true
+# allowed-merge-apps = ["workflows-crates-io"]
+
 [environments.crates-io]
 
 [environments.staging-crates-io]
-
--- a/repos/rust-lang/rust.toml
+++ b/repos/rust-lang/rust.toml
@@ -29,15 +29,15 @@ triage = "write"
 
 [[branch-protections]]
 pattern = "main"
-merge-bots = ["bors"]
+allowed-merge-apps = ["bors"]
 
 [[branch-protections]]
 pattern = "stable"
-merge-bots = ["bors"]
+allowed-merge-apps = ["bors"]
 
 [[branch-protections]]
 pattern = "beta"
-merge-bots = ["bors"]
+allowed-merge-apps = ["bors"]
 
 [[branch-protections]]
 pattern = "*"
@@ -54,37 +54,37 @@ pr-required = false
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "automation/bors/try"
-merge-bots = ["bors"]
+allowed-merge-apps = ["bors"]
 
 # Required for running try builds created by bors.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "automation/bors/try-merge"
-merge-bots = ["bors"]
+allowed-merge-apps = ["bors"]
 
 # Required for running auto builds created by bors.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "automation/bors/auto"
-merge-bots = ["bors"]
+allowed-merge-apps = ["bors"]
 
 # Required for running auto builds created by bors.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "automation/bors/auto-merge"
-merge-bots = ["bors"]
+allowed-merge-apps = ["bors"]
 
 # Required for unrolled PR builds created by perfbot.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "try-perf"
-merge-bots = ["rust-timer"]
+allowed-merge-apps = ["rust-timer"]
 
 # Required for unrolled PR builds created by perfbot.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "perf-tmp"
-merge-bots = ["rust-timer"]
+allowed-merge-apps = ["rust-timer"]
 
 [environments.bors]
 branches = ["automation/bors/auto", "automation/bors/try", "try-perf"]
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -251,15 +251,35 @@ pub enum MergeBot {
     Homu,
     RustTimer,
     Bors,
+    WorkflowsCratesIo,
+}
+
+impl MergeBot {
+    pub fn app_id(&self) -> Option<i64> {
+        match self {
+            MergeBot::WorkflowsCratesIo => Some(2201425),
+            MergeBot::Bors => Some(278306),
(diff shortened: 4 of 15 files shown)


repos/rust-lang/crates.io.toml
```diff
@@ -7,6 +7,21 @@ bots = ["renovate", "rustbot", "heroku-deploy-access"]
...
+[[branch-protections]]
+pattern = "main"
+ci-checks = [
+    "Backend / Lint",
+    "Backend / Test",
+    "Backend / dependencies",
+    "Frontend / Lint",
+    "Frontend / Test",
+]
+required-approvals = 0
```
marcoieni: It was already present in the branch protection. See https://github.com/rust-lang/team/pull/2046

docs/toml-schema.md
```diff
@@ -429,6 +429,19 @@ allowed-merge-teams = ["awesome-team"]
 # When "homu" is used, "bors" has to be in the `bots` array.
 # (optional)
 merge-bots = ["homu"]
+# Whether to enable a merge queue for this branch.
+# (optional - default `false`)
+merge-queue = false
+# GitHub App IDs that are allowed to bypass the branch protection rules.
+# The app ID can be found via the GitHub API (GET /orgs/{org}/installations).
+# (optional)
+bypass-app-ids = [2201425] # workflows-crates-io
```
marcoieni: for simplicity, I put directly the ID, to avoid having another rest API to call GitHub to query for the ID.
Also, this app is only used in crates-io, so I thought adding an enum between app name and ID was overengineering. We can refine this later if needed.

You can see from the screenshot of the issue linked in the PR body that this app is in the bypass list.

sync-team/src/github/api/mod.rs
```diff
@@ -562,12 +562,12 @@ pub(crate) struct RulesetBypassActor {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
-#[serde(rename_all = "snake_case")]
 pub(crate) enum RulesetActorType {
     Integration,
     OrganizationAdmin,
     RepositoryRole,
     Team,
+    DeployKey,
```
marcoieni, line 570: From https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#get-a-repository-ruleset:
Image

repos/rust-lang/crates.io.toml
```diff
@@ -7,7 +7,22 @@ bots = ["renovate", "rustbot", "heroku-deploy-access"]
 [access.teams]
 crates-io = "write"
 
+[[branch-protections]]
+name = "main"
```
marcoieni: adding this to minimize the sync team dry run diff

docs/toml-schema.md
```diff
@@ -429,6 +433,19 @@ allowed-merge-teams = ["awesome-team"]
 # When "homu" is used, "bors" has to be in the `bots` array.
 # (optional)
 merge-bots = ["homu"]
```
marcoieni: @Kobzol one doubt I have: does it makes sense to have `merge-bots` and `bypass-app-ids` separated?

marcoieni:
Here is the dry run diff:
```
Rulesets:
          Updating 'main'
            Include Branches: ["~DEFAULT_BRANCH"] => ["refs/heads/main"]
            Bypass Actors: None => Some([RulesetBypassActor { actor_id: 2201425, actor_type: Integration, bypass_mode: Always }])
```
updating the branch to a fixed one from the default is fine.
The bypass actors are changing, but I wonder if this is because the dry run read token doesn't have the read permission for the bypass actors.

EDIT: AI confirmed that's the case
image

docs/toml-schema.md
Kobzol: Tbh I would rather add the enum variant to document what are all the integrations that we care about (and hardcode the app ID in code, same as we do for bors and Renovatebot).

docs/toml-schema.md
Kobzol: No, I don't think so. I would suggest merging (:laughing:) `merge-bots` and `bypass-app-ids` into `allowed-merge-apps`, to mirror the already existing `allowed-merge-teams`.

sync-team/src/github/mod.rs
```diff
@@ -966,6 +966,16 @@ pub fn construct_ruleset(branch_protection: &rust_team_data::v1::BranchProtectio
 
     let mut rules: Vec<RulesetRule> = Vec::new();
 
+    // Add deletion protection if requested
+    if branch_protection.prevent_deletion {
+        rules.push(RulesetRule::Deletion);
+    }
+
+    // Add non-fast-forward protection if requested
+    if branch_protection.prevent_force_push {
```
Kobzol, line 978: This is not for this PR, but I think that those two (along with forbid creation) should be the default.

docs/toml-schema.md
marcoieni: If I delete `merge-bots` than retrocompatibility becomes a mess. I.e. I need to change all repos files 🤔
Should we do it in another PR?
Should I momentarely add this github app in "merge-bots"? and then rename it in a separate PR?

Kobzol: merge-bots is currently used only on `rust-lang/rust` and nowhere else, so I don't see why we would need to change all repo files :thinking:

rust_team_data/src/v1.rs
```diff
@@ -248,18 +248,36 @@ pub enum BranchProtectionMode {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub enum MergeBot {
-    Homu,
```
marcoieni, line 251: I removed homu because it was unused.
Also I kept the "MergeBot" name here to minimize LoC changed for this PR

marcoieni:
Not sure why after my latest changes the dry run diff increased:

```

    📝 Editing repo 'rust-lang/crates.io':
      Rulesets:
          Updating 'main'
            Include Branches: ["~DEFAULT_BRANCH"] => ["refs/heads/main"]
            Bypass Actors: None => Some([RulesetBypassActor { actor_id: 2201425, actor_type: Integration, bypass_mode: Always }])
            Require code owner review: deleting `false`
            Required status checks: deleting `Backend / Lint (integration_id: 15368), Backend / Test (integration_id: 15368), Backend / dependencies (integration_id: 15368), Frontend / Lint (integration_id: 15368), Frontend / Test (integration_id: 15368)`
            Strict policy for status checks: deleting `false`
            Required approvals: deleting `0`
            Dismiss stale reviews on push: deleting `false`
            Require review thread resolution: deleting `false`
            Require last push approval: deleting `false`
```

marcoieni:
I think AI is right:
image
I'm investigating how I can fix this

repos/rust-lang/rust.toml
```diff
@@ -29,15 +29,18 @@ triage = "write"
 
 [[branch-protections]]
 pattern = "main"
-merge-bots = ["bors"]
+pr-required = false
```
marcoieni: I decided to setup this field explicitly instead of inferring it via code (in case the app is bors).
But let's see the dry run.

sync-team/src/github/mod.rs
```diff
@@ -957,7 +954,7 @@ pub(crate) fn convert_branch_pattern_to_ref_pattern(pattern: &str) -> String {
 pub fn construct_ruleset(branch_protection: &rust_team_data::v1::BranchProtection) -> api::Ruleset {
     use api::*;
 
-    let uses_merge_bot = !branch_protection.merge_bots.is_empty();
+    let uses_merge_bot = !branch_protection.allowed_merge_apps.is_empty();
```
Kobzol: This should be changed to check if `allowed_merge_apps` contains Bors (not just that it is empty). That should fix the diff.

marcoieni: yes, but I thought that having the config more explicit was better instead of this custom logic. Wdyt? 🤔

Kobzol: I think that this logic is fine. The high-level config should be "bors managed this branch", and then `team` should do whatever is necessary to make that happen. Setting `pr-required = false` just exposes (just one) implementation detail of that. So I'd rather just say that bors can push to this branch, and interpret that is it being bors managed, and set the various protection options accordingly.

marcoieni:
The dry run looks good now:
```

    📝 Editing repo 'rust-lang/crates.io':
      Rulesets:
          Updating 'main'
            Include Branches: ["~DEFAULT_BRANCH"] => ["refs/heads/main"]
            Bypass Actors: None => Some([RulesetBypassActor { actor_id: 2201425, actor_type: Integration, bypass_mode: Always }])
```

sync-team/src/github/mod.rs
marcoieni: ok, I'll change this

rust_team_data/src/v1.rs
```diff
@@ -248,18 +248,36 @@ pub enum BranchProtectionMode {
...
     RustTimer,
     Bors,
+    WorkflowsCratesIo,
```
Kobzol, line 254: This will break triagebot. I think that we should split this PR into two, make the first backwards-compatible by just adding the new code, then merge it, update triagebot and only then start using the new variants.

rust_team_data/src/v1.rs
```diff
@@ -248,18 +248,36 @@ pub enum BranchProtectionMode {
...
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct BranchProtection {
     pub pattern: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
     pub dismiss_stale_review: bool,
     pub mode: BranchProtectionMode,
     pub allowed_merge_teams: Vec<String>,
-    pub merge_bots: Vec<MergeBot>,
+    pub allowed_merge_apps: Vec<MergeBot>,
```
Kobzol, line 277: Kind of a meta-point: I wonder if we should stop using the V1 API in sync-team. We often break the API by making changes that are *only* required for sync-team, but now that we have sync-team in-tree, we could just directly work on the TOML representation in it. Other tools don't need to know about allowed merge apps or even branch protections.

marcoieni: I don't know how other tools use the API. From me, +1 on anything that makes this repo simpler 👍

rust_team_data/src/v1.rs
marcoieni: is https://github.com/rust-lang/team/pull/2304/changes/5ec5088287bc89dad508a5013f4baa924ba0f61a what you had in mind?

Kobzol: It's not enough, because after this PR gets merged as-is, the API will return the crates-io-workflows variant of this enum, and triagebot will fail to deserialize it.

Kobzol (approved):

ubiratansoares (approved):
Awesome work here!

PR #2319, opened 2026-03-13
use BTreeSet for ruleset rules

marcoieni:
This PR fixes the empty sync-team diff we see in PRs

### DIFF
--- a/src/sync/github/api/mod.rs
+++ b/src/sync/github/api/mod.rs
@@ -17,6 +17,7 @@ use reqwest::{
 };
 use secrecy::ExposeSecret;
 use serde::{Deserialize, de::DeserializeOwned};
+use std::collections::BTreeSet;
 use std::fmt;
 use thiserror::Error;
 use tokens::GitHubTokens;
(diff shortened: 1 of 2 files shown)


ubiratansoares (approved):

Kobzol (approved):
I was thinking about this when examining the implementation; I hoped that GitHub keeps the order. Seems not, so keeping it sorted it is.

Even better would be to actually represent the rules as fields, rather than in a container, because we use at most one rule per type anyway. But that's a more complex change.

marcoieni:
It failed with the following error:

```
    📝 Editing repo 'rust-lang/bors':
      Rulesets:
          Updating 'Ruleset for main'
            Bypass Actors: Some([]) => None
    📝 Editing repo 'rust-lang/thorin':
      Branch Protections:
          main
            Dismiss Stale Reviews: false
            Is admin enforced: true
            Required Approving Review Count: 1
            Requires PR: true
            Required Checks: []
            Allowances: []
    📝 Editing repo 'rust-lang/triagebot':
      Branch Protections:
          master
            Dismiss Stale Reviews: false
            Is admin enforced: true
            Required Approving Review Count: 1
            Requires PR: true
            Required Checks: ["ci"]
            Allowances: []
    
Error: rust_team] failed: graphql error: Name already protected: main
```

marcoieni:
I'll try again to merge this. Maybe some flakiness from github..

PR #2323, opened 2026-03-16
move rustfmt to GitHub rulesets

marcoieni:
rustfmt has multiple branch protections, so it's interesting to see if `team` handles this correctly.

See https://github.com/rust-lang/team/blob/main/repos/rust-lang/rustfmt.toml

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -85,4 +85,5 @@ members-without-zulip-id = [
 enable-rulesets-repos = [
    "rust-lang/bors",
    "rust-lang/crates.io",
+   "rust-lang/rustfmt",
 ]
--- a/repos/rust-lang/rustfmt.toml
+++ b/repos/rust-lang/rustfmt.toml
@@ -37,4 +37,3 @@ required-approvals = 0
 
 [environments.github-pages]
 branches = ["main"]
-


marcoieni:
The `"libsyntax"` and `"rust-1.*"` have the following:
github com_rust-lang_rustfmt_settings_branch_protection_rules_891286

While `"main"` also allows force pushes.
github com_rust-lang_rustfmt_settings_branch_protection_rules_891286 (1)

marcoieni:
In the last commit, I edited the rustfmt branch protection to reflect the current settings

ubiratansoares:
> In the last commit, I edited the rustfmt branch protection to reflect the current settings

Should one expect something like

```
Forbid force pushes: no 
```

in a dry-run?

marcoieni:
We only record that rule when it's present:

https://github.com/rust-lang/team/blob/fe01a62618672a94edf4475a37a79e83cf107e37/src/sync/github/mod.rs#L1823

Notice that the other two rulesets dry-run have `Forbid force pushes: yes`.

marcoieni:
I just removed allow force push from the main branch after talking to the maintainer, so I updated this branch accordingly
image

ubiratansoares (approved):

PR #2325, opened 2026-03-16
simplify default ruleset name

marcoieni:
I don't like this default prefix in ruleset names. It's verbose and adds no extra information.
image


repos/rust-lang/crates.io.toml
```diff
@@ -8,7 +8,6 @@ bots = ["renovate", "rustbot", "heroku-deploy-access"]
 crates-io = "write"
 
 [[branch-protections]]
-name = "main"
```
marcoieni, line 11: this is no longer needed as it will be the default name

ubiratansoares (approved):
I prefer the shorter naming too 😄

PR #2326, opened 2026-03-16
publish libm via trusted publishing

marcoieni:


repos/rust-lang/compiler-builtins.toml
```diff
@@ -16,3 +16,9 @@ branches = ["main"]
 
 [environments.publish]
 branches = ["main"]
+
+[[crates-io]]
+crates = ["libm"]
+publish-workflow = "publish.yaml"
+publish-environment = "publish"
+teams = ["libs"]
```
marcoieni, line 24: libs team is an owner right now:

Image

repos/rust-lang/compiler-builtins.toml
```diff
@@ -16,3 +16,9 @@ branches = ["main"]
 
 [environments.publish]
 branches = ["main"]
+
+[[crates-io]]
+crates = ["libm"]
+publish-workflow = "publish.yaml"
```
marcoieni, line 22: https://github.com/rust-lang/compiler-builtins/blob/main/.github/workflows/publish.yaml

repos/rust-lang/compiler-builtins.toml
```diff
@@ -16,3 +16,9 @@ branches = ["main"]
 
 [environments.publish]
 branches = ["main"]
+
+[[crates-io]]
+crates = ["libm"]
```
marcoieni, line 21: This is the only public crate of the repo. You can verify by running this script:

https://github.com/rust-lang/team/blob/53e840af7c109c7148190abbd69a80ccbe69f653/docs/toml-schema.md#L494-L495

repos/rust-lang/compiler-builtins.toml
tgross35: What exactly does this key do by the way?

tgross35 (approved):

repos/rust-lang/compiler-builtins.toml
ubiratansoares: [non-blocking] Newbie question : do we validate whether the Workflow file exists? Asking mostly due to typo `publish.yaml` with `publish.yml` or similar, but not sure whether this breaks a link/validation somewhere else, hence we could check it upfront

ubiratansoares (approved):

repos/rust-lang/compiler-builtins.toml
Kobzol: We don't, crates.io validates it only from their frontend, I think.

repos/rust-lang/compiler-builtins.toml
Kobzol: It makes the `libs` team an owner of the crate on crates.io. Which means that they can still manually do yanks without going through an admin or CI (new versions are only deployable through CI though).

tgross35: Oh, does setting up trusted publishing here take over what's currently on crates.io? Meaning only rust-lang-owner and rust-lang/libs will have permissions?

(Which is fine, Jorge hasn't been involved in a while)

Kobzol: I think we only sync team owners, not user owners.

PR #2327, opened 2026-03-16
use rulesets in the rust repo

marcoieni:
Is there a preferred moment when we want to merge this?

git SHA of each branch before the switch:
* main: https://github.com/rust-lang/rust/commit/d12e1e12ae6fb828df1b663a07c9191bc48c6d6d
* automation/bors/auto: https://github.com/rust-lang/rust/commit/91367b0f73196f0d133066e0e9363da339ed87b5
* automation/bors/auto-merge: https://github.com/rust-lang/rust/commit/08ab4b28fd673cc431c0705e55a35b21eacf759d
* automation/bors/try: https://github.com/rust-lang/rust/commit/6bcaf55b079c0275f05466854848708f6e64319d
* automation/bors/try-merge: https://github.com/rust-lang/rust/commit/806804d8fd8d2ec4220af77bb77799fd0f7d8c54
* beta: https://github.com/rust-lang/rust/commit/8489f1adbe1c8f8921cf18b9962cf565237a5489
* cargo_update: https://github.com/rust-lang/rust/commit/b8271c3270f185f95f03843aca43f3bcec69e33e
* perf-tmp: https://github.com/rust-lang/rust/commit/a02fdc12a62a0b0bf55887ab916efae4d96f80ab
* stable: https://github.com/rust-lang/rust/commit/59807616e1fa2540724bfbac14d7976d7e4a3860
* try-perf: https://github.com/rust-lang/rust/commit/a02fdc12a62a0b0bf55887ab916efae4d96f80ab

### DIFF
--- a/repos/rust-lang/rust.toml
+++ b/repos/rust-lang/rust.toml
@@ -26,67 +26,127 @@ rust-analyzer = "write"
 style = "write"
 types = "write"
 triage = "write"
+# needed for the branch protection
+rust-timer = "write"
 
+# Only `bors` can push to `main`
 [[branch-protections]]
+name = "main"
 pattern = "main"
 allowed-merge-apps = ["bors"]
+prevent-update = true
 
+# No one can force push to main
 [[branch-protections]]
+name = "main - force-push"
+pattern = "main"
+# `main` already requires users to open PRs.
+# This allows bors to push without opening a PR.
+pr-required = false
+
+# Only `bors` and `promote-release` can push to `stable`
+[[branch-protections]]
+name = "stable"
 pattern = "stable"
-allowed-merge-apps = ["bors"]
+allowed-merge-apps = ["bors", "promote-release"]
+prevent-update = true
 
+# Only `promote-release` can force-push to `stable`
 [[branch-protections]]
+name = "stable - force-push"
+pattern = "stable"
+allowed-merge-apps = ["promote-release"]
+# `stable` already requires users to open PRs.
+# This allows bors to push without opening a PR.
+pr-required = false
+
+# No one create or delete the stable branch.
+[[branch-protections]]
+name = "stable - misc"
+pattern = "stable"
+pr-required = false
+prevent-force-push = false
+
+# Only `bors` and `promote-release` can push to `beta`
+[[branch-protections]]
+name = "beta"
 pattern = "beta"
-allowed-merge-apps = ["bors"]
+allowed-merge-apps = ["bors", "promote-release"]
+prevent-update = true
+
+# Only `promote-release` can force-push to `beta`
+[[branch-protections]]
+name = "beta - force-push"
+pattern = "beta"
+allowed-merge-apps = ["promote-release"]
+# `beta` already requires users to open PRs.
+# This allows bors to push without opening a PR.
+pr-required = false
+
+# no one can create or delete the beta branch.
+[[branch-protections]]
+name = "beta - misc"
+pattern = "beta"
+pr-required = false
+prevent-force-push = false
 
 [[branch-protections]]
 pattern = "*"
+allowed-merge-apps = ["promote-release"]
+prevent-update = true
 
 [[branch-protections]]
 pattern = "*/**/*"
 pr-required = false
+prevent-update = true
 
 [[branch-protections]]
 pattern = "cargo_update"
 pr-required = false
+prevent-deletion = false
+prevent-force-push = false
 
 # Required for running try builds created by bors.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "automation/bors/try"
 allowed-merge-apps = ["bors"]
+prevent-update = true
 
 # Required for running try builds created by bors.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "automation/bors/try-merge"
 allowed-merge-apps = ["bors"]
+prevent-update = true
 
 # Required for running auto builds created by bors.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "automation/bors/auto"
 allowed-merge-apps = ["bors"]
+prevent-update = true
 
 # Required for running auto builds created by bors.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "automation/bors/auto-merge"
 allowed-merge-apps = ["bors"]
+prevent-update = true
 
 # Required for unrolled PR builds created by perfbot.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "try-perf"
-allowed-merge-apps = ["rust-timer"]
-pr-required = false
+allowed-merge-teams = ["rust-timer"]
+prevent-update = true
 
 # Required for unrolled PR builds created by perfbot.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "perf-tmp"
-allowed-merge-apps = ["rust-timer"]
-pr-required = false
+allowed-merge-teams = ["rust-timer"]
+prevent-update = true
 
 [environments.bors]
 branches = ["automation/bors/auto", "automation/bors/try", "try-perf"]
--- a/src/sync/github/mod.rs
+++ b/src/sync/github/mod.rs
@@ -9,6 +9,7 @@ use crate::sync::Config;
 use crate::sync::github::api::{
     GithubRead, Login, PushAllowanceActor, RepoPermission, RepoSettings, Ruleset,
 };
+use anyhow::Context as _;
 use futures_util::StreamExt;
 use log::debug;
 use rust_team_data::v1::{Bot, BranchProtectionMode, MergeBot, ProtectionTarget};
@@ -447,6 +448,78 @@ impl SyncGitHub {
         !self.config.disable_rulesets_repos.contains(&repo_full_name)
     }
 
+    async fn construct_ruleset(
+        &self,
+        expected_repo: &rust_team_data::v1::Repo,
+        branch_protection: &rust_team_data::v1::BranchProtection,
+    ) -> anyhow::Result<api::Ruleset> {
+        let bypass_actors = self.bypass_actors(expected_repo, branch_protection).await?;
+
+        Ok(construct_ruleset(branch_protection, bypass_actors))
+    }
+
+    async fn bypass_actors(
+        &self,
+        expected_repo: &rust_team_data::v1::Repo,
+        branch_protection: &rust_team_data::v1::BranchProtection,
+    ) -> Result<Vec<api::RulesetBypassActor>, anyhow::Error> {
+        use api::{RulesetActorType, RulesetBypassActor, RulesetBypassMode};
+
+        let mut bypass_actors = Vec::new();
+        let allowed_teams = self
+            .allowed_merge_teams(expected_repo, branch_protection)
+            .await?;
+        bypass_actors.extend(allowed_teams);
+        let allowed_apps = branch_protection
+            .allowed_merge_apps
+            .iter()
+            .filter_map(|app| {
+                app.app_id().map(|app_id| RulesetBypassActor {
+                    actor_id: app_id,
+                    actor_type: RulesetActorType::Integration,
+                    bypass_mode: RulesetBypassMode::Always,
+                })
+            });
+        bypass_actors.extend(allowed_apps);
+        Ok(bypass_actors)
+    }
+
+    async fn allowed_merge_teams(
+        &self,
+        expected_repo: &rust_team_data::v1::Repo,
+        branch_protection: &rust_team_data::v1::BranchProtection,
+    ) -> Result<Vec<api::RulesetBypassActor>, anyhow::Error> {
+        use api::{RulesetActorType, RulesetBypassActor, RulesetBypassMode};
+
+        let mut allowed = vec![];
+
+        for team_name in &branch_protection.allowed_merge_teams {
+            let github_team = self
+                .github
+                .team(&expected_repo.org, team_name)
+                .await?
+                .with_context(|| {
+                    format!(
+                        "failed to find GitHub team '{team_name}' in org '{}' for repo '{}/{}'",
+                        expected_repo.org, expected_repo.org, expected_repo.name
+                    )
+                })?;
+            let team_id = github_team.id.with_context(|| {
+                format!(
+                    "GitHub team '{team_name}' in org '{}' is missing an ID",
+                    expected_repo.org
+                )
+            })?;
+
+            allowed.push(RulesetBypassActor {
+                actor_id: team_id as i64,
+                actor_type: RulesetActorType::Team,
+                bypass_mode: RulesetBypassMode::Always,
+            });
+        }
+        Ok(allowed)
+    }
+
     async fn diff_repo(
         &self,
         expected_repo: &rust_team_data::v1::Repo,
@@ -481,7 +554,9 @@ impl SyncGitHub {
                 let use_rulesets = self.should_use_rulesets(expected_repo);
                 if use_rulesets {
                     for branch_protection in &expected_repo.branch_protections {
-                        let ruleset = construct_ruleset(branch_protection);
+                        let ruleset = self
+                            .construct_ruleset(expected_repo, branch_protection)
+                            .await?;
                         rulesets.push(ruleset);
                     }
                 }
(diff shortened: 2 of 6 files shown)


marcoieni:
Backup (taken automatically with https://github.com/marcoieni/multicheese in case you are curious!)

001-github-com-rust-lang-rust-settings-branch-protection-rules-2097960
002-github-com-rust-lang-rust-settings-branch-protection-rules-2626273
003-github-com-rust-lang-rust-settings-branch-protection-rules-29679214
004-github-com-rust-lang-rust-settings-branch-protection-rules-31895354
005-github-com-rust-lang-rust-settings-branch-protection-rules-38663771
006-github-com-rust-lang-rust-settings-branch-protection-rules-42228944
007-github-com-rust-lang-rust-settings-branch-protection-rules-42230325
008-github-com-rust-lang-rust-settings-branch-protection-rules-63047047
009-github-com-rust-lang-rust-settings-branch-protection-rules-63047048
010-github-com-rust-lang-rust-settings-branch-protection-rules-69583533
011-github-com-rust-lang-rust-settings-branch-protection-rules-71071517

src/validate.rs
```diff
@@ -1150,18 +1150,7 @@ fn validate_branch_protections(data: &Data, errors: &mut Vec<String>) {
...
         for protection in &repo.branch_protections {
-            if !patterns.insert((protection.target, &protection.pattern)) {
-                bail!(
-                    r#"repo '{}' uses multiple {:?} protections with the pattern `{}`"#,
-                    repo.name,
-                    protection.target,
-                    protection.pattern,
-                );
-            }
-
```
marcoieni, line 1213: From https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets#about-rulesets-and-protected-branches:

> Unlike protection rules, multiple rulesets can apply at the same time, so you can be confident that every rule targeting a branch in your repository will be evaluated when someone interacts with that branch

I removed this because we need multiple rules to interact with the same branch (different bypass list for normal push and force push).

Kobzol: In that case we should make the name required when there are multiple identical patterns, because we used the pattern as an identifier.

marcoieni:
Here's how I tested the "stable" and "stable - force-pushes" rules.

- I created a repository in my test organization: https://github.com/marco-test-org/ruleset-test
- I created two rules
image
github com_marco-test-org_ruleset-test_settings_rules_14250638
github com_marco-test-org_ruleset-test_settings_rules_14250638 (1)

The user in the `test-team` was able to push to the repository but not to force-push. See https://github.com/marco-test-org/ruleset-test/issues/1

marcoieni:
There's one issue:
In the `try-perf` branch, the user [rust-timer](https://github.com/rust-timer) is allowed to push.
In rulesets, only teams and github apps can be allowed to push.
Should we create a team for rust-timer?
Another approach would be converting rust-timer to a github app but I'm not sure how difficult it is.

Kobzol:
We are relatively close to making the unrolled PRs by bors, instead of rustc-perf. If this can wait a few weeks, rust-timer will no longer need that access.

marcoieni:
> We are relatively close to making the unrolled PRs by bors, instead of rustc-perf. If this can wait a few weeks, rust-timer will no longer need that access.

https://github.com/rust-lang/team/pull/2343 is a small change that can be reverted easily. Since this is the only blocker to have rulesets in the rust I'm in favor of creating the team. What do you think? 🤔

Kobzol:
Sure, go ahead, it's a small change 👍

src/sync/github/mod.rs
```diff
@@ -370,6 +371,78 @@ impl SyncGitHub {
...
 
+    async fn construct_ruleset(
+        &self,
+        expected_repo: &rust_team_data::v1::Repo,
+        branch_protection: &rust_team_data::v1::BranchProtection,
+    ) -> anyhow::Result<api::Ruleset> {
+        let bypass_actors = self.bypass_actors(expected_repo, branch_protection).await?;
+
+        Ok(construct_ruleset(branch_protection, bypass_actors))
+    }
```
marcoieni, line 459: this function is only two lines but I left it like this to minimize the git diff of the PR.
We could inline the construct_ruleset free function later

src/sync/github/mod.rs
```diff
@@ -1083,19 +1163,6 @@ pub fn construct_ruleset(branch_protection: &rust_team_data::v1::BranchProtectio
...
-        .allowed_merge_apps
-        .iter()
-        .filter_map(|app| {
-            app.app_id().map(|app_id| RulesetBypassActor {
-                actor_id: app_id,
-                actor_type: RulesetActorType::Integration,
-                bypass_mode: RulesetBypassMode::Always,
-            })
-        })
-        .collect();
```
marcoieni, line 1285: this piece of code was moved up

src/validate.rs
```diff
@@ -1166,17 +1176,15 @@ fn validate_branch_protections(data: &Data, errors: &mut Vec<String>) {
                 let key = (repo.org.clone(), team.clone());
                 if !github_teams.contains(&key) {
                     bail!(
-                        r#"repo '{}' uses a branch protection for {} that mentions the '{}' github team;
-but that team does not seem to exist"#,
+                        r#"repo '{}' uses a branch protection for {} that mentions the '{}' github team; but that team does not seem to exist"#,
```
marcoieni, line 1228: I put this error and the next one on the same line to improve how they look in github actions.

repos/rust-lang/rust.toml
```diff
@@ -26,65 +26,105 @@ rust-analyzer = "write"
 style = "write"
 types = "write"
 triage = "write"
+# needed for the branch protection
+rust-timer = "write"
 
+# Only `bors` can push to `main`
 [[branch-protections]]
```
Kobzol, line 33: I don't understand why we need two separate branch protections to prevent force pushes. Force pushes should be disabled for everyone, even for bors.

repos/rust-lang/rust.toml
```diff
@@ -26,65 +26,105 @@ rust-analyzer = "write"
...
+allowed-merge-teams = ["rust-timer"]
+prevent-update = true
 
 # Required for unrolled PR builds created by perfbot.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "perf-tmp"
-allowed-merge-apps = ["rust-timer"]
+allowed-merge-teams = ["rust-timer"]
+prevent-update = true
```
Kobzol, line 149: What does `prevent-update` actually do? Any why isn't it the default?

marcoieni: From https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/available-rules-for-rulesets#restrict-updates:

> If selected, only users with bypass permissions can push to branches or tags whose name matches the pattern you specify.

At the moment it's not enabled by default. I will raise a PR to make it the default.

Kobzol: Hmm, it seems that the mental model of branch protections and rulesets quite diverges here. I guess that `prevent-update = true` should be automatically set by `pr-required = true`?

marcoieni: Yeah the fact that you both have "restrict updates" and "require a pull request before merging" as possible rules is weird. I think it makes sense for tags, but still..
I will raise a PR

marcoieni: I think we can resolve this after https://github.com/rust-lang/team/pull/2346 and https://github.com/rust-lang/team/pull/2348

repos/rust-lang/rust.toml
marcoieni: > Force pushes should be disabled for everyone, even for bors.

That's what the `main - force-push` rule does. It prevents everyone (including bors) to force push.
Bors is in the `main` rule, so that rule doesn't apply to bors. To restrict bors, we need another rule where bors isn't in the bypass list.

Kobzol: Ah, I see, I forgot about the other branch protection. That's quite stupid that rulesets work like this, tbh 😆  It would be better if we could bypass only specific properties.. but I understand that would make the options more complicated.

It's quite non-obvious though, and it seems hard to ensure that the other branch protection don't enable something that we don't want, hmm.

marcoieni: > It's quite non-obvious though, and it seems hard to ensure that the other branch protection don't enable something that we don't want, hmm.

Yeah, it requires a different mindset than before a bit.

Kobzol: Wait, but if the force push ruleset requires a PR by default, doesn't that mean that it will require a PR also for bors?

marcoieni:
We added "allow force push" to the `cargo_update` branch. See [#t-infra > cargo_update branch protection broken](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/cargo_update.20branch.20protection.20broken/with/582397089)
image

repos/rust-lang/rust.toml
marcoieni: yes, you are right, in my test I didn't have "requires a PR" in the settings.
Probably we should add a third rule related to main 😅

Kobzol: Not a third rule, just disable requiring PRs here.

marcoieni: With "here" you mean in the "main - force push"?
I think you are right! 
Because "main" already prevent everyone from opening PRs. 
I'll do it.

marcoieni: I did the same for stable and beta as well.

ubiratansoares (dismissed):
I've checked all screenshots and compared side-by-side with the new config. Looks Ok to me. Perhaps @Kobzol wants to confirm whether we are good to go from his side as well.

I think it'd be nice spreading the word about this PR on [T-infra/annoucements](https://rust-lang.zulipchat.com/#narrow/channel/533458-t-infra.2Fannouncements).

repos/rust-lang/rust.toml
```diff
@@ -26,66 +26,114 @@ rust-analyzer = "write"
...
 triage = "write"
+# needed for the branch protection
+rust-timer = "write"
 
+# Only `bors` can push to `main`
 [[branch-protections]]
+name = "main"
 pattern = "main"
 allowed-merge-apps = ["bors"]
+prevent-update = true
```
Kobzol, line 37: I realized that to mirror the previous behavior of branch protections, having a non-empty list of `allowed-merge-apps` should probably imply `prevent-update = true`. But that's non-blocking, we can do that later.

repos/rust-lang/rust.toml
```diff
@@ -26,66 +26,114 @@ rust-analyzer = "write"
...
+allowed-merge-apps = ["bors", "promote-release"]
+prevent-update = true
+
+# Only `promote-release` can force-push to `stable`
+[[branch-protections]]
+name = "stable - force-push"
+pattern = "stable"
+allowed-merge-apps = ["promote-release"]
+# `stable` already requires users to open PRs.
+# This allows bors to push without opening a PR.
```
Kobzol, line 60:
```suggestion
# This allows promote-release to push without opening a PR.
```

repos/rust-lang/rust.toml
```diff
@@ -26,66 +26,114 @@ rust-analyzer = "write"
...
+name = "beta - force-push"
+pattern = "beta"
+allowed-merge-apps = ["promote-release"]
+# `beta` already requires users to open PRs.
+# This allows bors to push without opening a PR.
+pr-required = false
 
 [[branch-protections]]
 pattern = "*"
+allowed-merge-apps = ["promote-release"]
```
Kobzol, line 95: Why promote release? For tags? In that case, we could mark this protection as being only for tags.

repos/rust-lang/rust.toml
```diff
@@ -26,66 +26,114 @@ rust-analyzer = "write"
...
 allowed-merge-apps = ["bors"]
+prevent-update = true
 
 # Required for unrolled PR builds created by perfbot.
 # Must support force-pushes.
 [[branch-protections]]
 pattern = "try-perf"
-allowed-merge-apps = ["rust-timer"]
+allowed-merge-teams = ["rust-timer"]
 pr-required = false
```
Kobzol: This also requires `prevent-update = true`, otherwise anyone with `write` access could access S3 data. Same for `perf-tmp`.

repos/rust-lang/rust.toml
marcoieni: I don't know, I'm just mirroring 1-to-1 the existing branch protection. I don't think it applies to tags.
But if you think I should add a rule for tags, I can add it

repos/rust-lang/rust.toml
marcoieni: no, I meant bors!
The `stable - force-push` branch protection is there to restrict the behavior of `bors`.
`promote-release` doesn't have any restrictions because it is in all the allowed lists, so it could even delete the branch.

repos/rust-lang/rust.toml
Kobzol: Ok then let's keep it as-is.

src/sync/github/mod.rs
```diff
@@ -1179,11 +1256,12 @@ fn github_int(value: u32) -> i32 {
...
 }
 
-pub fn construct_ruleset(branch_protection: &rust_team_data::v1::BranchProtection) -> api::Ruleset {
+pub fn construct_ruleset(
+    branch_protection: &rust_team_data::v1::BranchProtection,
+    bypass_actors: Vec<api::RulesetBypassActor>,
+) -> api::Ruleset {
     use api::*;
 
-    let branch_protection_mode = get_branch_protection_mode(branch_protection);
```
marcoieni, line 1185: As noted by Kobzol:

> we shouldn't turn off requiring PRs when bors is enabled, because that will turn off required PRs for everyone

repos/rust-lang/rust.toml
```diff
@@ -26,67 +26,131 @@ rust-analyzer = "write"
...
```
Kobzol:
```suggestion

```
Shouldn't be needed, as rust-timer bypasses this ruleset anyway (same below).

repos/rust-lang/rust.toml
```diff
@@ -26,67 +26,131 @@ rust-analyzer = "write"
...
+
+# no one can create or delete the beta branch.
+[[branch-protections]]
+name = "beta - misc"
+pattern = "beta"
+pr-required = false
+prevent-force-push = false
 
 [[branch-protections]]
 pattern = "*"
```
Kobzol, line 94: We have to check whether this still protects tags (like `1.94.1`).

repos/rust-lang/rust.toml
marcoieni: removed!

repos/rust-lang/rust.toml
marcoieni: it doesn't. Let's do a followup PR

repos/rust-lang/rust.toml
```diff
@@ -26,67 +26,127 @@ rust-analyzer = "write"
...
+allowed-merge-apps = ["promote-release"]
+prevent-update = true
```
marcoieni, line 96: the classic branch protection allows deletion, while this doesn't.
Jakub [here](https://github.com/rust-lang/team/commit/d9ec7fad66254a516af499b7fa98a20803a70edd#r181727275) says he prefers to prevent deletion for this and `"*/**/*"`

Kobzol: Hmm, I didn't realize deletion was allowed before. I don't think it's super important, but I guess we can try to be more strict and see if preventing deletion breaks anything.

ubiratansoares (approved):
From my end, I think we are good to go.

marcoieni:
New rules created

image

marcoieni:
Here are the rules that were created:

001-github-com-rust-lang-rust-settings-rules-15300153
002-github-com-rust-lang-rust-settings-rules-15300154
003-github-com-rust-lang-rust-settings-rules-15300160
004-github-com-rust-lang-rust-settings-rules-15300161
005-github-com-rust-lang-rust-settings-rules-15300156
006-github-com-rust-lang-rust-settings-rules-15300158
007-github-com-rust-lang-rust-settings-rules-15300148
008-github-com-rust-lang-rust-settings-rules-15300150
009-github-com-rust-lang-rust-settings-rules-15300151
010-github-com-rust-lang-rust-settings-rules-15300155
011-github-com-rust-lang-rust-settings-rules-15300143
012-github-com-rust-lang-rust-settings-rules-15300144
013-github-com-rust-lang-rust-settings-rules-15300163
014-github-com-rust-lang-rust-settings-rules-15300145
015-github-com-rust-lang-rust-settings-rules-15300146
016-github-com-rust-lang-rust-settings-rules-15300147
017-github-com-rust-lang-rust-settings-rules-15300162

Issue #2196, opened 2025-12-22
Discuss merge queue default settings


marcoieni:
<img width="1612" height="1614" alt="image" src="https://github.com/user-attachments/assets/42c1f690-a1b7-4930-a95e-e4c457a1bbeb" />

These are the default values of the merge queue if I try to create it from the UI. Why diverging from these values when we create a merge queue? See:

https://github.com/rust-lang/team/blob/04e7f0028d407c5d0938b49ed50d8e016e13f1a0/sync-team/src/github/mod.rs#L954-L960

For each parameter, write a comment if they diverge from the default value explaining why.

Note that merge queue support was added in https://github.com/rust-lang/team/pull/2168

PR #2333, opened 2026-03-18
prepare for moving cargo from branch protections to rulesets

marcoieni:
I ran a script to detect existing rulesets in our orgs. `rust-lang/cargo` is the only repository that has a ruleset not tracked in this repo.

This PR:
- Adds this ruleset (`"Only allow the release process to publish tags"`) to the `team` repository. You can verify that the dry run doesn't mention this repo, so we are just writing down without editing the ruleset in github
- Migrate `cargo` from classic branch protections to rulesets
- Improve the logging of rulesets, so that only relevant fields are showed in the dry run (default values are omitted)

In the following, I post the screenshots of the resources this PR touches:
- as a backup in case something goes wrong
- to better review the dry run

- [ ] For triagebot retrocompatibility we should first merge the code changes, update triagebot and then edit `cargo.toml`. After this PR is approved, I will comment the changes to `cargo.toml` before merging this.

### `Master`
github com_rust-lang_cargo_settings_branch_protection_rules_593520

### `rust-1.*`

(download to see the full picture)

github com_rust-lang_cargo_settings_branch_protection_rules_593520 (2)

## Existing ruleset
github com_rust-lang_cargo_settings_rules_1162109

### DIFF
--- a/repos/rust-lang/cargo.toml
+++ b/repos/rust-lang/cargo.toml
@@ -10,13 +10,21 @@ cargo = "write"
 [[branch-protections]]
 pattern = "master"
 ci-checks = ["conclusion"]
+# merge-queue = true
 
 [[branch-protections]]
 pattern = "rust-1.*"
 ci-checks = ["conclusion"]
 
+# [[branch-protections]]
+# name = "Only allow the release process to publish tags"
+# pattern = "0.*"
+# target = "tag"
+# pr-required = false
+# allowed-merge-apps = ["promote-release"]
+# prevent-update = true
+
 [environments.github-pages]
 
 [environments.release]
 tags = ["*"]
-
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -252,6 +252,7 @@ pub enum MergeBot {
     RustTimer,
     Bors,
     WorkflowsCratesIo,
+    PromoteRelease,
 }
 
 impl MergeBot {
@@ -259,15 +260,26 @@ impl MergeBot {
         match self {
             MergeBot::WorkflowsCratesIo => Some(2201425),
             MergeBot::Bors => Some(278306),
+            MergeBot::PromoteRelease => Some(217112),
             // These are user-based bots, not GitHub Apps
             MergeBot::RustTimer | MergeBot::Homu => None,
         }
     }
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum ProtectionTarget {
+    #[default]
+    Branch,
+    Tag,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct BranchProtection {
     pub pattern: String,
+    #[serde(default, skip_serializing_if = "is_branch_target")]
+    pub target: ProtectionTarget,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub name: Option<String>,
     pub dismiss_stale_review: bool,
@@ -276,11 +288,9 @@ pub struct BranchProtection {
     pub merge_bots: Vec<MergeBot>,
     pub allowed_merge_apps: Vec<MergeBot>,
     pub merge_queue: bool,
-    #[serde(default = "default_true")]
     pub prevent_creation: bool,
-    #[serde(default = "default_true")]
+    pub prevent_update: bool,
     pub prevent_deletion: bool,
-    #[serde(default = "default_true")]
     pub prevent_force_push: bool,
 }
 
@@ -312,6 +322,6 @@ pub struct People {
     pub people: IndexMap<String, Person>,
 }
 
-fn default_true() -> bool {
-    true
+fn is_branch_target(target: &ProtectionTarget) -> bool {
+    matches!(target, ProtectionTarget::Branch)
 }
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -861,12 +861,23 @@ pub(crate) enum AllowedMergeApp {
     RustTimer,
     Bors,
     WorkflowsCratesIo,
+    PromoteRelease,
+}
+
+#[derive(serde::Deserialize, Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) enum ProtectionTarget {
+    #[default]
+    Branch,
+    Tag,
 }
 
 #[derive(serde::Deserialize, Debug)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
 pub(crate) struct BranchProtection {
     pub pattern: String,
+    #[serde(default)]
+    pub target: ProtectionTarget,
     #[serde(default)]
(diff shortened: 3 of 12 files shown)


repos/rust-lang/cargo.toml
```diff
@@ -10,13 +10,21 @@ cargo = "write"
 [[branch-protections]]
 pattern = "master"
 ci-checks = ["conclusion"]
+merge-queue = true
```
marcoieni: merge queues in branch protections were enabled from the UI because the API only support merge queues in rulesets.

repos/rust-lang/cargo.toml
```diff
@@ -10,13 +10,21 @@ cargo = "write"
...
 ci-checks = ["conclusion"]
 
+[[branch-protections]]
+name = "Only allow the release process to publish tags"
+pattern = "0.*"
+target = "tag"
+pr-required = false
+allowed-merge-apps = ["promote-release"]
+prevent-update = true
+
```
marcoieni, line 26: this is the ruleset that already existed. I just imported the settings here.

rust_team_data/src/v1.rs
```diff
@@ -276,11 +288,16 @@ pub struct BranchProtection {
     pub merge_bots: Vec<MergeBot>,
     pub allowed_merge_apps: Vec<MergeBot>,
     pub merge_queue: bool,
-    #[serde(default = "default_true")]
+    #[serde(default = "branch_protection_default_prevent_creation")]
```
marcoieni: the fact that `prevent_creation` "was true by default" was encoded in too many places in the codebase. So I created functions to be reused across the codebase.
Similar with the other fields.

rust_team_data/src/v1.rs
```diff
@@ -276,11 +288,16 @@ pub struct BranchProtection {
...
     pub prevent_creation: bool,
-    #[serde(default = "default_true")]
+    #[serde(
+        default = "branch_protection_default_prevent_update",
+        skip_serializing_if = "is_false"
+    )]
+    pub prevent_update: bool,
```
marcoieni, line 292: I had to add this field because cargo needs it.

src/sync/github/api/mod.rs
marcoieni, line 1: I edited this file to encode the default of various fields, so that we know when to skip logging a rule

rust_team_data/src/v1.rs
```diff
@@ -252,22 +252,34 @@ pub enum MergeBot {
...
+    #[default]
+    Branch,
+    Tag,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct BranchProtection {
     pub pattern: String,
+    #[serde(default, skip_serializing_if = "is_branch_target")]
+    pub target: ProtectionTarget,
```
marcoieni, line 282: I had to add this field because the existing cargo ruleset targets a tag

src/validate.rs
```diff
@@ -1145,10 +1145,11 @@ fn validate_branch_protections(data: &Data, errors: &mut Vec<String>) {
         let mut patterns = HashSet::new();
 
         for protection in &repo.branch_protections {
-            if !patterns.insert(&protection.pattern) {
+            if !patterns.insert((protection.target, &protection.pattern)) {
                 bail!(
-                    r#"repo '{}' uses multiple branch protections with the pattern `{}`"#,
+                    r#"repo '{}' uses multiple {:?} protections with the pattern `{}`"#,
                     repo.name,
+                    protection.target,
```
marcoieni, line 1160: A repo can have the same pattern for branch and tag

rust_team_data/src/v1.rs
Kobzol: Wait, this is the v1 API, not GitHub API models, why do we need a default at all? Once triagebot has been updated, the field should always be present, right?

src/schema.rs
```diff
@@ -883,11 +894,13 @@ pub(crate) struct BranchProtection {
     pub allowed_merge_apps: Vec<AllowedMergeApp>,
     #[serde(default)]
     pub merge_queue: bool,
-    #[serde(default = "default_true")]
+    #[serde(default = "rust_team_data::v1::branch_protection_default_prevent_creation")]
```
Kobzol: The defaults should be chosen here, not in the v1 API.

src/validate.rs
Kobzol: Maybe we could create some `BranchProtectionId` to encode what is the unique identifier of the protection (target + pattern, maybe plus name?). But if it's only used here, it's fine.

src/schema.rs
marcoieni: the issue is that the `rust_team_data` crate doesn't import this crate. So it can't use the default.

marcoieni: oh, I saw your other comment.

src/sync/github/api/mod.rs
```diff
@@ -635,21 +644,48 @@ pub(crate) struct MergeQueueParameters {
     pub(crate) min_entries_to_merge_wait_minutes: i32,
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
+pub(crate) const DEFAULT_MERGE_QUEUE_TIMEOUT_MINUTES: i32 = 360;
+pub(crate) const DEFAULT_MERGE_QUEUE_MAX_ENTRIES_TO_BUILD: i32 = 5;
+pub(crate) const DEFAULT_MERGE_QUEUE_MAX_ENTRIES_TO_MERGE: i32 = 5;
+pub(crate) const DEFAULT_MERGE_QUEUE_MIN_ENTRIES_TO_MERGE: i32 = 0;
+pub(crate) const DEFAULT_MERGE_QUEUE_MIN_ENTRIES_TO_MERGE_WAIT_MINUTES: i32 = 0;
```
ubiratansoares, line 651: When I saw these defaults I remembered [this issue](https://github.com/rust-lang/team/issues/2196), should we use these constants also [here](https://github.com/rust-lang/team/blob/main/src/sync/github/mod.rs#L1053C1-L1066C6)? Ofc we don't need to discuss the defaults, just wandering whether we should centralise

rust_team_data/src/v1.rs
```diff
@@ -252,22 +252,34 @@ pub enum MergeBot {
     RustTimer,
     Bors,
     WorkflowsCratesIo,
+    PromoteRelease,
```
Kobzol, line 255: We don't have deny unknown fields on the branch protection, so adding the new fields there should be fine for downstream. But this enum variant will brick triagebot, because it won't know it. So we will again have to first land just the infra change, then update triagebot, and then land the cargo repo config change itself.

marcoieni: yes, this is what I wrote in the PR body, right?

src/sync/github/api/mod.rs
marcoieni: I'm already using these constants there indirectly. You linked code from the main branch, not from this branch.
image

ubiratansoares (approved):

rust_team_data/src/v1.rs
Kobzol: Indeed, sorry, I missed that.

Issue #1735, opened 2024-01-05
Support merge queue branch protections


ehuss:
This is an issue to track adding [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) branch protection support. This is **currently not possible** because it is not supported by GitHub's API. I figured it might be helpful to have an issue to track it nonetheless.

There are open requests for this at:
* https://github.com/orgs/community/discussions/50893
* https://github.com/orgs/community/discussions/77614

AFAIK, it *should* be OK for admins to manually set this even if the repo is managed in the team database. It looks like [`update_branch_protection`](https://github.com/rust-lang/sync-team/blob/9789ad2d5ac835c47cd2f84a828f16873b8ef760/src/github/api.rs#L587-L672) should be compatible. I have not tested this, yet, since all merge-queue repos I know of aren't tracked, yet.

amustaque97:
- Concepts & constraints
  - Branch protection rules (classic) now have a checkbox “Require merge queue”, but the public REST branches/branch-protection API does not expose a merge queue flag.​
  - GraphQL createBranchProtectionRule/updateBranchProtectionRule also do not support merge queue options yet (GitHub community discussion confirms this).​

- Rulesets do support Merge queue behaviour: merge method, min/max group size, build concurrency, status‑check behaviour, etc
Github API link: https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#create-a-repository-ruleset

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/41855a70-19e4-4d5a-9cd2-2bdcbce5731a" />


If we plan migration from `[[branch-protections]]` to `rulesets`, this should be doable. If we all agree, I can propose subcommand changes (`[[branch-protections]]` to `rulesets`). After migration, we can enable merge queues. 

I would love to hear your thoughts.

Kobzol:
The main required thing for that is to reimplement the branch config from protections to rulesets in `team`. Although we might want to do that in parts and support both protections and rulesets for some time, because changing everything from protections to rulesets at once seems scary (maybe we could do an allowlist of a few repos where we will use rulesets though).

amustaque97:
> reimplement the branch config from protections to rulesets in `team`

To begin with, I will create the PR that will replicate the branch protection to rulesets with the merge queues.

Do you have any repos in mind to start with?

Kobzol:
[...]

PR #2344, opened 2026-03-25
turn merge queue into a struct to support config option

marcoieni:
We want to allow people to configure merge queues (see https://github.com/rust-lang/team/issues/1735#issuecomment-4121756256).

### DIFF
--- a/src/static_api.rs
+++ b/src/static_api.rs
@@ -79,7 +79,7 @@ impl<'a> Generator<'a> {
                             AllowedMergeApp::PromoteRelease => v1::MergeBot::PromoteRelease,
                         })
                         .collect(),
-                    merge_queue: b.merge_queue,
+                    merge_queue: b.merge_queue.enabled,
                     prevent_creation: b.prevent_creation,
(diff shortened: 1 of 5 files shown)


src/static_api.rs
```diff
@@ -79,7 +79,7 @@ impl<'a> Generator<'a> {
                             AllowedMergeApp::PromoteRelease => v1::MergeBot::PromoteRelease,
                         })
                         .collect(),
-                    merge_queue: b.merge_queue,
+                    merge_queue: b.merge_queue.enabled,
```
marcoieni, line 82: this allows to keep the v1 API retro compatible (at least for now)

Kobzol (commented):
It would be slightly nicer if we could just use `merge-queue = true` to use the defaults, rather than always having to expand the config, but that's not super important.

marcoieni:
Having to support both `merge-queue = true` and `merge-queue = { enabled = true }` complicates the team codebase I think, so I would stick to this solution for now 👍

ubiratansoares (approved):

PR #2346, opened 2026-03-25
change prevent-update default

marcoieni:
Discussed in https://github.com/rust-lang/team/pull/2327#discussion_r2987717972

The dry run sets "Restrict updates: true" to all rulesets that have `pr-required: true`, which I think is good.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -447,8 +447,8 @@ prevent-deletion = true
 # (optional - default `true`)
 prevent-force-push = true
 # Whether to prevent updates to the branch.
-# (optional - default `false`)
-prevent-update = false
+# (optional - default matches `pr-required`)
+prevent-update = true
 # Whether the protection applies to branches or tags.
 # Options are "branch" or "tag".
 # (optional - default `branch`)
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -245,6 +245,12 @@ pub enum BranchProtectionMode {
     PrNotRequired,
 }
 
+impl BranchProtectionMode {
+    pub fn is_pr_required(&self) -> bool {
+        matches!(self, BranchProtectionMode::PrRequired { .. })
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub enum MergeBot {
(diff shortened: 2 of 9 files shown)


marcoieni:
> `rust_team_data/src/v1.rs` has been modified, it is used (as a git dependency) by multiple sub-projects like [triagebot](https://github.com/rust-lang/triagebot), the [www.rust-lang.org](https://github.com/rust-lang/www.rust-lang.org) website and others.
> 
> If you are changing the data structures, please make sure that the changes are not going to break serde deserialization (adding a field is fine; removing or renaming a field isn't).
> 
> If you must do a breaking change to the format, make sure to coordinate it with [all the users](https://github.com/search?q=org%3Arust-lang+%2Fname+%3D+%22rust_team_data%22%2F+NOT+repo%3Arust-lang%2Fteam+NOT+repo%3Arust-lang%2Fsync-team&type=code) of the `rust_team_data` crate.

The file was modified in a retro compatible way (I added a method to a struct) ✅

Kobzol:
Rulesets are complicated enough on their own, but having to think about supporting branch protections and rulesets at the same time is something else =D I assume that `prevent-update`  controls both whether you can push to a branch and whether you can approve a PR that targets that branch (but I haven't checked that).

Also for bypass actors, even if a PR is required, they don't actually have to create a PR, right? This is complicated.. :)

So:
- `PR required = false + prevent-updates = false` means that anyone can push to a branch. Probably not very common, but can still be useful, as you might still want to prevent deleting a branch.
- `PR required = true + prevent-updates = false` means that you have to open a PR, and anyone with write access can then merge it.
- `PR required = false + prevent-updates = true` means that a PR is not required, but only bypass actors can push to a branch. This is the use-case for bors.
- `PR required = true + prevent-updates = true` means that you have to open a PR, and only bypass actors can merge that PR. But since bypass actors can actually override `PR required = true`, this setting is pretty much the same as the previous setting, I think? It seems that crates.io uses this one, so that bypass actors can actually push to the branch even without opening a PR, I think?

Kobzol:
Based on what I wrote above, I think that if we want to support both `pr-required` and `prevent-update`, and we have `pr-required = true` as the default, then in the end `prevent-update` should default to `false`, otherwise it would mean that people with write access won't be able to merge PRs targeting that branch?

Edit: yeah, I tested it and this is how it works.

marcoieni:
ok, since the current default is the right thing to do, I'm closing this PR

PR #2353, opened 2026-03-28
add merge queue settings

marcoieni:
This PR makes merge queue configurable to allow to encode in the toml files the remaining repositories listed in https://github.com/rust-lang/team/issues/1735#issuecomment-4147589452

I made no data changes to the repositories and the v1 API doesn't return the new fields if the data is the default so that triagebot is retrocompatible. Because the V1 API JSON response shouldn't differ after we merge this PR.

- [ ] After this PR is merged, update triagebot

After we update triagebot, we can configure the remaining merge queues in the toml files.

GitHub docs at https://docs.github.com/en/rest/repos/rules?apiVersion=2026-03-10#create-a-repository-ruleset--parameters

### DIFF
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -275,6 +275,15 @@ pub enum ProtectionTarget {
     Tag,
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum MergeQueueMethod {
+    #[default]
+    Merge,
+    Squash,
+    Rebase,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct BranchProtection {
     pub pattern: String,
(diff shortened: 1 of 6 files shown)


rust_team_data/src/v1.rs
```diff
@@ -325,3 +356,39 @@ pub struct People {
...
+    *value == default_merge_queue_max_entries_to_merge()
+}
+
+fn default_merge_queue_check_response_timeout_minutes() -> u32 {
+    60
+}
+
+fn is_default_merge_queue_check_response_timeout_minutes(value: &u32) -> bool {
+    *value == default_merge_queue_check_response_timeout_minutes()
+}
```
marcoieni, line 394: all these defaults can be removed after we update triagebot I think.
Because we can always serialize everything.

src/sync/github/mod.rs
```diff
@@ -1049,7 +1053,7 @@ pub fn construct_ruleset(branch_protection: &rust_team_data::v1::BranchProtectio
                 dismiss_stale_reviews_on_push: branch_protection.dismiss_stale_review,
                 require_code_owner_review: REQUIRE_CODE_OWNER_REVIEW_DEFAULT,
                 require_last_push_approval: REQUIRE_LAST_PUSH_APPROVAL_DEFAULT,
-                required_approving_review_count: *required_approvals as i32,
+                required_approving_review_count: github_int(*required_approvals),
```
ubiratansoares, line 1056: non-blocking : do we have any particular reasons to have these fields as `i32` rather than `u32`? 

Asking since I'm not sure I'm missing something from Github API docs, but intuitively I'd not expect negative values from PR parameters or MQ parameters

ubiratansoares (approved):
Left a non-blocking question

src/sync/github/mod.rs
marcoieni: GitHub documents these fields as "integers". So I was afraid that they encode "-1" as special values for infinite for example.
It's very defensive and I'm not sure it's the right choice. but we can revisit in other PRs

PR #2362, opened 2026-03-31
support require up-to-date branches option

marcoieni:
This setting was set manually from the UI. With this PR we add in `team` the ability to set this field via toml.

For the docs, check the field `requiresStrictStatusChecks` in: https://docs.github.com/en/graphql/reference/objects#branchprotectionrule

- [x] Blocked by https://github.com/rust-lang/team/pull/2364 and https://github.com/rust-lang/triagebot/pull/2359

## Why is the dry run not empty?
This PR disables this field for repositories that don't have checks. As you can see from the settings page, this field does nothing in this case:
image

- [x] When reviewing this PR, I checked that all the repositories shown in the dry run don't have any required checks.

PR #2364, opened 2026-03-31
add `require_up_to_date_branches` to v1 api

marcoieni:
Before merging https://github.com/rust-lang/team/pull/2362/, we merge this PR and update triagebot.


rust_team_data/src/v1.rs
```diff
@@ -296,6 +296,8 @@ pub struct BranchProtection {
     pub allowed_merge_teams: Vec<String>,
     pub merge_bots: Vec<MergeBot>,
     pub allowed_merge_apps: Vec<MergeBot>,
+    #[serde(default, skip_serializing_if = "is_false")]
+    pub require_up_to_date_branches: bool,
```
marcoieni, line 300: we don't serialize if false, so that merging this change doesn't break triagebot.

ubiratansoares (approved):

rust_team_data/src/v1.rs
Kobzol: Shouldn't matter anyway, as new fields shouldn't break downstream consumers, they will be ignored.

Kobzol (approved):

PR #2373, opened 2026-04-02
google-workspace: add data models

ubiratansoares:
First PR towards #2363 

A few notes : 

- I preferred to frame naming around `Google Workspace` for now, but happy to discuss
- Added `#[allow(dead_code)]` in some places to make clippy happy, will remove them afterwards
- I'm pointing generically to [rust-forge/infra/docs](https://forge.rust-lang.org/infra/docs) while we don't have a dedicated doc for Google SSO and access management

Regarding data modelling, some considerations on `first-name` and `last-name` requirements.

(1) unfortunately, although the related [UserName](https://developers.google.com/workspace/admin/directory/reference/rest/v1/users#UserName) data object predicts `fullName`, I'm pretty much sure this is a computed field. There is no such option when creating a new user either through the GWS admin console or the [corresponding Terraform Provider](https://registry.terraform.io/providers/hashicorp/googleworkspace/latest/docs/resources/user#required).

(2) we could try to split person#name to fullfill what the GWS Admin API expects, but for some users it won't work ([example](https://github.com/rust-lang/team/blob/main/people/walterhpearce.toml))

Regarding the validation, requesting a personal/work email : that will be used as recovery email address, which is not mandatory, but useful for other purposes, including sending the initial password to the new user.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -13,6 +13,7 @@ zulip-id = 123456   # Zulip ID of the person (required)
 discord-id = 123456 # Discord ID of the person (optional)
 # You can also set `email = false` to explicitly disable the email for the user.
 # This will, for example, avoid adding the person to the mailing lists.
+# Note: if you need Google SSO access, this field is mandatory
 email = "john@doe.com"  # Email address used for mailing lists (optional)
 irc = "jdoe"  # Nickname of the person on IRC, if different than the GitHub one (optional)
 matrix = "@john:doe.com" # Matrix username (MXID) of the person (optional)
@@ -24,6 +25,13 @@ github-sponsors = true
 
 [permissions]
 # Optional, see the permissions documentation
+
+[google-workspace]
+# Optional, available only for users who require managed infrastructure access
+# See https://forge.rust-lang.org/infra/docs
+first-name = "John"  # Your first name
+last-name = "Doe"   # Your last name or preferred surname
+account-handle = "john.doe" # The handle to define a @rust-lang.org account (e.g. john.doe@rust-lang.org)
 ```
 
 The file must be named the same as the GitHub username.
@@ -45,6 +53,11 @@ top-level = true
 # - project-group
 # - marker-team
 kind = "working-group"
+
+# Optional, defaults to false.
+# Set this field to true to grant team members access to specific infrastructure systems via Google Groups.
+# See https://forge.rust-lang.org/infra/docs
+google-workpace-saml-group = false
 
 [people]
 # Leads of the team, can be more than one and must be members of the team.
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -39,6 +39,7 @@ static CHECKS: &[Check<fn(&Data, &mut Vec<String>)>] = checks![
     validate_subteam_of,
     validate_team_leads,
     validate_team_members,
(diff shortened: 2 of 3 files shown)


src/validate.rs
```diff
@@ -545,6 +546,22 @@ fn validate_people_addresses(data: &Data, errors: &mut Vec<String>) {
     });
 }
 
+/// Ensure personal email provided for people with access to Google Workspace
+fn validate_people_sso(data: &Data, errors: &mut Vec<String>) {
```
Kobzol: Another validation might be that members of teams with `google-workpace-saml-group = false` should have the `[google-workspace]` section set? Or would the SAML group be applied only to those team members who have it enabled, so that we don't have to make it required?

ubiratansoares: > Another validation might be that members of teams with google-workpace-saml-group = false should have the [google-workspace] section set? 

Very good question, I did not consider that. I think we could validate that setting 

```
google-workpace-saml-group = true
```

requires all team members to have a GWS object in-place, because the goal of this field is provisioning the corresponding Google Group (when needed) and include/update team members on it. 

Ie, setting it on `teams/infra.toml` implies a Google Group conventionally named `infra-saml` with all infra team `members` on it. Moving someone from one team to another automatically changes which SAML apps that person has access to, for example.

On the other hand, perhaps we may have use cases that require a standalone Workspace not necessarily associated with any Google Group. Not sure tbh, but not a problem for now :P 

Did I clarify things a bit? Please let me know :)

Kobzol: Yeah I think it makes sense to have the groups associated with teams for now, it's certainly easier. I'd then add the validation that in teams with the SAML group, all members have the worspace set up.

ubiratansoares: Added a validation!

docs/toml-schema.md
```diff
@@ -13,6 +13,7 @@ zulip-id = 123456   # Zulip ID of the person (required)
 discord-id = 123456 # Discord ID of the person (optional)
 # You can also set `email = false` to explicitly disable the email for the user.
 # This will, for example, avoid adding the person to the mailing lists.
+# Note : if you need Google SSO access, this field becomes mandatory
```
```suggestion
# Note: if you need Google SSO access, this field is mandatory
```

docs/toml-schema.md
```diff
@@ -46,6 +54,11 @@ top-level = true
 # - marker-team
 kind = "working-group"
 
+# Optional, defaults to false
+# Set this field to true to grant team members access to specific infrastructure systems via Google Groups
```
```suggestion
# Optional, defaults to false.
# Set this field to true to grant team members access to specific infrastructure systems via Google Groups.
```

src/validate.rs
```diff
@@ -248,6 +250,27 @@ fn validate_team_members(data: &Data, errors: &mut Vec<String>) {
     });
 }
 
+/// Ensure that teams mapped to Google Groups have all GWS accounts already provisioned
```
marcoieni:
```suggestion
/// Ensure that teams mapped to Google Groups have all Google Workspace (GWS) accounts already provisioned
```
Let's expand GWS in the comment so that people knows how to read the function name

marcoieni (dismissed):
looks good. Very minor comments.

marcoieni (approved):

PR #2385, opened 2026-04-07
rulesets: log when default fields change

marcoieni:
Fix https://github.com/rust-lang/team/issues/2376

I also convert `crates.io-index-archive` to prove that this PR works. In fact, the dry run now shows `prevent-creation = false`


ubiratansoares (approved):
This additional logging definitely helps 🙂

PR #2397, opened 2026-04-09
use rulesets by default

marcoieni:
Switch from an allow approach to a deny approach.


src/sync/github/mod.rs
```diff
@@ -444,7 +444,7 @@ impl SyncGitHub {
     /// Check if a repository should use rulesets instead of branch protections
     fn should_use_rulesets(&self, repo: &rust_team_data::v1::Repo) -> bool {
         let repo_full_name = format!("{}/{}", repo.org, repo.name);
-        self.config.enable_rulesets_repos.contains(&repo_full_name)
+        !self.config.disable_rulesets_repos.contains(&repo_full_name)
```
marcoieni, line 447: here is where the logic changes

src/sync/github/mod.rs
```diff
@@ -1569,7 +1569,7 @@ struct UpdateRepoDiff {
     // old, new
     settings_diff: (RepoSettings, RepoSettings),
     permission_diffs: Vec<RepoPermissionAssignmentDiff>,
-    branch_protection_diffs: Vec<BranchProtectionDiff>,
+    branch_protection_diffs: BTreeSet<BranchProtectionDiff>,
```
marcoieni: I had to change to a BTreeSet because the tests was flaky. Instead, a fixed order fixed it.

This change requires the additional derives you see in the PR

Kobzol: This is fishy, `Vec` is ordered too. Using sorting based on some custom comparison, rather than on insertion order, seems like it hides a different bug. Maybe some of the input collections iterated in `diff_branch_protections` do not have a stable order?

src/sync/github/api/read.rs
```diff
@@ -74,7 +74,7 @@ pub(crate) trait GithubRead {
         &self,
         org: &str,
         repo: &str,
-    ) -> anyhow::Result<HashMap<String, (String, BranchProtection)>>;
+    ) -> anyhow::Result<BTreeMap<String, (String, BranchProtection)>>;
```
marcoieni, line 77: changed this so that we order the branch protections when we read them from GitHub

src/sync/github/mod.rs
marcoieni: I think I addressed this comment, let me know

EDIT: I saw you upvoted my other change 👍

Kobzol: Yes, thank you!

ubiratansoares (approved):
🚀🚀🚀

PR #2402, opened 2026-04-10
CI: Set minimal `permissions:` blocks in workflows

Turbo87:
This addresses `excessive-permissions` and `undocumented-permissions` findings reported by `zizmor`. Each workflow now defaults to `permissions: {}` at the top level and grants only the scopes it actually needs at the job level, with inline comments documenting the reason for every non-trivial grant.

See:
- https://docs.zizmor.sh/audits/#excessive-permissions
- https://docs.zizmor.sh/audits/#undocumented-permissions

### DIFF
--- a/.github/workflows/dry-run.yml
+++ b/.github/workflows/dry-run.yml
@@ -18,6 +18,8 @@ on:
     types:
       - completed
 
+permissions: {}
+
 jobs:
   dry-run:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


marcoieni (approved):

.github/workflows/dry-run.yml
```diff
@@ -27,7 +29,7 @@ jobs:
       group: dry-run-${{ github.event.workflow_run.head_branch }}
       cancel-in-progress: true
     permissions:
-      pull-requests: write
+      pull-requests: write  # Needed to post and edit the dry-run result comment on the PR
```
Kobzol, line 32: I wonder if this won't also need `contents: read` to clone the contents of the PR/main branch. I guess we'll see soon! 😆

Turbo87: hmm, good question 😅 

was kinda hard to test locally. let me know if you need me to adjust anything :)

PR #2335, opened 2026-03-19
move cargo to rulesets

marcoieni:

PR #2406, opened 2026-04-12
allow creation of rust branches in cargo

marcoieni:
Fix https://github.com/rust-lang/team/pull/2335#issuecomment-4232572054

### DIFF
--- a/repos/rust-lang/cargo.toml
+++ b/repos/rust-lang/cargo.toml
@@ -15,6 +15,7 @@ merge-queue = { enabled = true }
 [[branch-protections]]
 pattern = "rust-1.*"
 ci-checks = ["conclusion"]
+prevent-creation = false
 
 [[branch-protections]]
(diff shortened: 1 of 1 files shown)


ubiratansoares (approved):

repos/rust-lang/cargo.toml
```diff
@@ -15,6 +15,7 @@ merge-queue = { enabled = true }
 [[branch-protections]]
 pattern = "rust-1.*"
 ci-checks = ["conclusion"]
+prevent-creation = false
```
Mark-Simulacrum, line 18: Hm, I wonder if we should grant bypass access only to promote-release. It's probably fine as-is, we'd notice if someone else created the branch when promote-branches fails.

marcoieni: Feel free to open a PR in case you want to make this change 👍

PR #2408, opened 2026-04-13
Add concurrency block to prevent overlapping scheduled runs

Sandijigs:
This  PR  fixes the 1 concurrency-limits finding zizmor flagged under `--persona=pedantic`. I added  a workflow-level `concurrency:` block to the `check-untracked-repos` workflow, preventing multiple runs from overlapping when a scheduled cron and a manual `workflow_dispatch` are triggered around the same time.

Reference: https://woodruffw.github.io/zizmor/audits/#concurrency-limits

### DIFF
--- a/.github/workflows/check-untracked-repos.yml
+++ b/.github/workflows/check-untracked-repos.yml
@@ -6,6 +6,10 @@ on:
     - cron: '0 */6 * * *'
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
 permissions: {}
 
 jobs:


Kobzol:
That sounds like overkill, tbh, that situation likely ~never happens and even if it did, it wouldn't really be an issue :D

Sandijigs:
> That sounds like overkill, tbh, that situation likely ~never happens and even if it did, it wouldn't really be an issue :D

Sandijigs:
The change was actually  driven by zizmor flagging it under `--persona=pedantic` . But I can  close this if it's not worth the noise. or would you prefer I keep going through the remaining pedantic findings and bundle it together into a single PR instead ?

marcoieni:
Thanks for raising this. For cases like this where we don't think a zizmor lint is necessary to address, I wonder if there's a way to allow it. Similar to how you would do `#[allow]` in clippy. I'm not sure.

Anyway, feel free to go through the remaining pedantic findings. Either with one PR or many, as you prefer 👍

PR #2409, opened 2026-04-13
Fixed zizmor pedantic findings across workflow files

Sandijigs:
zizmor flagged 10  findings in the workflow files under` --persona=pedantic`.I was able to fix 6 out of the 10 findings .

- Added `persist-credentials: false` to the test job checkout in main.yml (https://docs.zizmor.sh/audits/#artipacked)
-  Moved `${{ }} `expressions in `run: `blocks into` env: `vars in main.yml (3 findings) (https://docs.zizmor.sh/audits/#template-injection)
- Replaced` $GITHUB_ENV` with `$GITHUB_OUTPUT` for the PR number step in dry-run.yml (https://docs.zizmor.sh/audits/#github-env)

4 findings remain unfixed:

- `secrets-outside-env `(3):  I think fixing this requires a GitHub environment to be set up in the repo settings first, so I didnt fix it (https://docs.zizmor.sh/audits/#secrets-outside-env)

- `dangerous-triggers` (1): The `workflow_run` trigger in `dry-run.yml` this runs in the default branch context so the workflow can safely post PR comments without checking out untrusted code [(https://docs.zizmor.sh/audits/#secrets-outside-env)](https://docs.zizmor.sh/audits/#dangerous-triggers)

### DIFF
--- a/.github/workflows/dry-run.yml
+++ b/.github/workflows/dry-run.yml
@@ -101,12 +101,13 @@ jobs:
           cat comment.txt
 
       - name: Extract PR number
+        id: extract-pr
         run: |
           # We read the PR number that is stored in the uploaded archive
           # and check that it is an integer (as the workflow could upload whatever it wants).
           UNSANITIZED_PR=`cat team-api/pr.txt`
           if [[ ${UNSANITIZED_PR} =~ ^[0-9]+$ ]]; then
-            echo "PR_NUMBER=${UNSANITIZED_PR}" >> $GITHUB_ENV
+            echo "pr_number=${UNSANITIZED_PR}" >> $GITHUB_OUTPUT
           else
             echo "Invalid PR number passed: ${UNSANITIZED_PR}"
             exit 1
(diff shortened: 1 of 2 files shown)


.github/workflows/check-untracked-repos.yml
```diff
@@ -6,6 +6,10 @@ on:
     - cron: '0 */6 * * *'
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
```
marcoieni: The other PR was closed because a maintainer said they didn't like this change. Why including it in this PR?

Sandijigs: I have removed the part from the PR .

marcoieni:
This looks good, thanks, can you squash your commits into one? We don't want to include the merge commit in the git history

marcoieni (approved):

.github/workflows/check-untracked-repos.yml
ubiratansoares: @Sandijigs  an alternative here could be ignoring this particular finding, adding a comment about why it is not a big deal and therefore, safe to ignore. That could be a follow-up PR if you prefer :)

Sandijigs: Oh Nice , I didnt  think about it that way . can I Also  do that for the other 4 findings that  were not solved . @ubiratansoares

PR #2415, opened 2026-04-14
Make library team zulip channels managed by teams repo

joshtriplett:
This makes sure we don't miss adding people to these channels.


jieyouxu (approved):

jieyouxu:
Hold on, I think the failure is genuine:

> failed: Id of stream 't-libs/private' not found. The stream probably doesn't exist and sync-team doesn't support creating it yet. Please create the stream manually and add the rust-lang-owner user to it.

AFAICT this private stream does not yet exist. @joshtriplett or @Mark-Simulacrum since y'all are on `T-{libs,libs-contributors}`, can you manually create the private stream + invite `rust-lang-owner` zulip account?

PR #2417, opened 2026-04-15
add mejrs, mu001999 and ShoyuVanilla to t-compiler

davidtwco:
🎉 Congrats to @mejrs, @mu001999 and @ShoyuVanilla for joining the compiler team!

r? @wesleywiser


wesleywiser (approved):
Congrats and well deserved! 🎉

people/mejrs.toml
```diff
@@ -0,0 +1,5 @@
+name = "Bruno Kolenbrander"
+email = "59372212+mejrs@users.noreply.github.com"
```
jackh726, line 2: Hmm. So, the idea is that these emails are for communication. This email isn't *real* in that it can't receive emails.

This hasn't come up previously, but I'm not sure we should accept this. I'll note that if there is a worry about sharing an email publicly, we *do* have a way to encrypt an added email, and should talk to the infra folks about that.

jackh726 (changes requested):

people/mejrs.toml
Kobzol: We don't currently have any requirement for Project members to share any kind of "usable" e-mail, we only have a requirement for a GitHub and a Zulip account. I'm sure there are more people in the `team` DB that use e-mails here that are not read or can't even receive e-mails. We know historically that our e-mail lists are not working well (though hopefully that can change in the future), which is why we created the Zulip account requirement.

If you'd like to propose a requirement that "valid" e-mail addresses are used, then that can be proposed to the Council, but it is not existing practice and this PR should not be blocked on that.

mejrs: So far I've not plastered my email anywhere public (though one can probably dox it if they want). It's just a preference of mine. If someone seconds your concern then @davidtwco feel free to put my actual email there.

jackh726 (approved):

PR #2337, opened 2026-03-19
Create GPU notification group

apiraino:
As per discussion on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/233931-t-compiler.2Fmajor-changes/topic/Create.20a.20GPU.20notification.20group.20compiler-team.23960/near/568629680).

@Kobzol before sending pings to everyone mentioned in this new file, can you let me know if this looks correct (never created a notification group before).

Thanks :)

PR #2419, opened 2026-04-15
remove myself from gpu-target group

RalfJung:
I didn't consent to be added and don't have capacity for this, unfortunately.

### DIFF
--- a/teams/gpu-target.toml
+++ b/teams/gpu-target.toml
@@ -23,6 +23,5 @@ members = [
     "Sa4dUs",
     "ZuseZ4",
     # other members for T-compiler
-    "RalfJung",
     "workingjubilee"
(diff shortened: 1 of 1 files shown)


marcoieni:
you don't want to be listed in the alumni section? (if yes, just add the alumni array as it is not present)

RalfJung:
No, I'd rather not. I wasn't ever actually a member. (Also this is more a ping group than a team so tracking alumni seems less relevant.)

marcoieni (approved):

apiraino:
To be clear, this was a misunderstanding from me (see [here](https://github.com/rust-lang/team/pull/2337/files#r3087962126)), I'm sorry about that!

PR #2420, opened 2026-04-15
Added CLI command to automate team and repo archival

### LINKED ISSUE #1547: Create a command for archiving teams and repos
Archiving a team/repo happens relatively often, and currently it's a bit of a manual process with unclear rules that are mostly only found on CI (move the TOML file to a correct repository, remove all members of a team, remove all permissions of a repo). It should be simple to create a command that would automate this.

Sandijigs:
This PR  adds a new archive subcommand to the CLI that handles moving TOML files to the archive/ directory, clearing team access on repos, and moving team members to alumni .
Example usage:

- `cargo run -- archive repo rust-lang/homu`

- `cargo run -- archive team project-generic-associated-types`

For repos, it moves the file to repos/archive/{org}/ and clears all entries under` [access.teams]`.
For teams, it moves the file to teams/archive/, sets leads and members to empty, and moves everyone into alumni.

I used toml_edit to modify the TOML files so that only the parts that need to change are touched  everything else (comments, formatting, ordering) stays the same.

Closes #1547

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ tokio = { version = "1", default-features = false, features = ["net", "rt-multi-
 tempfile = "3.19.1"
 thiserror = "2.0.18"
 toml = "1.0"
+toml_edit = "0.22"
 
 [dev-dependencies]
 ansi_term = "0.12.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ tokio = { version = "1", default-features = false, features = ["net", "rt-multi-
 tempfile = "3.19.1"
 thiserror = "2.0.18"
 toml = "1.0"
+toml_edit = "0.25"
 
 [dev-dependencies]
 ansi_term = "0.12.1"
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -1,0 +1,194 @@
+use anyhow::{Context, bail, format_err};
+use indexmap::IndexSet;
+use log::info;
+use std::path::Path;
+
+fn get_access_teams(doc: &mut toml_edit::DocumentMut) -> Option<&mut toml_edit::Table> {
+    doc.get_mut("access")?.get_mut("teams")?.as_table_mut()
+}
+
+fn archive_toml_file<F>(
+    src: &Path,
+    dest_dir: &Path,
+    dest: &Path,
+    entity: &str,
+    transform: F,
+) -> anyhow::Result<()>
+where
+    F: FnOnce(&mut toml_edit::DocumentMut),
+{
+    if !src.is_file() {
+        bail!("{entity} file not found: {}", src.display());
+    }
+    if dest.is_file() {
+        bail!("{entity} is already archived: {}", dest.display());
+    }
+
+    let mut doc = read_toml_mut(src)?;
+
+    transform(&mut doc);
+
+    std::fs::create_dir_all(dest_dir)
+        .with_context(|| format!("failed to create directory {dest_dir:?}"))?;
+    std::fs::write(dest, doc.to_string()).with_context(|| format!("failed to write {dest:?}"))?;
+    std::fs::remove_file(src).with_context(|| format!("failed to remove {src:?}"))?;
+
+    info!("archived {entity} {src:?} -> {dest:?}");
+    Ok(())
+}
+
+fn read_toml_mut(src: &Path) -> anyhow::Result<toml_edit::DocumentMut> {
+    let content =
+        std::fs::read_to_string(src).with_context(|| format!("failed to read {src:?}"))?;
+    let doc: toml_edit::DocumentMut = content
+        .parse()
+        .with_context(|| format!("failed to parse {src:?}"))?;
+    Ok(doc)
+}
+
+/// Archive a repository by moving its TOML file to `repos/archive/<org>/`
+/// and clearing every entry from the `[access.teams]` table.
+pub fn archive_repo(data_dir: &Path, name: &str) -> anyhow::Result<()> {
+    let (org, repo_name) = name
+        .split_once('/')
+        .ok_or_else(|| format_err!("repository must be in 'org/name' format, got '{name}'"))?;
+
+    let repos_dir = data_dir.join("repos");
+    let src = repos_dir.join(org).join(format!("{repo_name}.toml"));
+    let dest_dir = repos_dir.join("archive").join(org);
+    let dest = dest_dir.join(format!("{repo_name}.toml"));
+
+    archive_toml_file(&src, &dest_dir, &dest, "repo", |doc| {
+        if let Some(table) = get_access_teams(doc) {
+            table.clear();
+        }
+    })
+}
+
+/// Gather every username from a team's `leads`, `members`, and `alumni`
+/// arrays into a single deduplicated, order-preserving set.
+///
+/// Handles both bare strings (`"alice"`) and inline tables (`{ github = "alice" }`),
+/// skipping any entries that don't match either shape or that have an empty
+/// `github` field.
+fn collect_all_team_members(people_table: &toml_edit::Table) -> IndexSet<String> {
+    let mut all = IndexSet::new();
+    for key in ["leads", "members", "alumni"] {
+        let Some(arr) = people_table.get(key).and_then(|v| v.as_array()) else {
+            continue;
+        };
+        for item in arr.iter() {
+            let username = if let Some(s) = item.as_str() {
+                s.to_string()
+            } else if let Some(tbl) = item.as_inline_table() {
+                match tbl.get("github").and_then(|v| v.as_str()) {
+                    Some(s) => s.to_string(),
+                    None => continue,
+                }
+            } else {
+                continue;
+            };
+            if !username.is_empty() {
+                all.insert(username);
+            }
+        }
+    }
+    all
+}
+
+/// Build a TOML array of usernames laid out one per line with `    ` indentation
+/// and a trailing comma — matching the style used elsewhere in the team repo.
+fn build_alumni_array(usernames: &IndexSet<String>) -> toml_edit::Array {
+    let mut arr = toml_edit::Array::new();
+    for person in usernames {
+        let mut val = toml_edit::Value::from(person.as_str());
+        val.decor_mut().set_prefix("\n    ");
+        arr.push_formatted(val);
+    }
+    arr.set_trailing("\n");
+    arr.set_trailing_comma(true);
+    arr
+}
+
+/// Move everyone listed in a team's `leads`, `members`, and existing `alumni`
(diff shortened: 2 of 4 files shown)


marcoieni:
I tested archiving a team, but alumni are in one line. Can we put one member per line?
image

Sandijigs:
The  alumni are now listed one per line.

Screenshot 2026-04-15 at 10 37 38 pm

marcoieni:
One thing I noticed: when archiving a team, you should remove the team access from other repos, like it was done in https://github.com/rust-lang/team/pull/2214
Probably this is more complicated, so no need to do this if you don't have time.

EDIT: actually you already have the logic to remove teams from a repo so maybe it's not that difficult!

Sandijigs:
I added a `remove_team_from_repos function` that runs after the team file is archived. It scans all repo TOML files and removes the archived team's entry from any [access.teams] block where it appears.

Cargo.toml
```diff
@@ -28,6 +28,7 @@ tokio = { version = "1", default-features = false, features = ["net", "rt-multi-
 tempfile = "3.19.1"
 thiserror = "2.0.18"
 toml = "1.0"
+toml_edit = "0.22"
```
marcoieni: Why not using the latest version of this package? I.e. what `cargo add` adds.

src/main.rs
```diff
@@ -647,3 +672,165 @@ async fn perform_sync(opts: SyncOpts, data: Data) -> anyhow::Result<()> {
...
+    std::fs::write(&dest, doc.to_string())
+        .with_context(|| format!("failed to write {}", dest.display()))?;
+    std::fs::remove_file(&src).with_context(|| format!("failed to remove {}", src.display()))?;
+
+    info!("archived team {} -> {}", src.display(), dest.display());
+
+    remove_team_from_repos(data_dir, name)?;
+
+    Ok(())
+}
```
marcoieni: this function is very long. I would extract it into smaller ones.

src/main.rs
```diff
@@ -647,3 +672,165 @@ async fn perform_sync(opts: SyncOpts, data: Data) -> anyhow::Result<()> {
...
+        && let Some(table) = teams.as_table_mut()
+    {
+        table.clear();
+    }
+
+    std::fs::create_dir_all(&dest_dir)
+        .with_context(|| format!("failed to create directory {}", dest_dir.display()))?;
+    std::fs::write(&dest, doc.to_string())
+        .with_context(|| format!("failed to write {}", dest.display()))?;
+    std::fs::remove_file(&src).with_context(|| format!("failed to remove {}", src.display()))?;
```
marcoieni: this code is duplicated among the two functions. So it would be nice to extract it into a function and reuse it.

In general, please inspect your code and find opportunities similar to this one to extract common code.

src/main.rs
```diff
@@ -647,3 +672,165 @@ async fn perform_sync(opts: SyncOpts, data: Data) -> anyhow::Result<()> {
...
+    let mut doc: toml_edit::DocumentMut = content
+        .parse()
+        .with_context(|| format!("failed to parse {}", src.display()))?;
+
+    if let Some(access) = doc.get_mut("access")
+        && let Some(teams) = access.get_mut("teams")
+        && let Some(table) = teams.as_table_mut()
+    {
+        table.clear();
+    }
```
marcoieni: this code is also duplicated with the other function. It would be nice to extract it

Sandijigs:
Thanks for pointing that out. I  have extracted the [access.teams] navigation into a shared get_access_teams helper and updated both archive_repo and remove_team_from_repos to use it.

src/main.rs
```diff
@@ -647,3 +672,163 @@ async fn perform_sync(opts: SyncOpts, data: Data) -> anyhow::Result<()> {
...
+        .with_context(|| format!("failed to read {}", src.display()))?;
+    let mut doc: toml_edit::DocumentMut = content
+        .parse()
+        .with_context(|| format!("failed to parse {}", src.display()))?;
+
+    if let Some(people) = doc.get_mut("people")
+        && let Some(people_table) = people.as_table_mut()
+    {
+        let mut all_alumni: Vec<String> = Vec::new();
+        let mut seen = HashSet::new();
```
ubiratansoares: Could we use only the hashset here?

src/main.rs
```diff
@@ -647,3 +672,163 @@ async fn perform_sync(opts: SyncOpts, data: Data) -> anyhow::Result<()> {
...
+                    };
+                    if !username.is_empty() && seen.insert(username.clone()) {
+                        all_alumni.push(username);
+                    }
+                }
+            }
+        }
+
+        people_table.insert("leads", toml_edit::value(toml_edit::Array::new()));
+        people_table.insert("members", toml_edit::value(toml_edit::Array::new()));
```
ubiratansoares: It seems that

```rust
 toml_edit::Array::new().into()
```
works equality as

```rust
toml_edit::value(toml_edit::Array::new())
```

but it reads a bit shorter (also better, imho)

src/main.rs
```diff
@@ -647,3 +672,163 @@ async fn perform_sync(opts: SyncOpts, data: Data) -> anyhow::Result<()> {
...
+            .with_context(|| format!("failed to read {}", org_path.display()))?
+        {
+            let repo_path = repo_entry?.path();
+            if !repo_path.is_file() || repo_path.extension() != Some(std::ffi::OsStr::new("toml")) {
+                continue;
+            }
+
+            let content = std::fs::read_to_string(&repo_path)
+                .with_context(|| format!("failed to read {}", repo_path.display()))?;
+            let mut doc: toml_edit::DocumentMut = content
```
ubiratansoares: I _think_ the turbofish notation speaks better to what we have in other parts of the code. It worth's a look in case you are not familiar with :)

ubiratansoares (commented):
@Sandijigs Added some comments

src/main.rs
marcoieni: Ubi, personally I prefer the `: Type` notation as it's simpler. 🤔

Ubi, can you show an example of other parts of the codebase where we use the turbofish notation?

ubiratansoares: I don't have super strong opinion tbh, both read Ok to me :) 

Regarding examples, I found some of them ([here](https://github.com/rust-lang/team/blob/main/src/main.rs#L507) and [here](https://github.com/rust-lang/team/blob/main/src/sync/zulip/api.rs#L56)), in which typing an _existing var_ could be used instead of turbofish. 

Anyhow, no reasons to waste too much time here :P

src/main.rs
Sandijigs: Thanks for pointing this out , Should I switch IndexSet from indexmap? since  it handles both deduplication and preserves insertion order in a single data structure. which was what I was going for when i used hashset

Sandijigs:
I've switched to IndexSet instead of the  HashSet combo , It is definately  more  cleaner. Also swapped toml_edit::value() for .into() like @ubiratansoares suggested

src/main.rs
ubiratansoares: @Sandijigs I agree with Marco. We can definitely benefit from smaller functions here.

ubiratansoares:
@Sandijigs Thanks for addressing some of our comments. Feel free to catch up with @marcoieni suggestions, so we can get this PR sorted :)

Sandijigs:
@marcoieni @ubiratansoares I  have pushed the changes based on your reviews.

For the duplicated code, I pulled the shared file-handling bits (existence checks, read/parse, write/remove, log) out of `archive_repo` and `archive_team` and put them in a new function `archive_toml_file`. 

 For `archive_team` being too long, I split the people-table work into three smaller functions: `collect_all_team_members`, `build_alumni_array`, and `move_team_members_to_alumni`.

marcoieni:
can you move the code in a new file (eg `archive.rs`) so that `main.rs` doesn't grow too much?
After that please squash the commits into one to cleanup the commit history 🙏

src/archive.rs
```diff
@@ -0,0 +1,194 @@
...
+    doc.get_mut("access")?.get_mut("teams")?.as_table_mut()
+}
+
+fn archive_toml_file<F>(
+    src: &Path,
+    dest_dir: &Path,
+    dest: &Path,
+    entity: &str,
+    transform: F,
+) -> anyhow::Result<()>
```
marcoieni, line 16: I don't like importing directly anyhow::Error. I like `Error` to always be the one from std.
But this is a matter of taste.

src/archive.rs
```diff
@@ -0,0 +1,194 @@
...
+    F: FnOnce(&mut toml_edit::DocumentMut),
+{
+    if !src.is_file() {
+        bail!("{entity} file not found: {}", src.display());
+    }
+    if dest.is_file() {
+        bail!("{entity} is already archived: {}", dest.display());
+    }
+
+    let mut doc = read_toml_mut(src)?;
```
marcoieni, line 27: I created this function to extract a common behavior

src/archive.rs
```diff
@@ -0,0 +1,194 @@
...
+    let dest = dest_dir.join(format!("{name}.toml"));
+
+    archive_toml_file(&src, &dest_dir, &dest, "team", move_team_members_to_alumni)?;
+    remove_team_from_repos(data_dir, name)?;
+    Ok(())
+}
+
+fn remove_team_from_repos(data_dir: &Path, team_name: &str) -> anyhow::Result<()> {
+    let repos_dir = data_dir.join("repos");
+    assert!(repos_dir.is_dir(), "`repos` directory does not exist");
```
marcoieni, line 145: here you were failing silently (`return Ok`).
Instead assert is better:
* while reading the code you understand that it is not part of the logic so your brain can skip it if you just want to understand the code
* if `repos` is not a directory at this point it's better to fail because something is going wrong

src/archive.rs
```diff
@@ -0,0 +1,194 @@
...
+        );
+        if org_path.file_name() == Some(std::ffi::OsStr::new("archive")) {
+            continue;
+        }
+
+        for repo_entry in
+            std::fs::read_dir(&org_path).with_context(|| format!("failed to read {org_path:?}"))?
+        {
+            let repo_path = repo_entry?.path();
+            remove_team_from_repository(team_name, &repo_path)?;
```
marcoieni, line 163: I extracted this function so that remove_team_from_repos is shorter

marcoieni (commented):
looks good! I pushed some minor changes.
Here are some comments to explain why I made these changes. (These comments do not require actions from you, they are just to explain what I did)
If they look good to you and you tested this command, please squash the commits and then we can merge this PR!

Sandijigs:
All Done @marcoieni  .  Squashed it down to one commit . Ran the command on a couple of teams/repos and the error cases, all good. Let me know if anything else is needed.

marcoieni (approved):
Thanks!

PR #2421, opened 2026-04-16
adopt renovate org preset

marcoieni:


.github/renovate.json5
```diff
@@ -1,10 +1,15 @@
...
+  extends: ["github>rust-lang/renovate"],
+  lockFileMaintenance: {
+    extends: ["schedule:monthly"]
+  },
+  packageRules: [
+    {
+      matchManagers: ["github-actions"],
+      extends: ["schedule:monthly"]
+    }
+  ]
```
ubiratansoares:
```suggestion
  ],
```

This makes the config valid 😄

ubiratansoares (commented):
I found an issue in the config!

.github/renovate.json5
marcoieni: thanks! I wonder if we should embrace json5 and add trailing commas everywhere in our docs.

ubiratansoares (approved):
Agree with json5 recommendation, looks like something to bring to forge docs. :)

PR #2423, opened 2026-04-17
Give a couple teams triaging permissions for rust-analyzer

Veykril:
This allows bjorn3 to keep triaging without being put into the contributors team solely for the permissions, and also allows some project folks to just fill out labels as they see fit when filing issues. (I wonder if we should just put down the `all` team here or whether that would be too much 🤔)


jieyouxu (approved):

jieyouxu:
Starting with a few limited teams seem fine, all is likely unncesaary, though its only triage

PR #2428, opened 2026-04-20
allow bots to bypass some rust rulesets

marcoieni:
Discussed in [#t-infra > Moving rust repo to rulesets @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Moving.20rust.20repo.20to.20rulesets/near/587783121)

I think these rules could prevent bots from working correctly.
Out of precaution, let's not apply these rules to bots.

### DIFF
--- a/repos/rust-lang/rust.toml
+++ b/repos/rust-lang/rust.toml
@@ -92,12 +92,12 @@ prevent-force-push = false
 
 [[branch-protections]]
 pattern = "*"
-allowed-merge-apps = ["promote-release"]
+allowed-merge-apps = ["bors", "promote-release", "rust-timer"]
 prevent-update = true
 
 [[branch-protections]]
 pattern = "*/**/*"
-pr-required = false
+allowed-merge-apps = ["bors", "promote-release", "rust-timer"]
 prevent-update = true
 
 [[branch-protections]]


ubiratansoares (dismissed):

repos/rust-lang/rust.toml
```diff
@@ -92,12 +92,12 @@ prevent-force-push = false
 
 [[branch-protections]]
 pattern = "*"
-allowed-merge-apps = ["promote-release"]
+allowed-merge-apps = ["bors", "promote-release", "rust-timer"]
 prevent-update = true
 
 [[branch-protections]]
 pattern = "*/**/*"
-pr-required = false
```
marcoieni, line 100: I had to remove `pr-required = false` because of the following CI error:

```
validation error: repo 'rust' uses bors, but its branch protection for */**/* uses invalid
    attributes (`required-approvals`, `dismiss-stale-review`, `pr-required` or `allowed-merge-teams`).
    Please remove the attributes when using bors
```

but I don't think it's an issue since these types of branches are only used by bots and we are allowing them with this PR.

ubiratansoares (approved):

PR #2436, opened 2026-04-21
rename branch-protections to rulesets

marcoieni:
rename `branch-protections` to `rulesets` in the repositories toml files.

v1 API is unchanged to be backwards compatible.


ubiratansoares (approved):
🚀

PR #2437, opened 2026-04-21
feat(repos/rustup): add protection rule for `release/*` branches

rami3l:
Part of https://github.com/rust-lang/rustup/issues/4738; replaces #2372.

As discussed with @Kobzol in https://github.com/rust-lang/rustup/issues/4738#issuecomment-4276417728, given that the GitHub Merge Queue cannot be enabled for globs, the second best would be to have the following on those branches:

- "Require status checks" set to "true", to allow proper auto-merge (seems to be enabled already).
- ~~"Require a PR" and~~ "Require review" set to "false", to prevent spamming the repo for trivial fixes.
- ~~"Bypass actors" set to "t-rustup", to provide enough flexibility to the team regarding the release by reducing wait time when necessary. Combined with the previous point, this also seems to allow the team to push directly onto the backport branches as per https://github.com/rust-lang/rustup/issues/4738#issuecomment-4276417728.~~
- "Prevent creation" set to "false" since obviously `t-rustup` will have to create those branches from time to time.

Please feel free to point out if I have introduced anything wrong/redundant, many thanks in advance 🙏

### DIFF
--- a/repos/rust-lang/rustup.toml
+++ b/repos/rust-lang/rustup.toml
@@ -12,6 +12,12 @@ pattern = "main"
 ci-checks = ["conclusion"]
 merge-queue = { enabled = true, method = "rebase" }
 
+[[branch-protections]]
+pattern = "release/*"
+ci-checks = ["conclusion"]
+required-approvals = 0
+prevent-creation = false
+
 [environments.github-pages]
 
 [environments.www-deploy]


Kobzol (commented):
Now that I think about it, if you configure `allowed-merge-teams`, it will actually override `prevent-creation = true` and `pr-required = true`. So the last two items are sort of redundant, but they shouldn't hurt anything either, I think.

`team` doesn't consider `pr-required = false` with non-empty `ci-checks` to be valid, so it gives a CI error now. We can change that.

Btw, I don't want to dictate which workflow you should be using, but I was wondering about:

> to prevent spamming the repo for trivial fixes.

Do you really make changes to the release branches so often? I think it's better to have a change trail in the form of PRs for any change, plus if you go through a PR, it won't be possible to push a fix to the release branch which doesn't pass CI, which is always too easy to do when it is not gated :)

rami3l:
@Kobzol I guess both the CI and your comment is guiding me towards that direction, so `pr-required = true` it is. Will fix...

rami3l:
@Kobzol I think we should be pretty ready now, right?

Kobzol (commented):
Yeah, looks good :) (I don't have approval rights here)

ubiratansoares (approved):

PR #2445, opened 2026-04-23
gws: add diff for google accounts

ubiratansoares:
First PR to enable diffs betwen our configuration and Google Workspace state. 

The whole feature would be a much bigger PR, so I decided to split it. This PR looks big but it brings a non negletable amount of boilterplate, so hopefully it is not hard to read. I hope the code style is familiar enough compared with other services we also diff and sync.

For now, this PR adds support for diffing google workspace **_users_**, supporting `creation` and `deletion` operations. Everything is still detached from the app entrypoint, and integration will come after we are able to run live dry-runs in CI (which requires further configuration on Github Actions).

To make it sure we can define and safely test some rules we want to encode as part of the diff logic, I added a fake GWS API similar approach similar to what we use in `github`. 

Towards #2363

### DIFF
--- a/rust_team_data/src/v1.rs
+++ b/rust_team_data/src/v1.rs
@@ -26,6 +26,15 @@ pub struct Team {
     pub github: Option<TeamGitHub>,
     pub website_data: Option<TeamWebsite>,
     pub roles: Vec<MemberRole>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub google_workspace_saml_group: Option<bool>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct GoogleWorkspace {
+    pub first_name: String,
+    pub last_name: String,
+    pub account_handle: String,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
@@ -36,6 +45,8 @@ pub struct TeamMember {
     pub is_lead: bool,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
     pub roles: Vec<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub google_workspace: Option<GoogleWorkspace>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -76,13 +76,22 @@ pub(crate) struct Funding {
     github_sponsors: bool,
 }
 
-#[allow(dead_code)]
-#[derive(serde::Deserialize, Debug)]
+#[derive(serde::Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
 pub(crate) struct GoogleWorkspace {
-    first_name: String,
-    last_name: String,
-    account_handle: String,
+    pub first_name: String,
+    pub last_name: String,
+    pub account_handle: String,
+}
+
+impl From<GoogleWorkspace> for rust_team_data::v1::GoogleWorkspace {
+    fn from(gws: GoogleWorkspace) -> Self {
+        rust_team_data::v1::GoogleWorkspace {
+            first_name: gws.first_name,
+            last_name: gws.last_name,
+            account_handle: gws.account_handle,
+        }
+    }
 }
 
 #[allow(dead_code)]
@@ -157,8 +166,8 @@ impl Person {
         &self.permissions
     }
 
-    pub(crate) fn google_workspace(&self) -> &Option<GoogleWorkspace> {
-        &self.google_workspace
+    pub(crate) fn google_workspace(&self) -> Option<&GoogleWorkspace> {
+        self.google_workspace.as_ref()
     }
 
     pub(crate) fn validate(&self) -> Result<(), Error> {
@@ -194,7 +203,6 @@ impl std::fmt::Display for TeamKind {
     }
 }
 
-#[allow(dead_code)]
 #[derive(serde::Deserialize, Debug)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
 pub(crate) struct Team {
--- a/src/static_api.rs
+++ b/src/static_api.rs
@@ -7,7 +7,9 @@ use anyhow::{Context as _, Error, ensure};
 use indexmap::IndexMap;
 use log::info;
 use rust_team_data::v1;
-use rust_team_data::v1::{BranchProtectionMode, Crate, CrateTeamOwner, RepoMember};
+use rust_team_data::v1::{
+    BranchProtectionMode, Crate, CrateTeamOwner, GoogleWorkspace, RepoMember,
+};
 use std::collections::HashMap;
 use std::path::Path;
 
@@ -531,14 +533,18 @@ fn convert_teams<'a>(
 
         let leads = team.leads();
         let mut members = Vec::new();
-        for github_name in &team.members(data)? {
+        for github_name in team.members(data)? {
             if let Some(person) = data.person(github_name) {
                 members.push(v1::TeamMember {
                     name: person.name().into(),
                     github: (*github_name).into(),
                     github_id: person.github_id(),
                     is_lead: leads.contains(github_name),
-                    roles: website_roles.get(*github_name).cloned().unwrap_or_default(),
+                    roles: website_roles.get(github_name).cloned().unwrap_or_default(),
+                    google_workspace: person
+                        .google_workspace()
+                        .cloned()
+                        .map(GoogleWorkspace::from),
                 });
             }
         }
@@ -557,6 +563,10 @@ fn convert_teams<'a>(
                         .get(alum.github.as_str())
                         .cloned()
                         .unwrap_or_default(),
+                    google_workspace: person
+                        .google_workspace()
+                        .cloned()
+                        .map(GoogleWorkspace::from),
                 });
             }
         }
@@ -606,6 +616,7 @@ fn convert_teams<'a>(
                     description: role.description.clone(),
                 })
                 .collect(),
+            google_workspace_saml_group: team.google_workspace_saml_group(),
         };
         team_map.insert(team.name().into(), team_data);
     }
--- a/src/sync/gws/api.rs
+++ b/src/sync/gws/api.rs
@@ -1,0 +1,41 @@
+use crate::sync::gws::RUST_LANG_GWS_DOMAIN;
+use async_trait::async_trait;
+use rust_team_data::v1::GoogleWorkspace;
+
+/// https://developers.google.com/workspace/admin/directory/reference/rest/v1/groups
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct Group {
+    pub name: String,
+    pub email: String,
+}
+
+/// https://developers.google.com/workspace/admin/directory/reference/rest/v1/users#UserName
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub(crate) struct UserName {
+    pub given_name: String,
+    pub family_name: String,
+}
+
+/// https://developers.google.com/workspace/admin/directory/reference/rest/v1/users
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub(crate) struct User {
+    pub name: UserName,
+    pub primary_email: String,
+}
+
+impl From<&GoogleWorkspace> for User {
+    fn from(gws: &GoogleWorkspace) -> Self {
+        Self {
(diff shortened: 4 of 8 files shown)


src/sync/mod.rs
```diff
@@ -78,6 +79,9 @@ pub async fn run_sync_team(
                     diff.apply(&sync).await?;
                 }
             }
+            "google-workspace" => {
+                println!("Nothing to diff");
```
marcoieni:
```suggestion
                println!("google-workspace: Nothing to diff");
```
Otherwise, CI will print just "Nothing to diff" and that's confusing.

Btw I assume the diff itself will be implemented in a later PR 👍

rust_team_data/src/v1.rs
```diff
@@ -36,6 +45,7 @@ pub struct TeamMember {
     pub is_lead: bool,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
     pub roles: Vec<String>,
+    pub google_workspace: Option<GoogleWorkspace>,
```
marcoieni, line 49:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub google_workspace: Option<GoogleWorkspace>,
```
Otherwise the response contains `"google_workspace": null`

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+                None => None,
+                Some(opt_in) => {
+                    if opt_in {
+                        Some(&team.members)
+                    } else {
+                        None
+                    }
+                }
+            })
+            .flatten()
```
marcoieni:
```suggestion
            .filter(|team| team.google_workspace_saml_group.unwrap_or_default())
            .flat_map(|team| &team.members)
```
obtained by just asking codex 5.5 low reasoning "make this more idiomatic"

rust_team_data/src/v1.rs
```diff
@@ -26,6 +26,15 @@ pub struct Team {
...
     pub roles: Vec<MemberRole>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub google_workspace_saml_group: Option<bool>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct GoogleWorkspace {
+    pub name: String,
+    pub surname: String,
+    pub account_handle: String,
```
marcoieni, line 37: why do we have a mismatch between these fields and the fields in the schema?
Isn't it easier if we call them the same? 🤔

src/static_api.rs
```diff
@@ -557,6 +566,13 @@ fn convert_teams<'a>(
                         .get(alum.github.as_str())
                         .cloned()
                         .unwrap_or_default(),
+                    google_workspace: person.google_workspace().clone().map(|gws| {
+                        GoogleWorkspace {
+                            name: gws.first_name,
+                            surname: gws.last_name,
+                            account_handle: gws.account_handle,
```
marcoieni: this code is duplicated from above. Maybe we can have a [From](https://doc.rust-lang.org/std/convert/trait.From.html) implementation or something like this?

src/static_api.rs
```diff
@@ -539,6 +541,13 @@ fn convert_teams<'a>(
                     github_id: person.github_id(),
                     is_lead: leads.contains(github_name),
                     roles: website_roles.get(*github_name).cloned().unwrap_or_default(),
+                    google_workspace: person.google_workspace().clone().map(|gws| {
```
marcoieni: I noticed that the function `google_workspace()` returns `&Option<Something>`, but `Option<&Something>` is more idiomatic.

We could change the function like this:
```
    pub(crate) fn google_workspace(&self) -> Option<&GoogleWorkspace> {
        self.google_workspace.as_ref()
    }
```

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+        diffs.extend(users_to_remove);
+
+        Ok(diffs)
+    }
+
+    fn matches_by_user_email(&self, email: &str, users: &[User]) -> bool {
+        users.iter().any(|account| account.primary_email == email)
+    }
+
+    fn matches_by_saml_group(&self, target: &Group, groups: &[Group]) -> bool {
```
marcoieni: this function is unused

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+                    }
+                }
+            })
+            .flatten()
+            .collect::<Vec<_>>();
+
+        let declared_users = members_from_gws_enabled_teams
+            .iter()
+            .filter_map(|member| member.google_workspace.as_ref().map(User::from))
+            .collect::<Vec<_>>();
```
marcoieni:
```suggestion
        let mut seen_primary_emails = BTreeSet::new();
        let declared_users = members_from_gws_enabled_teams
            .iter()
            .filter_map(|member| member.google_workspace.as_ref().map(User::from))
            .filter(|user| seen_primary_emails.insert(user.primary_email.clone()))
            .collect::<Vec<_>>();
```
Should we deduplicated by some kind of criteria (eg email)?
So that we don't try to create users multiple times?

rust_team_data/src/v1.rs
ubiratansoares: No particular reason, I can make them equal for simplicity! :)

src/sync/mod.rs
ubiratansoares: Indeed, but happy to change it to make it clear

src/sync/gws/mod.rs
ubiratansoares: I did not focus on dedup because, in practice, the number of users will be quite small (~ 30 after everyone onboarded). But happy to try a better approach. Commit to follow

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+
+        let mut diffs = Vec::new();
+        diffs.extend(users_to_create);
+        diffs.extend(users_to_remove);
+
+        Ok(diffs)
+    }
+
+    fn matches_by_user_email(&self, email: &str, users: &[User]) -> bool {
```
marcoieni: this function doesn't use self, so it can be a free function

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+            google_users: google_accounts_diff,
+        };
+        Ok(diff)
+    }
+
+    async fn diff_groups(&self) -> anyhow::Result<Vec<GoogleGroupDiff>> {
+        Ok(vec![])
+    }
+
+    async fn diff_users(&self) -> anyhow::Result<Vec<GoogleUserDiff>> {
```
marcoieni: this function is async but doesn't `await`.
same `diff_groups`

src/sync/gws/api.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct User {
+    pub name: UserName,
+    pub primary_email: String,
+}
+
+impl From<&GoogleWorkspace> for User {
+    fn from(gws: &GoogleWorkspace) -> Self {
+        Self {
+            primary_email: format!("{}@rust-lang.org", gws.account_handle),
```
marcoieni: `@rust-lang.org` is used twice in this PR. Let's use a constant.

src/sync/gws/mod.rs
ubiratansoares: Added some optimizations [here](https://github.com/rust-lang/team/pull/2445/commits/2e026df046eedea8014db9d3d782fcfcde6611c7) (I can rebase iteractively afterwards if you agree with them)

Essentially, I now dedup by email references and I also leverage [Iterator::chain](https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.chain) to optimize creating `GoogleUserDiff` instances in a single pass. As a nice side effect, I no longer need `matches_by_user_email` function. 

This implementation also will make it easier to implement Edit diffs in an upcoming PR. Hope this is close to what you had in mind :)

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+            google_users: google_accounts_diff,
+        };
+        Ok(diff)
+    }
+
+    async fn diff_groups(&self) -> anyhow::Result<Vec<GoogleGroupDiff>> {
+        Ok(vec![])
+    }
+
+    fn diff_users(&self) -> anyhow::Result<Vec<GoogleUserDiff>> {
```
marcoieni, line 68: this function uses hashsets, so the vec returns unordered results.
That's bad because when testing (eg in `diff_spots_user_creation`) we compare using vectors.

Suggested solution: by using BTreeSet this function would return a vec ordered consistently.

(issue spot thanks to AI review)

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+        let diff = GoogleWorkspaceDiff {
+            google_groups: google_groups_diff,
+            google_users: google_accounts_diff,
+        };
+        Ok(diff)
+    }
+
+    async fn diff_groups(&self) -> anyhow::Result<Vec<GoogleGroupDiff>> {
+        Ok(vec![])
+    }
```
marcoieni, line 66: I assume this function will be async eventually.
But diff_users is not 🤔

ubiratansoares: A leftover here, yeah. Most likely we won't need any async here. 😅

I'll update!

src/sync/gws/mod.rs
ubiratansoares: Great suggestion, I'll change that

src/sync/gws/api.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/// https://developers.google.com/workspace/admin/directory/reference/rest/v1/users#UserName
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub(crate) struct UserName {
+    pub given_name: String,
+    pub family_name: String,
+}
+
+///https://developers.google.com/workspace/admin/directory/reference/rest/v1/users
```
marcoieni:
```suggestion
/// https://developers.google.com/workspace/admin/directory/reference/rest/v1/users
```

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+
+/// A diff between the team repo and the state on Google Workspace
+#[allow(dead_code)]
+#[derive(Debug)]
+pub(crate) struct GoogleWorkspaceDiff {
+    google_groups: Vec<GoogleGroupDiff>,
+    google_users: Vec<GoogleUserDiff>,
+}
+
+/// The engine that evaluate diffs between our current configuration and
```
marcoieni:
```suggestion
/// The engine that evaluates diffs between our current configuration and
```

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+    ) -> anyhow::Result<Self> {
+        let gws_users = gws_api_client.get_users().await?;
+        let sync = Self {
+            actual_users: gws_users,
+            configured_teams: teams,
+        };
+        Ok(sync)
+    }
+
+    pub(crate) async fn diff_all(&self) -> anyhow::Result<GoogleWorkspaceDiff> {
```
marcoieni:
```suggestion
    pub(crate) fn diff_all(&self) -> anyhow::Result<GoogleWorkspaceDiff> {
```
doesn't need to be async

src/schema.rs
```diff
@@ -76,13 +76,22 @@ pub(crate) struct Funding {
...
 pub(crate) struct GoogleWorkspace {
-    first_name: String,
-    last_name: String,
-    account_handle: String,
+    pub first_name: String,
+    pub last_name: String,
+    pub account_handle: String,
+}
+
+impl From<&GoogleWorkspace> for rust_team_data::v1::GoogleWorkspace {
```
marcoieni: from what I know the convention is that From consumes the type.

I confirmed by asking gemini the question "is the convention that impl From in rust takes a reference or a value?" and this is the answer:

Image

ubiratansoares: I went for `From<&GoogleWorkspace>` following an IDE suggestion; the trait signature read a bit weird afterwards, but since it was working, I thought it was ok.

Definetly learned something here :) I'll update the trait implementation.

src/sync/gws/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+                    .map(|u| GoogleUserDiff::Delete(u.clone())),
+            )
+            .collect();
+
+        Ok(diffs)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
marcoieni, line 105: This could be under the `tests/` directory, similar to github to avoid long files, but this can be done in a follow-up PR

marcoieni (approved):
🚀

PR #2002, opened 2025-09-27
Give permission to do try builds to wg-triage

Kobzol:
Discussed in https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/.E2.9C.94.20start.20a.20try-job.20on.20my.20PR.20please/with/541811337.

PR #2449, opened 2026-04-28
Give crater permissions to the triage team

cuviper:
We in the release team have appreciated the triage team's help with beta crater reports, especially @theemathas, and they may do more with the ability to queue crater jobs themselves.

r? @Mark-Simulacrum

### DIFF
--- a/teams/triage.toml
+++ b/teams/triage.toml
@@ -42,6 +42,9 @@ members = [
 ]
 alumni = []
 
+[permissions]
+crater = true
+
 [website]
 name = "Triage Team"
 description = "Triaging repositories under the rust-lang organisation"


cyrgani:
https://github.com/rust-lang/team/pull/2002 was a previous proposal for this (but only for `try` permissions), which was declined.

cuviper:
Well at least for beta crater runs, they don't need the ability to create new builds, `try` or otherwise.

jackh726:
I have a general negative sentiment on this, echoing [@jieyouxu's concern](https://github.com/rust-lang/team/pull/2002#issuecomment-3342127383). The bar for triage membership is low. I like the idea of it, but expanding permissions more than what is already allowed (which, tbh, I have thought a ton about) makes me weary.

I think it's better to do one or more of the following:
1) Limit membership to "trusted and active contributors"
2) Split triage into two teams, a "trusted" team and then a untrusted team (where the former can have try and crater permissions, probably move issues across repos, close issues, close PRs; the latter can label issues).
3) Make sure that people doing things like crater triage are part of a team that *does* have these permissions.

Mark-Simulacrum (approved):

Mark-Simulacrum:
I would personally continue to be hesitant about try permissions, but crater should be fairly safe. In principle you can still cause damage but we grant triage members permissions to do that anyway in other ways (and, honestly, I don't know how worried I am - the worst case is pretty benign here, I think).

jieyouxu:
Another possible approach (without having to explicitly recharter triage team into two teams just yet, which I last understand was something that's in the works) could also be something like a `release/crater-triage` subteam (who don't receive the 'full' `release` team permissions but do get `try` / `crater` permissions, e.g. I would trust `theemathas` with `try` permissions). Up to y'all.

cuviper:
Withdrawing for now, given the specific member addition in #2460. (Congrats and welcome!)

PR #2455, opened 2026-05-01
Add GSoC 2026 contributors to the `gsoc-contributors` team

Kobzol:
Welcome @sa4dus @lapla-cogito @sgasho @moabo3li @CoCo-Japan-pan @NONnonHere @dybucc @vishruth-thimmaiah @Shourya742 @xonx4l @zedddie @Walnut356 @Cloud0310! :tada:

### DIFF
--- a/people/Cloud0310.toml
+++ b/people/Cloud0310.toml
@@ -1,0 +1,4 @@
+name = "Cloud"
+github = "Cloud0310"
+github-id = 60375730
+zulip-id = 1075679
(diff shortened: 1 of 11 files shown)


src/validate.rs
jieyouxu, line 1: Question: should this use `excluded-people` for the zulip group instead? Last recall I changed the validation in https://github.com/rust-lang/team/pull/1977

jieyouxu (dismissed):
Team changes looks good

src/validate.rs
Kobzol: Ah, I recalled that there was something like this, but I forgot about this field. Yeah, that is better, thanks!

Kobzol:
Did something better, found the missing Zulip ID :)

jieyouxu (approved):

PR #2463, opened 2026-05-13
Configure `T-bootstrap` to use FCP

jieyouxu:
Noticed in https://github.com/rust-lang/rust-forge/pull/1040#issuecomment-4438100535.

(I'm not actually sure if this is sufficient to allow T-bootstrap to use FCP?)

cc @Mark-Simulacrum (bootstrap lead)
FYI @rust-lang/bootstrap


Mark-Simulacrum (approved):
I'm not sure either but this seems reasonable.

PR #2471, opened 2026-05-15
Add types team to rust-forge repo

clarfonthey:
For what it's worth, I have absolutely no authority to make this change and want to underline that I'm only doing this to promote a discussion.

Since rust-lang/rust-forge#1040 was limited to collaborators, but the FCP includes the types team, I noted that the types team is not actually included in the list of people who count as maintainers for that repo. It makes sense to include them even though it's likely all the relevant people are already on at least one of the other teams required to have access. It's just that the PR got heated enough to need to be locked down, and I want to make sure everyone who does have a stake in the discussion can do so.

### DIFF
--- a/repos/rust-lang/rust-forge.toml
+++ b/repos/rust-lang/rust-forge.toml
@@ -21,6 +21,7 @@ edition = "maintain"
 release = "maintain"
 rustdoc = "maintain"
 triagebot = "maintain"
+types = "maintain"
 rustc-dev-guide = "maintain"
 
 [[rulesets]]


jieyouxu (commented):
I'm surprised that types doesn't already have this persmission, this is fine if any of the types co-leads approve.
cc @jackh726 @lcnr

jieyouxu:
> For what it's worth, I have absolutely no authority to make this change and want to underline that I'm only doing this to promote a discussion.

Proposing changes to fix such missing access / access oversight is completely fine.

clarfonthey:
> Proposing changes to fix such missing access / access oversight is completely fine.

I figured so, and that's why I proposed it; I just wanted to point out that this needs external approval so that the folks who do know who needs to discuss this can make it happen.

jackh726 (approved):

PR #2472, opened 2026-05-16
Give some basic repo access (back) to moderators

steffahn:
As long as the basic concerns for moderator access to issue/PR discussions aren’t really solved any other way, I’d like to ask if we can add back write access to `rust-lang/rust` to moderators since it’s by far the highest-traffic repo. (This thus undoes part of #1848.)

Given the moderation relevance for current LLM-policy discussions, I’d also like to include `rust-lang/rust-forge`. (This one is new, not part of something that #1848 removed.)


jieyouxu (approved):
This makes sense for venue mods (I think most of current set of venue mods have write access from being on other teams).

Since this mosifies r-l/r an infra-admin should also give approval.

Mark-Simulacrum (approved):

PR #2477, opened 2026-05-24
add funding/private zulip stream

nikomatsakis:


Mark-Simulacrum (approved):
I'm not sure if we support channel creation, if this bounces we'll need to create it, add rust-lang-owner, and then try again.

jieyouxu:
Yeah, @nikomatsakis will probably need to create the private stream, invite rust-lang-owner, then we can requeue this PR

PR #2482, opened 2026-05-25
Rename `gsoc-contributors` team to `mentees` team and update mentorship membership

Kobzol:
To generalize the concept, because also have OSPP and Outreachy. I also updated the membership of the mentor teams and consolidated the OSPP team into the `mentees` team, and added this year's Outreachy contributors.

An alternative name could be `interns`.

CC @jackh726


jackh726 (approved):
Thanks for doing this @Kobzol ❤

PR #2496, opened 2026-06-02
compiler: update fcp list

davidtwco:
Per team-internal Zulip thread soliciting interest.

cc @BoxyUwU


jieyouxu (approved):
This was discussed internally in the compiler team private stream.

PR #2498, opened 2026-06-02
add assumptions on binders project & repo

BoxyUwU:
cc https://rust-lang.github.io/rust-project-goals/2026/assumptions_on_binders.html

I want a repo so I can store design docs/misc thoughts and also track all the high level bits of work and goals that are covered under this work. Not sure who exactly should sign off on this :sparkles: I don't know that actually creating the associated team is useful? I just did so so that if/when other contributors start working on this they can get shoved into a group to recognise that :thinking: 

cc @lcnr

### DIFF
--- a/repos/rust-lang/project-assumptions-on-binders.toml
+++ b/repos/rust-lang/project-assumptions-on-binders.toml
@@ -1,0 +1,12 @@
+org = "rust-lang"
+name = "project-assumptions-on-binders"
+description = ""
+homepage = "https://rust-lang.github.io/project-assumptions-on-binders/"
+bots = ["rustbot"]
+
+[access.teams]
+project-assumptions-on-binders = "maintain"
+all = "write"
+
+[environments.github-pages]
+
--- a/teams/assumptions-on-binders.toml
+++ b/teams/assumptions-on-binders.toml
@@ -1,0 +1,24 @@
+name = "project-assumptions-on-binders"
+kind = "project-group"
+subteam-of = "types"
+
+[people]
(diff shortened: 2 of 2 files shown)


repos/rust-lang/project-assumptions-on-binders.toml
```diff
@@ -0,0 +1,12 @@
+org = "rust-lang"
+name = "project-assumptions-on-binders"
+description = ""
+homepage = "https://rust-lang.github.io/project-assumptions-on-binders/"
+bots = ["rustbot"]
+
+[access.teams]
+project-assumptions-on-binders = "maintain"
+all = "write"
```
BoxyUwU, line 9: my understanding from the docs in this repo is that maintain gives you approve PRs rights and write doesn't? but rfcs repo has all = "write" and you can approve PRs there so I don't fully understand the distinction 😅 

my intention here is just that i really dont care who can do what in this repo and trust people to not mess with things so i want to bias towards just letting people do whatever

jackh726 (approved):

repos/rust-lang/project-assumptions-on-binders.toml
ehuss: Sorry for the confusion. I can try to post a PR to clarify things.

The distinction was largely historical. It only really mattered when using bors, and nowadays there's only really one repo that does that (rust-lang/rust). Also, this differed in the past because we used protected branches. People with "maintain" access can push to protected branches. However, we have since switched to rulesets which IIUC don't allow that without specific bypass rules.

PR #2507, opened 2026-06-04
Add myself to `cloud-compute` team

Human9000-bit:
as discussed with @Kivooeo


Kivooeo (approved):
Vouching for Egor as it was discussed internally on Zulip. He indeed did a nice amount of work and he asked me if we can provide him access to improve his workflow, which I'm happy with!

jackh726 (approved):

PR #2510, opened 2026-06-05
add working group for Field Projections

BennoLossin:
Cc @tmandry @Nadrieril @dingxiangfei2009

### DIFF
--- a/teams/wg-field-projections.toml
+++ b/teams/wg-field-projections.toml
@@ -1,0 +1,28 @@
+name = "wg-field-projections"
+subteam-of = "lang"
+kind = "working-group"
+
+[people]
+leads = [
+    "BennoLossin",
+    "tmandry",
+]
+members = [
+    "BennoLossin",
+    "tmandry",
+    "Nadrieril",
+    "dingxiangfei2009",
+]
+alumni = [
+]
+
+[[github]]
+orgs = ["rust-lang"]
+
+[website]
+name = "Field Projections Working Group"
+description = "Designing a language feature for field projections in Rust"
+zulip-stream = "t-lang/custom-refs"
+
+[[zulip-groups]]
+name = "WG-field-projections"
--- a/people/BennoLossin.toml
+++ b/people/BennoLossin.toml
@@ -2,3 +2,4 @@ name = "Benno Lossin"
 github = "BennoLossin"
 github-id = 94611769
 zulip-id = 458499
+email = "benno.lossin@proton.me"


BennoLossin:
ugh why is the github default a merge...

tmandry (approved):

Nadrieril:
Oh I thought working groups weren't a thing anymore. Fun :)

teams/wg-field-projections.toml
```diff
@@ -0,0 +1,27 @@
+name = "field-projections"
+subteam-of = "lang"
```
jieyouxu: Question: just to double-check: is this supposed to be a "proper" *team*, or is this supposed to be a "working group"?

The team file name and PR says "wg-field-projections" / WG, was this intended to have `kind = "working-group"`?

See https://github.com/rust-lang/team/blob/main/docs/toml-schema.md#teams

(NB: one of the viable options is also to *start out* as a working group, but can evolve to a proper subteam with a rechartering. Just depends on what is wanted here.)

teams/wg-field-projections.toml
```diff
@@ -0,0 +1,27 @@
...
+[[github]]
+orgs = ["rust-lang"]
+
+[website]
+name = "Field Projections Working Group"
+description = "Designing a language feature for field projections in Rust"
+zulip-stream = "t-lang/custom-refs"
+
+[[zulip-groups]]
+name = "field-projections"
```
jieyouxu: Suggestion: by convention, zulip groups typically use `T-*` prefix for teams, and `WG-*` prefix for working groups. Maybe this should be `WG-field-projections` for consistency?

teams/wg-field-projections.toml
Nadrieril: Definitely just a working group

teams/wg-field-projections.toml
BennoLossin: I looked at `wg-const-eval` to see how they are done and there's no `WG-` prefix there :)

I'll add one here though

BennoLossin:
Thanks for taking a look Jieyou, let me know if anything else looks off, first time creating a working group :)

teams/wg-field-projections.toml
Nadrieril: Could it be that the prefix is added automatically? Because `@WG-const-eval` exists on Zulip

Nadrieril: And so does `@const-eval`, interestingly

Nadrieril: They have different descriptions x) I think they're not the right example to follow

jieyouxu (approved):

PR #2512, opened 2026-06-09
Add custom properties

### LINKED ISSUE #2504: Allow setting repository custom properties
For https://github.com/rust-lang/infra-team/issues/260 we need repositories custom properties. Because we'll run https://github.com/rust-lang/crabwatch on repositories that set a certain custom property (name to be decided) to `true`.

Tasks:
- Add the ability to set custom properties in repositories from the `team` repo
- Propose a name for this custom property
- Add the custom property to the `crabwatch` repository itself

API docs: https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2026-03-10

Sandijigs:
Adds support for ```[custom-properties]``` in the team repo's TOML, and sets ```crabwatch = true``` on rust-lang/crabwatch.
Booleans only for now. Properties on a repo but not declared in TOML are left alone.
Closes #2504.


src/sync/github/mod.rs
```diff
@@ -870,6 +881,41 @@ impl SyncGitHub {
...
+            // GitHub stores values as strings, even bools.
+            let expected = value.to_string();
+            let operation = match actual_by_name.get(name) {
+                // Missing on the repo, or value is null.
+                None | Some(None) => CustomPropertyDiffOperation::Create(expected),
+                Some(Some(actual)) if actual != &expected => {
+                    CustomPropertyDiffOperation::Update(actual.clone(), expected)
+                }
+                Some(Some(_)) => continue,
+            };
```
ubiratansoares: Perhaps is room to make this code more concise with some pattern matching tricks. I'd recommend having a look on how to change

```
let operation = match actual_by_name.get(name)
```

in order to get rid of the nested Option value :)

tests/static-api/_expected/v1/repos/archived_repo.json
```diff
@@ -35,5 +35,6 @@
   "environments": {},
   "archived": true,
   "private": false,
-  "auto_merge_enabled": true
+  "auto_merge_enabled": true,
+  "custom_properties": {}
 }
```
ubiratansoares: It's probaly a good idea adding an empty line in these files, so we keep the standard

src/sync/github/mod.rs
```diff
@@ -870,6 +881,41 @@ impl SyncGitHub {
         Ok(ruleset_diffs)
     }
 
+    async fn diff_custom_properties(
```
ubiratansoares, line 912: I missed some tests that actually prove this function works as intended in `sync/github/tests/mod.rs`.

I propose adding at least two cases : 

- adding a propery to a repo
- removing a property from a repo

rust_team_data/src/v1.rs
```diff
@@ -195,6 +196,8 @@ pub struct Repo {
     // Is the GitHub "Auto-merge" option enabled?
     // https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request
     pub auto_merge_enabled: bool,
+    #[serde(default)]
+    pub custom_properties: BTreeMap<String, bool>,
```
ubiratansoares: In general we've been using `IndexMap` to represent key-value pairs in this file. Do you have any particular reason to go with BTreeMap instead?

ubiratansoares (commented):
@Sandijigs thanks for this PR, looks we are in the right direction here. Added a few comments :)

rust_team_data/src/v1.rs
```diff
@@ -195,6 +196,8 @@ pub struct Repo {
     // Is the GitHub "Auto-merge" option enabled?
     // https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request
     pub auto_merge_enabled: bool,
+    #[serde(default)]
```
Kobzol: The `#[serde(default)]` annotation is not necessary, because the generator (the team API CI) will be updated sooner than the consumers (the crates that depend on the `team` crate).

src/sync/github/mod.rs
```diff
@@ -2295,6 +2376,50 @@ enum RulesetDiffOperation {
...
+                )
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+enum CustomPropertyDiffOperation {
+    Create(String),
+    Update(String, String), // old, new
```
Kobzol: This should perhaps also contain `Delete`, in case the custom property is removed from the team config.

It would mean that team is the ground truth for all our custom properties, although it would also delete all custom properties that are not set in `team`.

@ubiratansoares Do you want team to be the ground truth for all custom properties? If it is not, then once a property is created through team, it won't ever be deleted again.

docs/toml-schema.md
```diff
@@ -527,6 +527,23 @@ branches = ["develop", "staging"]
...
+
+Only boolean values are supported.
+
+[Repository custom properties]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/managing-custom-properties-for-a-repository
+
+```toml
+# Repository custom properties (optional)
+[custom-properties]
+# Set a property name to a boolean value
+crabwatch = true
```
Kobzol, line 571: The property values on GitHub are strings, and the code currently converts bools to a string via its `Display` impl, which is a bit opaque and potentially fragile. Maybe we could just expose the value as a string?

Kobzol (commented):
Left a few comments :)

src/sync/github/mod.rs
ubiratansoares: > Do you want team to be the ground truth for all custom properties?

That's an interesting question, my gut feeling says _yes_, since I see these custom properties as tool to enable governance-related automations over Github repos or Github orgs, and `team` is the place where we want represent and enable such stuff with IaC.

Kobzol: I agree, but we should be then careful with the first sync (or ideally a dry-run executed manually) to ensure that we don't delete any existing custom properties; those would have to be backported into `team` first.

ubiratansoares: I'll check whether there are any custom properties out there, so we don't break anything by mistake here

rust_team_data/src/v1.rs
Sandijigs: Actually there is no strong reason. BTreeMap gives deterministic key ordering, but IndexMap fits the file's convention better. I will switched it

docs/toml-schema.md
Sandijigs: That's a fair point. The reason I  went with bool is that @marcoieni suggested it earlier. TOML supports it natively, and the conversion to string happens on the Rust side. 
I can switch to strings throughout if you'd rather, but probably worthchecking with @marcoieni  first so we don't end up bouncing between the two.

src/sync/github/mod.rs
Sandijigs: Refactored to flatten the nested Option using ```.and_then(|v| v.as_deref())```.
The match now has three flat arms instead of three nested ones.

tests/static-api/_expected/v1/repos/archived_repo.json
Sandijigs: Updated the JSON writer in `static_api.rs` to add a trailing newline.

src/sync/github/mod.rs
Sandijigs: Added two tests, one for adding a property and one for removing one as suggested. Both run through the diff function end to end. Also extended the mock so tests can set what's on the GitHub side independently from the TOMLside (similar to how rulesets/environments work).

rust_team_data/src/v1.rs
Sandijigs: Removed. Makes sense given the generator updates before consumers.

src/sync/github/mod.rs
Sandijigs: Added the Delete variant. The diff emits a. Delete for any property on GitHub that isn't declared in TOML, so team becomes the source of truth. The apply sends a null value through the existing PATCH endpoint, which is GitHub's way of unsetting.

docs/toml-schema.md
marcoieni: Yeah, I was wrong on this. The fact that the property crabwatch can be only `true` or `false` isn't true for all properties, so it makes sense to use strings. Sorry about this!

src/sync/github/mod.rs
ubiratansoares: > _I'll check whether there are any custom properties out there, so we don't break anything by mistake_

@Sandijigs @Kobzol just to confirm : I ran a custom script and confirmed that no repos actually use custom properties (for `rust-lang` org)

Kobzol: Ok, good to know. In that case we can directly switch to using `team` as a source of truth (unless repos in some of our other orgs use them.. :) ).

docs/toml-schema.md
```diff
@@ -527,6 +527,23 @@ branches = ["develop", "staging"]
 # No branch or tag patterns specified - any branch or tag can deploy
 ```
 
+### Repository custom properties
+
+[Repository custom properties] are values set on a repository to opt it into org-wide tooling. The property must first be defined at the organization level.
+
+Only boolean values are supported.
+
+[Repository custom properties]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/managing-custom-properties-for-a-repository
```
ubiratansoares: This link returns 404, probably you meant

https://docs.github.com/en/organizations/managing-organization-settings/managing-custom-properties-for-repositories-in-your-organization

?

ubiratansoares (commented):
@Sandijigs looks good to me, one small thing from my end

docs/toml-schema.md
Sandijigs: done , the link has been replace with the working url

src/sync/github/api/mod.rs
```diff
@@ -746,3 +746,15 @@ pub(crate) struct BranchPolicy {
 fn default_branch_policy_type() -> String {
     "branch".to_string()
 }
+
+/// A GitHub repository custom property. Values are strings even for booleans.
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub(crate) struct CustomPropertyValue {
+    pub(crate) property_name: String,
+    pub(crate) value: Option<String>,
```
ubiratansoares: Right now we are modelling propertie values as optional string fields, but actually what we get here in the API response will depend on specifics of custom property settings:

Image

As an example, when we set the property type as `multi-select` we'll get an array of strings in the API response. Tested with one my repos:

```shell
gh api \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2026-03-10" \
  /repos/dotanuki-labs/testbed/properties/values

[
  {
    "property_name": "Category",
    "value": [
      "DevOps"
    ]
  }
]
```

Image

Perhaps we don't need to support all types of custom properties to get started, but we must clarify that we support only text-field like properties for now

src/sync/github/api/write.rs
```diff
@@ -696,4 +696,27 @@ impl GitHubWrite {
...
+        org: &str,
+        repo: &str,
+        property: &CustomPropertyValue,
+    ) -> anyhow::Result<()> {
+        debug!(
+            "Setting custom property '{}' on '{}/{}'",
+            property.property_name, org, repo
+        );
+        if !self.dry_run {
+            // REST API: PATCH /repos/{owner}/{repo}/properties/values
```
ubiratansoares, line 754: Since custom properties must be pre-defined at org-level to be applied, I think we should capture whether we are able to apply the custom property in the dry-run

docs/toml-schema.md
```diff
@@ -527,7 +528,25 @@ branches = ["develop", "staging"]
 # No branch or tag patterns specified - any branch or tag can deploy
 ```
 
+### Repository custom properties
+
+[Repository custom properties] are values set on a repository to opt it into org-wide tooling. The property must first be defined at the organization level.
+
+Only boolean values are supported.
```
ubiratansoares: Please check my comment on custom property types, we should elaborate better here

ubiratansoares (commented):
@Sandijigs Some additional considerations after trying out custom properties in a org/repo I own.

src/sync/github/api/write.rs
Sandijigs: Added the check. When a TOML property isn't defined at the org level, the diff emits a `CannotApply` operation and the apply step skips the API call.
https://github.com/rust-lang/team/blob/dc12bfbe7ef374232865259f218b92e547eefa04/src/sync/github/mod.rs#L927-L935

src/sync/github/api/mod.rs
Sandijigs: for this I kept as Option<String> for now and updated the schema docs to note only text-type properties are supported.
https://github.com/rust-lang/team/blob/dc12bfbe7ef374232865259f218b92e547eefa04/docs/toml-schema.md?plain=1#L558

docs/toml-schema.md
Sandijigs: Done, added the note about text-only support.

docs/toml-schema.md
```diff
@@ -550,7 +551,25 @@ branches = ["develop", "staging"]
...
+[Repository custom properties]: https://docs.github.com/en/organizations/managing-organization-settings/managing-custom-properties-for-repositories-in-your-organization
+
+```toml
+# Repository custom properties (optional)
+[custom-properties]
+# Set a property name to a string value
+crabwatch = "true"
+```
+
+Properties set on GitHub but not declared here are left unchanged.
```
marcoieni: So how do I delete a value that I set here before? Manually?
I expect that if I delete the value from the toml file, the team repo deletes that.

Don't worry about retro compatibility. We can write all existing values in the team repo before merging this. 


```suggestion
Properties set on GitHub but not declared here are removed.
```
Maybe this is a more correct statement of the behavior of this PR?

docs/toml-schema.md
```diff
@@ -550,7 +551,25 @@ branches = ["develop", "staging"]
 # No branch or tag patterns specified - any branch or tag can deploy
 ```
 
+### Repository custom properties
+
+[Repository custom properties] are values set on a repository to opt it into org-wide tooling. The property must first be defined at the organization level.
+
+Values are stored as strings. Only text-type custom properties are supported for now.
```
```suggestion
Only text-type custom properties are supported for now.
```
Remove redundant implementation detail

docs/toml-schema.md
```diff
@@ -550,7 +551,25 @@ branches = ["develop", "staging"]
 # No branch or tag patterns specified - any branch or tag can deploy
 ```
 
+### Repository custom properties
+
+[Repository custom properties] are values set on a repository to opt it into org-wide tooling. The property must first be defined at the organization level.
```
```suggestion
[Repository custom properties] are values set on a repository to add metadata. For example, they are used to opt it into org-wide tooling. The property must first be defined at the organization level by an infrastructure admin.
```
- specify that custom properties can also be used for other things
- Make clear that a contributor can't do it. So that they know who to ask in case they want a new property.

marcoieni:
we should at least support also booleans additionally to strings, so that we can define the crabwatch label as a boolean from the beginning, so that we don't have to do migrations later.

Supporting only booleans is also ok.

I can work on this if you want.

EDIT: I asked gpt 5.5 to implement it and it looks good. I committed these changes.

marcoieni:
I created the custom property for the rust-lang org
image

Sandijigs:
> I created the custom property for the rust-lang org image

Thanks for pushing this through. I've pulled the changes locally and tests pass. Reading through the diff now to make sure I understand the design.

marcoieni:
I think this is ready to be merged. Can you reorganize your commits in a way that you like? If you prefer, I can squash them into a single one, too.

Sandijigs:
Great, I will squash all commits into a single one

Sandijigs:
It has been squashed to a single commit. Ready to merge.

marcoieni (dismissed):

marcoieni:
apparently github api doesn't support booleans for custom properties 😅 I will make some changes quickly, then try to merge again.

marcoieni (approved):

marcoieni:
https://github.com/rust-lang/crabwatch/custom-properties 🥳
image

PR #2519, opened 2026-06-15
rust-lang/rust: Enable `forking-renovate` bot

Turbo87:
This enables Renovate to pin GitHub Actions to SHAs and keep them up-to-date.

See https://github.com/rust-lang/rust/pull/155089#issuecomment-4262931121

r? @marcoieni

### DIFF
--- a/repos/rust-lang/rust.toml
+++ b/repos/rust-lang/rust.toml
@@ -2,7 +2,7 @@ org = "rust-lang"
 name = "rust"
 description = "Empowering everyone to build reliable and efficient software."
 homepage = "https://www.rust-lang.org"
-bots = ["bors", "craterbot", "glacierbot", "log-analyzer", "rustbot", "rfcbot", "rust-timer", "datadog"]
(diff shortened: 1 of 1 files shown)


marcoieni:
Before approving this we need to change https://github.com/rust-lang/rust/blob/main/.github/renovate.json5 to only update github actions. Otherwise renovate starts opening a lot of PRs.

The code examples here (of the renovate config) should be helpful: https://github.com/rust-lang/renovate#security-only-updates

Turbo87:
> Otherwise renovate starts opening a lot of PRs.

are you sure about that? https://github.com/rust-lang/rust/blob/main/.github/renovate.json5#L6 looks to me like it currently has everything disabled except for the lockfile updates 🤔

marcoieni:
oh, you are right. But we should also delete `lockFileMaintenance` imo, because that was broken I remember

Turbo87:
see https://github.com/rust-lang/rust/pull/157952

Turbo87:
with https://github.com/rust-lang/rust/pull/157952 merged, can we merge this one now too? :)

marcoieni (approved):

PR #2522, opened 2026-06-17
Rename project-stable-mir

makai410:
cc @celinval


makai410:
What would the bot do to the repo? Rename or create a new one? What we want is to rename the `project-stable-mir` repo to `rustc_public`.

teams/project-rustc-public.toml
```diff
@@ -19,10 +19,10 @@ alumni = [
 ]
 
 [website]
-name = "Stable MIR Project Group"
+name = "Rustc Librarification Project Group"
```
celinval: Maybe let's just call it Rustc Public Project Group. I think there was a librarification effort that predates rustc_public, which someone called it out during the rust week.

celinval (commented):
Sweet! Thanks for following up on this.

BTW, I think we'll also need to update the triagebot: https://github.com/rust-lang/rust/blob/9e2abe0c6ab27fcbb95c30695188a75776e2feb1/triagebot.toml#L1642

Kobzol:
Renaming the repo has to be done manually with an onfra admin, team would just create a new one (this PR will be merged right after the rename).

teams/project-rustc-public.toml
makai410: I think you probably mean this one: https://smallcultfollowing.com/babysteps/blog/2020/04/09/libraryification/ :D Makes sense, let me fix that

repos/rust-lang/rustc_public.toml
```diff
@@ -1,13 +1,12 @@
 org = "rust-lang"
-name = "project-stable-mir"
+name = "rustc_public"
```
marcoieni, line 2: to double check, why the `_`? we always use `-` in repo names as far as I can tell

makai410: our crate is `rustc_public`, I think this is like `rustc_codegen_cranelift`? We want this repo to be the home of the `rustc_public` crate on crates.io

marcoieni (approved):

marcoieni:
> Renaming the repo has to be done manually with an onfra admin, team would just create a new one (this PR will be merged right after the rename).

I renamed the repo and I'm merging this PR. I have to say this process is quite scary and error prone. Eg if I make a typo in the repo rename, the repo gets deleted 😅

I like more the terraform approach that keeps track of resources and detects when something gets renamed

PR #2525, opened 2026-06-19
create funding advisors team

nikomatsakis:
Create a private `funding/with-advisors` stream and modify the `funding@` email alias to include the advisors; `funding-private@` can be used if we want narrow-cast.

To be discussed in our @rust-lang/funding meeting this morning.

### DIFF
--- a/teams/funding-advisors.toml
+++ b/teams/funding-advisors.toml
@@ -19,5 +19,5 @@ description = "People coordinating closely with the Funding team"
 name = "T-funding-advisors"
 
 [[zulip-streams]]
-name = "t-funding/with-advisors"
+name = "funding/with-advisors"
 extra-teams = ["funding"]
--- a/teams/funding-advisors.toml
+++ b/teams/funding-advisors.toml
@@ -1,0 +1,21 @@
+name = "funding-advisors"
+subteam-of = "funding"
+
+[people]
+leads = []
+members = [
+    "jlizen",
+    "becrumbul",
+]
+alumni = []
+
+[website]
+name = "Funding advisors team"
+description = "People coordinating closely with the Funding team"
+
+[[zulip-groups]]
+name = "T-funding-advisors"
+
+[[zulip-streams]]
+name = "t-funding/with-advisors"
+extra-teams = ["funding"]
--- a/teams/funding-advisors.toml
+++ b/teams/funding-advisors.toml
@@ -6,6 +6,8 @@ leads = []
 members = [
     "jlizen",
     "becrumbul",
+    "m-ou-se",
+    "erikjee",
 ]
 alumni = []
 
--- a/teams/funding-advisors.toml
+++ b/teams/funding-advisors.toml
@@ -8,6 +8,7 @@ members = [
     "becrumbul",
     "m-ou-se",
     "erikjee",
+    "obi1kenobi",
 ]
 alumni = []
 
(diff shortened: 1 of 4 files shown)


Kobzol:
Sounds good! I would propose to also include @m-ou-se, as she has a lot of experience with finding funding for Rust maintainers, and she can represent the NLNet fund, with which we could cooperate.

Kobzol:
Technical note: we already have the `t-funding/outreach` stream, which essentially already has the same members as this PR. We should rename it to `t-funding/with-advisors` before merging this, to avoid creating another stream and losing history.

nikomatsakis:
> Sounds good! I would propose to also include @m-ou-se, as she has a lot of experience with finding funding for Rust maintainers, and she can represent the NLNet fund, with which we could cooperate.

Makes sense, we should include Eric Jonkers too. I'll reach out to those two.

nikomatsakis:
> Technical note: we already have the `t-funding/outreach` stream, which essentially already has the same members as this PR. We should rename it to `t-funding/with-advisors` before merging this, to avoid creating another stream and losing history.

OK, I have a few others edits to make

teams/funding-advisors.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[website]
+name = "Funding advisors team"
+description = "People coordinating closely with the Funding team"
+
+[[zulip-groups]]
+name = "T-funding-advisors"
+
+[[zulip-streams]]
+name = "t-funding/with-advisors"
```
Kobzol:
```suggestion
name = "funding/with-advisors"
```
Just did a double check and found that on Zulip it's named without the `t-` prefix :)

nikomatsakis: Oh geez. 🤦‍♀️

Kobzol (approved):

jlizen:
Is Predrag also included? He has given me extremely useful advice in my own fundraising, and I'd prefer to move those communications out of DMs.

jlizen:
Also, I think that Abi should probably be in there?

jieyouxu:
(Holding off approving in case there's more additions.)

nikomatsakis:
I'll add the two of them

nikomatsakis:
Ah, for Abi, I need a github id

nikomatsakis:
I propose we merge this now and we can add Abi when I have the full information

jieyouxu (approved):

teams/funding-advisors.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[website]
+name = "Funding advisors team"
+description = "People coordinating closely with the Funding team"
+
+[[zulip-groups]]
+name = "T-funding-advisors"
+
+[[zulip-streams]]
+name = "funding/with-advisors"
```
jieyouxu, line 23: > [ERROR rust_team] failed: Id of stream 'funding/with-advisors' not found. The stream probably doesn't exist and sync-team doesn't support creating it yet. Please create the stream manually and add the rust-lang-owner user to it.

@nikomatsakis @Kobzol could u please manually invite Rust Lang Owner account to this private zulip stream?

jieyouxu: Screenshot 2026-07-23 215334

https://rust-lang.zulipchat.com/#user/494485

Kobzol: Done.

LoriLorusso:
Abi's github is https://github.com/abibroom

PR #2531, opened 2026-06-22
Add missing `rustbot` to the `log` repository

Urgau:
The `log` repo has had a [`triagebot.toml`](https://github.com/rust-lang/log/blob/master/triagebot.toml) for quite some time now, but never gave the permissions to rustbot/triagebot.

This is starting to cause some issues for triagebot as we have been trying to update some review comments for a few days. Let's fix the inconsistency by giving `rustbot` the permissions.

### DIFF
--- a/repos/rust-lang/log.toml
+++ b/repos/rust-lang/log.toml
@@ -2,7 +2,7 @@ org = 'rust-lang'
 name = 'log'
 description = 'Logging implementation for Rust'
 homepage = "https://docs.rs/log"
-bots = []
+bots = ["rustbot"]
 
 [access.teams]
 crate-maintainers = 'maintain'


Mark-Simulacrum (approved):
Hm, this seems fine, but I'd expect us to not try to run handlers if we don't have permissions to do so (or else we need this in every/all repositories, right?)

Urgau:
> I'd expect us to not try to run handlers if we don't have permissions to do so (or else we need this in every/all repositories, right?)

We current don't do any check permission check, we (maybe naïvely) trust that if a repo has a `triagebot.toml` that it has the right permissions.

We do have some [exceptions](https://github.com/rust-lang/triagebot/blob/98f85d92ba7547ce6a1634e5883a899cce7a5411/rust-lang.triagebot.toml#L23) for org-wide features, but it's manual and rely on the same principal (assume rights if `triagebot.toml` file present).

PR #2534, opened 2026-06-26
simpleinfra: Enable forking-renovate bot

Turbo87:
The repo has no renovate.json yet, so enabling the bot will open an onboarding PR with a suggested config. This surfaces outdated dependencies, such as those in the Fastly Compute@Edge code for static.crates.io.

r? @rust-lang/infra

### DIFF
--- a/repos/rust-lang/simpleinfra.toml
+++ b/repos/rust-lang/simpleinfra.toml
@@ -1,7 +1,7 @@
 org = "rust-lang"
 name = "simpleinfra"
 description = "Rust Infrastructure automation"
-bots = []
+bots = ["forking-renovate"]
 
 [access.teams]
 infra = "write"


marcoieni:
Personally I'm ok with having renovate, but as a beginning, we should just have manual triggered PRs via the dependency dashboard. We have countless of outdated dependencies and I'm not comfortable with dealing with them all.
Plus, I don't know what's the strategy of renovate updates (eg if in one PR it updates ALL aws providers to the latest, we can't merge that PR, those things need to be done in group).
And right now we don't have the time to figure out renovate groups and stuff like that.

Turbo87:
> Personally I'm ok with having renovate, but as a beginning, we should just have manual triggered PRs via the dependency dashboard

that is something we can adjust in the onboarding PR

> We have countless of outdated dependencies and I'm not comfortable with dealing with them all.

you don't have to deal with them all at once. but right now they are somewhat hidden, and having renovatebot active means that we at least are aware of the list of outdated dependencies. whether they are opened as PRs or not can be configured.

> Plus, I don't know what's the strategy of renovate updates (eg if in one PR it updates ALL aws providers to the latest, we can't merge that PR, those things need to be done in group).

yeah, I think by default it updates the dependencies across the whole project, but it might be possible to configure it differently. I expect this to be a bit of trial and error, but should be achievable.

> And right now we don't have the time to figure out renovate groups and stuff like that.

happy to help

marcoieni (approved):
ok, let's merge this and adjust the onboarding PR to manually trigger renovate as a start 👍

Issue #2518, opened 2026-06-15
Control github pages settings from the team repo


marcoieni:
Docs: https://docs.github.com/en/rest/pages/pages?apiVersion=2026-03-10

At the moment, github pages are setup manually.

It would be nice to configure them via the `repo.toml` files.

From the toml config file, you should be able to setup both methods (branch and github actions)

cuba0001:
Hello! This feature sounds really interesting to me. I'd like to work on it and use it as an opportunity to get a deeper understanding of the team repo internals. Could I be assigned to this issue?

marcoieni:
done! enjoy!

marcoieni:
are you a human or a bot? Since you are new in github.
Is this your first open source contribution ever? This issue is not so easy for a first contribution.

cuba0001:
I’m a human, and yes, I’m really new to GitHub. But I’m not new to programming. I’m a Senior Java Developer with 7 years of experience, mostly in fintech and banking. I’m currently looking for open-source projects where I can contribute and gain more hands-on experience with Rust.

At the same time, I’ve also picked up an issue in rust-lang/bors, where I’ve already made my first PR and it is currently under review. Here is also my Zulip profile: https://rust-lang.zulipchat.com/#user/1096210

I can prove that I’m not a bot over a video call, although I can’t promise perfect English :)

I agree that this issue is not the easiest one. But before offering to work on it, I spent a couple of evenings studying this repository. At a high level, my current understanding is this:

Someone edits the TOML files, opens a PR, and merges it. As a result, GitHub Actions starts the build and runs the sync process. The sync checks the diff and sends requests to GitHub, Zulip, Mailgun, and so on.

In this flow, I would need to add support for calling the GitHub Pages API, with two possible configuration options: via a branch or via GitHub Actions. If you still have doubts, before I start the implementation, I can write a more detailed feature design a bit later and share it with you first.

marcoieni:
Sounds great, thank you for explaining.

Sorry if we had doubts. 
We noticed you were new to github and we saw lots of bots recently so we were scared a bit. 

Looking forward to review your PR 😁

marcoieni:
closed by https://github.com/rust-lang/team/pull/2535 and https://github.com/rust-lang/team/pull/2550

PR #2535, opened 2026-06-26
Adds support for managing GitHub Pages settings from repo TOML data

cuba0001:
Related to #2518

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -361,6 +361,29 @@ Here are the available bots that can be added to the `bots` array:
 - `renovate`, `forking-renovate`: See the [Renovate Forge docs](https://forge.rust-lang.org/infra/docs/renovate.html).
 - TODO: Document other bots.
 
+### GitHub Pages
+
+GitHub Pages settings can optionally be managed from the repo TOML file. If the
+`[pages]` section is absent, sync-team tool will not modify the repository's existing
+GitHub Pages settings.
+
+```toml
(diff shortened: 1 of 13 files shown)


marcoieni (approved):

marcoieni:
Thanks for this!
The next step is implementing the deletion mechanism, right? I.e. if a `[pages]` block is not present, `team` deletes the GitHub Page.
One we have this mechanism, we can import all existing GitHub pages and we should see an empty diff in a test PR similar to https://github.com/rust-lang/team/pull/2539 (since I can push to the repo, the diff runs with the rust code from the branch).

Do you want to work on the deletion mechanism? I can take care of importing all existing github pages 👍

cuba0001:
Thanks!
Yes, sure, I'd like to work on the deletion mechanism. I wasn't completely sure that it was required at this stage because it's really too dangerous. Anyway, I'll continue working on it and take into account the cleanups you made.

As for importing all existing GitHub Pages settings, thanks for offering to help with that. From what I understand, fetching the current GitHub Pages settings for each repo requires authorized access to the repositories. So, I am not able to do that. But if I'm wrong, I have no problem taking care of that part as well.

marcoieni:
I also think it requires permission to fetch github pages permission.

Yeah, the deletion is dangerous but we have a dry run mechanism (the PR comment) that hopefully should catch errors before applying. Ok, look forward to reviewing your next PR 😁

Issue #2543, opened 2026-07-01
Verify that `github-pages` environment exists


marcoieni:
As ehuss noticed in https://github.com/rust-lang/team/pull/2541#issuecomment-4849685554:

> GitHub automatically generates a github-pages environment whenever you use the deploy action

We should add a [validation](https://github.com/rust-lang/team/blob/main/src/validate.rs) that checks that this environment is defined in case we have GitHub Pages configured with the workflow build type.

I.e. repositories with the following toml:

```toml
[pages]
build-type = "workflow"
```

should also have:

```toml
[environments.github-pages]
```

Related to https://github.com/rust-lang/team/issues/2518

cuba0001:
Ready for this.

marcoieni:
thanks, I assigned this issue to you 🚀

PR #2560, opened 2026-07-06
Require github-pages environment for workflow Pages

cuba0001:
Related to #2543

Should notice 2 things.

1. I used a LLM to double-check the existing repo TOML files for configs that already violate this requirement. It found `repos/rust-lang/funding.toml`, so I added `[environments.github-pages]` there in this PR. If that is premature, let me know I can remove it.
2. `static_api.rs` still does not take this new validation into account (as I understand #2543 requires `validate.rs` only). If that is expected, just omit it.

### DIFF
--- a/docs/toml-schema.md
+++ b/docs/toml-schema.md
@@ -384,7 +384,7 @@ build-type = "workflow"
 ```
 
 For `legacy`, `branch` is required and `path` must be either `/` or `/docs`.
-For `workflow`, `branch` and `path` must not be set, and the repository must
+For the `workflow` build type, `branch` and `path` must not be set, and the repository must
 define the `github-pages` environment (GitHub automatically creates this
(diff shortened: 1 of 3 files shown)


docs/toml-schema.md
```diff
@@ -379,10 +379,15 @@ path = "/docs"
 # Build Pages from a GitHub Actions workflow.
 [pages]
 build-type = "workflow"
+
+[environments.github-pages]
 ```
 
 For `legacy`, `branch` is required and `path` must be either `/` or `/docs`.
-For `workflow`, `branch` and `path` must not be set.
+For `workflow`, `branch` and `path` must not be set, and the repository must
```
```suggestion
For the `workflow` build type, `branch` and `path` must not be set, and the repository must
```

marcoieni (approved):

PR #2570, opened 2026-07-09
Set crabwatch custom property on rust-lang/crates-io-auth-action

Sandijigs:
Opts crates-io-auth-action into crabwatch scanning by setting the crabwatch custom property.


marcoieni (approved):
thanks! Yeah, let's use this to test crabwatch 😎

PR #2571, opened 2026-07-09
update to rust v1.97

marcoieni:


.github/actions/setup-rust/action.yml
```diff
@@ -13,7 +13,7 @@ runs:
     - name: Install Rust Stable
       shell: bash
       env:
-        RUST_VERSION: "1.96.1"
+        RUST_VERSION: "1.97.0"
```
jieyouxu: Suggestion: 1.97.1 now that we have a patch release

jieyouxu (changes requested):

jieyouxu (approved):
Thanks

Issue #2564, opened 2026-07-07
Allow custom properties to be string arrays


marcoieni:
https://github.com/rust-lang/team/pull/2512 solved https://github.com/rust-lang/team/issues/2504 by allowing setting custom properties of type string and bool. However, custom properties can also be of type "multi select".

<img width="1180" height="1298" alt="Image" src="https://github.com/user-attachments/assets/ff90b6f3-f108-4f63-8995-1fa9744d83ba" />

In this case, the org admin can select a list of allowed values (no need to code this into `team`).

So the team repo should support a custom property like the following:

```toml
[custom-properties]
colors = ["red", "green"]
```

cuba0001:
Ready for this.

marcoieni:
Thanks! I assigned this to you 😎

PR #2572, opened 2026-07-09
Allow custom properties to be string arrays

cuba0001:
Related to #2564


marcoieni (approved):
Thanks, this looks good. We don't have any array custom properties right now. Imo there's no need to test this now. We can test it when we need to create the first array custom property.

marcoieni:
the failure is not related to your code, I fixed it in https://github.com/rust-lang/team/pull/2573

PR #2573, opened 2026-07-10
fix: exclude parents rulesets

marcoieni:
I defined an organization ruleset for the repositories with the `crabwatch = true` custom properties.
Unfortunately, now CI fails because when retrieving repository rulesets, it doesn't recognize the "workflows" type.

This PR unblocks CI.

Here is the error I saw when applying https://github.com/rust-lang/team/pull/2572:

```
**Error:** rust_team] failed: Cannot deserialize type `rust_team::sync::github::api::Ruleset` from the following response body:
    {
      "_links": {
        "html": {
          "href": "[https://github.com/rust-lang/crabwatch/rules/18624409](https://github.com/rust-lang/crabwatch/rules/18624409)"
        },
        "self": {
          "href": "[https://api.github.com/repos/rust-lang/crabwatch/rulesets/18624409](https://api.github.com/repos/rust-lang/crabwatch/rulesets/18624409)"
        }
      },
      "bypass_actors": [],
      "conditions": {
        "ref_name": {
          "exclude": [],
          "include": [
            "~DEFAULT_BRANCH"
          ]
        }
      },
      "created_at": "2026-07-07T16:08:44.564+02:00",
      "current_user_can_bypass": "never",
      "enforcement": "active",
      "id": 18624409,
      "name": "crabwatch",
      "node_id": "RRS_lACkVXNlcs4AUt55zgEcL5k",
      "rules": [
        {
          "parameters": {
            "do_not_enforce_on_create": false,
            "workflows": [
              {
                "path": ".github/workflows/crabwatch.yml",
                "ref": "refs/heads/main",
                "repository_id": 1256932671
              }
            ]
          },
          "type": "workflows"
        }
      ],
      "source": "rust-lang",
      "source_type": "Organization",
      "target": "branch",
      "updated_at": "2026-07-09T16:53:19.260+02:00"
    }
**Error:** rust_team] caused by: unknown variant `workflows`, expected one of `creation`, `update`, `deletion`, `required_linear_history`, `merge_queue`, `required_deployments`, `required_signatures`, `pull_request`, `required_status_checks`, `non_fast_forward` at line 1 column 224
**Error:** Process completed with exit code 1.
```

## AI disclosure

I used GPT5.6 with the codex harness to generate this change. I reviewed its output and changed it where necessary.

PR #2574, opened 2026-07-10
Sync t-cargo/private with membership

ehuss:
This adds the #t-cargo/private Zulip stream so that it tracks the team membership. This stream already exists, and I have added Rust Owner Account.


jieyouxu (approved):
(CI might need https://github.com/rust-lang/team/pull/2573)

PR #2583, opened 2026-07-16
Add command for finding inactive users and moving users to alumni

Kobzol:
This PR adds two new commands. The first one helps finding users who haven't sent a message on the Rust Zulip for a long time and/or haven't sent a comment on the `rust-lang` GitHub organization for a long time.

The second command moves a given user to alumni in the teams that they are a member of (optionally filtered to a subset of the teams).

The motivation for this is https://github.com/rust-lang/leadership-council/blob/main/policies/membership/auto-alumni.md - we want to have some automation to help LC detect people who have not been active in the Project for some time, and then move them to alumni if they don't respond.

### DIFF
--- a/src/api/github.rs
+++ b/src/api/github.rs
@@ -332,7 +332,7 @@ query($query: String!, $issueLimit: Int!, $commentLimit: Int!) {
         }
 
         // Sort by creation date (most recent first) and take the limit
-        all_comments.sort_by(|a, b| b.created_at.cmp(&a.created_at));
+        all_comments.sort_by_key(|b| std::cmp::Reverse(b.created_at));
         all_comments.truncate(limit);
 
         Ok(all_comments)
--- a/src/api/github.rs
+++ b/src/api/github.rs
@@ -157,7 +157,7 @@ impl GitHubApi {
                 let text = response.text().await?;
                 return Err(anyhow::anyhow!("Request failed with {status}: {text}"));
             } else {
-                return Ok(response.json_annotated().await?);
+                return response.json_annotated().await;
             }
         }
     }
--- a/src/api/github.rs
+++ b/src/api/github.rs
@@ -2,10 +2,10 @@ use crate::sync::utils::ResponseExt;
 use anyhow::{Context, Error, bail};
 use base64::Engine;
 use base64::prelude::BASE64_STANDARD;
-use chrono::{DateTime, Utc};
-use reqwest::Method;
+use chrono::{DateTime, Duration, Utc};
 use reqwest::header::{self, HeaderValue};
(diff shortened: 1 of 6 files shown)


marcoieni:
ci is failing ❌

Mark-Simulacrum (dismissed):

Kobzol:
@jieyouxu Could you please create an `alumni` label? I'll modify the PR template to automatically apply it, so that we can track the alumni PRs in an easier way.

Kobzol:
@rustbot label +C-enhancement

(Just testing if triagebot works here)

Kobzol:
I added a bunch of functionality that I found useful while using this command to actually find inactive members and move them to alumni status. I have some extensions planned on top of it, but I think that at this point the PR is large enough.

src/api/github.rs
```diff
@@ -124,12 +125,41 @@ impl GitHubApi {
...
+                        .and_then(|v| v.to_str().ok())
+                        == Some("0")
+                {
+                    let reset_at = headers
+                        .get("x-ratelimit-reset")
+                        .and_then(|v| v.to_str().ok())
+                        .and_then(|t| t.parse::<u64>().ok())
+                        .and_then(|timestamp| chrono::DateTime::from_timestamp(timestamp as i64, 0))
+                        .map(|d| d + chrono::Duration::seconds(1))
+                        .unwrap_or(Utc::now() + chrono::Duration::minutes(1));
```
Mark-Simulacrum, line 148: I guess this sequence seems OK, but should we maybe warn if we're failing to parse? I guess it'll be obvious because we'll keep hitting the rate limit?

src/main.rs
```diff
@@ -566,6 +611,114 @@ async fn run() -> Result<(), Error> {
...
+                            .collect::<Vec<_>>()
+                            .join("\n"),
+                        date.format("%d.%m.%Y")
+                    ),
+                };
+                Command::new("git")
+                    .arg("add")
+                    .arg("teams")
+                    .spawn()?
+                    .wait()?;
```
Mark-Simulacrum: nit: .status()? is probably better? And maybe check that it's OK?

Mark-Simulacrum (dismissed):
Happy to merge as-is or with nits fixed

src/api/github.rs
Kobzol: The ratelimit that we in practice hit here resets every minute (it's the 30/minute search one, not the 5k/hour one), and this is right now only used locally when running the command, so I think it should be fine.

Mark-Simulacrum (approved):

Kobzol:
(I can't merge here :) )

marcoieni:
I didn't review this PR but I'm merging because it was approved 👍

PR #2586, opened 2026-07-16
Renamed allowed-merge-apps to bypass-apps

cuba0001:
Related to #2394 

1) Renamed `allowed-merge-apps` to `bypass-apps` in the TOML files and `schema.rs`.
2) To maintain backward compatibility and avoid deserialization errors in triagebot, I kept the old `allowed_merge_apps` field in the public `rust_team_data::v1` schema with its values copied from the new.
3) The deprecated field can be removed in a follow-up PR after updating and deploying triagebot.


marcoieni (approved):

marcoieni:
tested in https://github.com/rust-lang/team/pull/2616
the dry run is clean, thanks!

PR #2594, opened 2026-07-18
Configure trusted publishing for futures-rs

Kobzol:
Discussed in https://rust-lang.zulipchat.com/#narrow/channel/187312-wg-async/topic/.E2.9C.94.20futures-rs.200.2E3.2E33.3F/with/611444229.

### DIFF
--- a/repos/rust-lang/futures-rs.toml
+++ b/repos/rust-lang/futures-rs.toml
@@ -17,3 +17,23 @@ pattern = "main"
 required-approvals = 0
 
 [environments.github-pages]
+
+[environments.publish]
+branches = ["main"]
+
+[[crates-io]]
+crates = [
+    "futures",
+    "futures-channel",
+    "futures-core",
+    "futures-executor",
+    "futures-io",
+    "futures-macro",
+    "futures-sink",
+    "futures-task",
+    "futures-test",
+    "futures-util",
+]
+publish-workflow = "publish.yml"
+publish-environment = "publish"
+teams = ["wg-async"]
--- a/repos/rust-lang/futures-rs.toml
+++ b/repos/rust-lang/futures-rs.toml
@@ -19,7 +19,7 @@ required-approvals = 0
 [environments.github-pages]
 
 [environments.publish]
-branches = ["main"]
+branches = ["0.3"]
 
 [[crates-io]]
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (dismissed):

repos/rust-lang/futures-rs.toml
```diff
@@ -17,3 +17,23 @@ pattern = "main"
...
+    "futures-io",
+    "futures-macro",
+    "futures-sink",
+    "futures-task",
+    "futures-test",
+    "futures-util",
+]
+publish-workflow = "publish.yml"
+publish-environment = "publish"
+teams = ["wg-async"]
```
taiki-e, line 39: Is this an approver? Since cramertj stopped being involved (5-6 years ago?) I'm the only current member of wg-async who has been involved in maintaining futures-rs, it seems odd that all wg members would have permission over releases.

repos/rust-lang/futures-rs.toml
```diff
@@ -17,3 +17,23 @@ pattern = "main"
 required-approvals = 0
 
 [environments.github-pages]
+
+[environments.publish]
+branches = ["main"]
```
taiki-e:
```suggestion
branches = ["0.3"]
```

At this time, we don't create any release from main.

taiki-e (changes requested):
Thanks!

repos/rust-lang/futures-rs.toml
Kobzol: This specifies which team will own the crate on crates.io. But in practice with trusted publishing turned on, the only thing that this is useful for is yanking.

By the way, it would be great if the current active roster of wg-async was updated to match today's state!

Mark-Simulacrum: Technically these people have access to turn off trusted publishing and manually publish, but that would be a loud-ish action. Long-term I personally would like us to remove all human access to crates.io' (probably proxying yanks via some github/zulip bot) and/or have stronger yank-only enforcemeent.

marcoieni:
Maybe we should remove the `needs infra admin review` label because the corresponding PR is not ready to be reviewed. Maybe put this pr in draft?

Issue #2582, opened 2026-07-16
Notify the infrastructure team when the untracked repos workflow fails


marcoieni:
see https://github.com/rust-lang/team/pull/2580#issuecomment-4991093703

cuba0001:
As I can see there are two possible ways to implement this:

1. Add a workflow-level notification job. This would notify the infrastructure team about any workflow failure. But the notification would probably only contain a link to the failed run.

2. Send the notification directly from check_untracked_repos in src/ci.rs. This would allow the notification to include the repository names, but it would not cover failures occurring before or after that point.

Or even both :)

Also, where exactly notifications should be sent to?

Kobzol:
I would do it in the GHA workflow directly. For other tooling, we send messages to Zulip: https://github.com/rust-lang/josh-sync/blob/main/.github/workflows/rustc-pull.yml#L162.

marcoieni:
#2630 fixed this issue.

PR #2611, opened 2026-07-20
Notify the infrastructure team when the untracked repos workflow fails

cuba0001:
Related to #2582

This adds a `notify-infra` job to the untracked repositories workflow. The job runs when `check-untracked-repos` fails and sends a Zulip message containing a link to the failed workflow run. The notification is currently posted to the `t-infra` stream under the `Untracked repositories` topic. If a direct message would be preferable, then Zulip user_ids are required.

### DIFF
--- a/.github/workflows/check-untracked-repos.yml
+++ b/.github/workflows/check-untracked-repos.yml
@@ -27,3 +27,23 @@ jobs:
         run: |
           cargo build --release
           ./target/release/rust-team ci check-untracked-repos
+
+  notify-infra:
+    name: Notify the infrastructure team
+    needs: [ check-untracked-repos ]
+    if: ${{ !cancelled() && needs.check-untracked-repos.result == 'failure' }}
+    runs-on: ubuntu-latest
+    environment: deploy
+    steps:
+      - name: Send Zulip message
+        uses: zulip/github-actions-zulip/send-message@f675f2b4eb2a95fae974215476dcb7ad8dfeff6b # v2.0.2
+        with:
+          api-key: ${{ secrets.ZULIP_API_TOKEN }}
+          email: ${{ secrets.ZULIP_USERNAME }}
+          organization-url: "https://rust-lang.zulipchat.com"
+          to: "t-infra"
+          type: "stream"
+          topic: "Untracked repositories"
+          content: |
+            The untracked repositories workflow failed.
+            Check out the [workflow run](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}).


.github/workflows/check-untracked-repos.yml
```diff
@@ -27,3 +27,23 @@ jobs:
...
+    needs: [ check-untracked-repos ]
+    if: ${{ !cancelled() && needs.check-untracked-repos.result == 'failure' }}
+    runs-on: ubuntu-latest
+    environment: deploy
+    steps:
+      - name: Send Zulip message
+        uses: zulip/github-actions-zulip/send-message@f675f2b4eb2a95fae974215476dcb7ad8dfeff6b # v2.0.2
+        with:
+          api-key: ${{ secrets.ZULIP_API_TOKEN }}
+          email: ${{ secrets.ZULIP_USERNAME }}
```
Kobzol, line 42: Just noting that this doesn't have to be secret, and we can just hardcode it once the bot for it is created.

Kobzol (commented):
Thanks! Just noting that I would like to use a similar mechanism to also track optional CI job failures on `rust-lang/rust`, so this could be a good test for this approach.

.github/workflows/check-untracked-repos.yml
cuba0001: Don't have a permission for creating one.

image

cuba0001: Or maybe this one would be good?

image

Kobzol: Yeah we have to create a new bot for this. I don't have the permission either, one of the infra admins will have to do it.

cuba0001: Should I contact an infra admin directly, or start a topic in #t-infra?

Kobzol: @marcoieni can create a Zulip bot for this, once he gets to this PR :)

marcoieni:
Mark in today's meeting ([#t-infra > meeting 2026-07-20 @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/meeting.202026-07-20/near/611673046)) suggested a different approach to this:

> for team IMO the ideal would be that CI just cuts a PR importing it and humans can finish getting the PR into shape

I think this would be a better approach rather than notifying the infra team on Zulip. What do you both think?

Kobzol:
That's a good idea, yeah. `josh-sync` can again be used as inspiration, because it does the exact same thing, if it detects some changes, it will create a PR (unless a PR with the same title and author already expects, then it just does nothing).

cuba0001:
From what I understand, the desirable behavior for check-untracked-repos job is:

1. GHA starts a check-untracked-repos
2. it finds untracked repos
3. instead of bail! do:
3.1 create a config file (like repos/rust-lang/bors.toml) with pre-filled fields such as org, name, description etc.
3.2 create branch, commit, push, create PR (avoiding duplicates, I use josh-sync’s rustc-pull.yml as a reference)
3.3 notify infra-admin (not sure if it's still needed, but how else he knows?)

If I am right, then it sounds very reasonable and interesting.

marcoieni:
> 3\. 3.3 notify infra-admin (not sure if it's still needed, but how else he knows?)

I don't think this is needed. Some of us already watch the repository and github assign us as reviewers 👍

image

The rest of the plan looks good to me!

Kobzol:
I would suggest for the `check-untracked-repos` to get a new parameter (`--create-missing`), which will create the necessary TOML files for the missing repositories. And then it would still exit with an error, e.g. exit code 1.

And then do the branch/commit/push/PR creation thing in the YAML file in bash (if it finds that the command ended with exit code 1), similar to how josh-sync does it, as that would be a bit annoying to do in Rust.

PR #2615, opened 2026-07-21
Archive community-events and community-localization teams

Kobzol:
While browsing recently inactive people and teams, it seemed to me that these two teams have not been active for some time.

Let me know if I am wrong!

CC @m-ou-se @Manishearth

PR #2618, opened 2026-07-22
Automatic synchronization for the new `all/rustacean-connection` channel

steffahn:
This should keep the current list of members in sync with future changes to T-all membership. It also adds the foundation staff members to `all/rustacean-connection` (which AFAICT are currently not members yet) for full access parity with `all/private`.

*the people this will add, AFAIK:*
Bildschirmfoto_20260722_161539

cc @apiraino 

---

*Edit:* Actually the diff is slightly larger than this, adding 9 people (10 if you count the owner bot; and including 4 of the 5 people listed above, 1 is already in the new channel) and removing 3.

*Edit2:* Okay… **part** of the discrepancy is from #2615 happening in between, which explains 2 of the 3 removals. 
*Edit3:* Actually, the 3rd removal is also a recent alumni move, namely #2607.

### DIFF
--- a/teams/all.toml
+++ b/teams/all.toml
@@ -39,3 +39,11 @@ name = "all/private"
 extra-teams = [
     "foundation-staff",
 ]
+
+# Secondary channel for people sharing vacation plans for better separation from all/private.
+# [Same level of privacy as all/private though.]
+[[zulip-streams]]
+name = "all/rustacean-connection"
+extra-teams = [
+    "foundation-staff",
+]


apiraino:
ok, now I think I understand (it was quite confusing).

thanks

jieyouxu (approved):

jieyouxu:
~~Oh right, someone needs to invite rust-lang-owner~~
Forgot that I could

PR #2628, opened 2026-07-24
Exclude users from all/rustaceans-connections

apiraino:
this way we can prevent people from being auto-readded to a Zulip stream when they don't want

r? @Kobzol

(let's see if someone else wants to be added)

### DIFF
--- a/teams/all.toml
+++ b/teams/all.toml
@@ -47,3 +47,6 @@ name = "all/rustacean-connection"
 extra-teams = [
     "foundation-staff",
 ]
+excluded-people = [
+    "ChayimFriedman2",
+]


steffahn:
Some alternatives besides individually listing people here may include e. g. 

* telling people to just mute the channel (unless there are significant shortcomings then still, compared to properly unsubscribing)
* general solutions[^1] involving the "who can subscribe to this channel" permissions (users with this permission on a private channel can always see/preview that private channel's contents, and can subscribe or unsubscribe at will)

[^1]: "general" because I don't see any strong reason not to allow users this possibility for *any* private channel. I'd assume we'd still - somehow - want to auto-subscribe them *once* when joining (and auto-unsubscribe when leaving the team)

jieyouxu (approved):
Will merge this, easy to change anyway

PR #2630, opened 2026-07-25
Create configs when untracked repos found

cuba0001:
Related to #2582

According to the discussion in the previous https://github.com/rust-lang/team/pull/2611 it creates missing toml files. I added an exit code 2 to signal that untracked repos are found and configs are created, because this is the simplest way for the workflow to handle it. This still keeps code 1 for unexpected errors.

### DIFF
--- a/.github/workflows/check-untracked-repos.yml
+++ b/.github/workflows/check-untracked-repos.yml
@@ -38,42 +38,34 @@ jobs:
 
       - name: Check untracked repositories
         id: check
-        # exit code 2 means that configurations were created, so do not exit immediately
-        shell: bash {0}
         run: |
           cargo build --release
-          CARGO_BUILD_EXIT_CODE=$?
 
-          if [[ "$CARGO_BUILD_EXIT_CODE" -ne 0 ]]; then
-            exit "$CARGO_BUILD_EXIT_CODE"
-          fi
+          CHECK_UNTRACKED_REPOS_EXIT_CODE=0
+          # Capture the status without failing because
+          # exit code 2 means that configurations were created.
+          ./target/release/rust-team ci check-untracked-repos --create-missing \
+            || CHECK_UNTRACKED_REPOS_EXIT_CODE=$?
 
-          ./target/release/rust-team ci check-untracked-repos --create-missing
-          CHECK_UNTRACKED_REPOS_EXIT_CODE=$?
+          case "$CHECK_UNTRACKED_REPOS_EXIT_CODE" in
+            0) configurations_created=false ;;
+            2) configurations_created=true ;;
+            *) exit "$CHECK_UNTRACKED_REPOS_EXIT_CODE" ;;
+          esac
 
-          if [[ "$CHECK_UNTRACKED_REPOS_EXIT_CODE" -eq 0 ]]; then
-            echo "configurations_created=false" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-
-          if [[ "$CHECK_UNTRACKED_REPOS_EXIT_CODE" -eq 2 ]]; then
-            echo "configurations_created=true" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-
-          exit "$CHECK_UNTRACKED_REPOS_EXIT_CODE"
+          echo "configurations_created=$configurations_created" >> "$GITHUB_OUTPUT"
 
       - name: Check for an existing pull request
         id: pull_request
         env:
-          GITHUB_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ github.token }}
         run: |
           PULL_REQUEST_EXISTS=$(
             gh pr list \
               --head "$AUTOMATION_BRANCH" \
               --state open \
-              --json number \
-              --jq 'length > 0'
+              --json isCrossRepository \
+              --jq 'map(select(.isCrossRepository == false)) | length > 0'
           )
 
           echo "Pull request exists: $PULL_REQUEST_EXISTS"
@@ -82,7 +74,7 @@ jobs:
       - name: Close previous pull request
         if: ${{ steps.check.outputs.configurations_created == 'false' && steps.pull_request.outputs.exists == 'true' }}
         env:
-          GITHUB_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ github.token }}
         run: gh pr close "$AUTOMATION_BRANCH"
 
       - name: Commit and push generated configurations
@@ -114,7 +106,7 @@ jobs:
       - name: Create pull request
         if: ${{ steps.check.outputs.configurations_created == 'true' && steps.pull_request.outputs.exists == 'false' }}
         env:
-          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
           PULL_REQUEST_BASE_BRANCH: ${{ github.event.repository.default_branch }}
           PULL_REQUEST_BODY: >-
             This pull request was created automatically because untracked repositories were found.
--- a/.github/workflows/check-untracked-repos.yml
+++ b/.github/workflows/check-untracked-repos.yml
@@ -1,3 +1,6 @@
+# This workflow checks the default branch every six hours for repositories that do not have a corresponding repos/*/*.toml file.
+# If missing configurations are generated, it force-pushes them to a fixed automation branch and creates a pull request if needed.
+# If no configurations are needed, it closes any open pull request from the automation branch.
 name: Check Untracked Repositories
 
 on:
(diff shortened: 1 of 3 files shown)


.github/workflows/check-untracked-repos.yml
```diff
@@ -6,24 +16,134 @@ on:
...
-      - name: Check for untracked repositories
-        shell: bash
+      # a previous run may already have opened an pull request
+      # if so then just add a new commit (because new untracked repos can also be found with this run too)
+      - name: Prepare automation branch
+        id: prepared_branch
+        env:
+          GITHUB_TOKEN: ${{ github.token }} # this one is generated by GHA
+        run: |
+          EXISTING_PULL_REQUEST_BRANCH=$(
```
Kobzol: The branch name is a constant, same as in josh-sync. I don't think we need this lookup complexity.

1. Run the `team` command and check if any TOML file was generated, according to its exit code
2. If yes, create the new branch, add the TOML files, commit it, and push to the branch
3. If a PR does not exist at the moment, create it from that branch

Kobzol (commented):
Hi, thanks for the PR! I quickly glanced at the YAML file only, left one comment.

.github/workflows/check-untracked-repos.yml
cuba0001: Just to clarify. Should the team command run from the automation branch or from main? This was the main point I was unsure about.

If we run it from main, there is a problem. If an infra-admin makes changes between the first and second workflow runs, those changes will be overwritten because we would have to force-push the branch, as each run would detect the same untracked repositories again (not the same for automation branch, because missing configs are already commited).

Is this ok?

As for dynamic branch name, I added it to avoid making the same branch actual in every workflow run, since it could fall far behind main if nobody rebases it or something.

Kobzol: I'd say that it should always run from main. If there is a diff, then force push to the branch. If there is no diff, and a PR is open, then close the PR (unless that somehow prevents us from opening a PR from the same branch in the future).

cuba0001: Done. It force pushes the automation branch if PR exists, its name is constant. Also, deletes PRs which are open and there is no diff.

.github/workflows/check-untracked-repos.yml
```diff
@@ -6,24 +9,118 @@ on:
...
+        id: pull_request
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          PULL_REQUEST_EXISTS=$(
+            gh pr list \
+              --head "$AUTOMATION_BRANCH" \
+              --state open \
+              --json number \
+              --jq 'length > 0'
```
marcoieni:
```suggestion
              --json isCrossRepository \
              --jq 'map(select(.isCrossRepository == false)) | length > 0'
```
limit the search to branches of this repository.

.github/workflows/check-untracked-repos.yml
```diff
@@ -6,24 +9,118 @@ on:
...
           cargo build --release
-          ./target/release/rust-team ci check-untracked-repos
+          CARGO_BUILD_EXIT_CODE=$?
+
+          if [[ "$CARGO_BUILD_EXIT_CODE" -ne 0 ]]; then
+            exit "$CARGO_BUILD_EXIT_CODE"
+          fi
+
+          ./target/release/rust-team ci check-untracked-repos --create-missing
+          CHECK_UNTRACKED_REPOS_EXIT_CODE=$?
```
marcoieni:
```suggestion
        run: |
          cargo build --release

          CHECK_UNTRACKED_REPOS_EXIT_CODE=0
          # Capture the status without failing because
          # exit code 2 means that configurations were created.
          ./target/release/rust-team ci check-untracked-repos --create-missing \
            || CHECK_UNTRACKED_REPOS_EXIT_CODE=$?
```

avoid `shell: bash {0}`

.github/workflows/check-untracked-repos.yml
```diff
@@ -6,24 +9,118 @@ on:
...
+            echo "configurations_created=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          if [[ "$CHECK_UNTRACKED_REPOS_EXIT_CODE" -eq 2 ]]; then
+            echo "configurations_created=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          exit "$CHECK_UNTRACKED_REPOS_EXIT_CODE"
```
marcoieni:
```suggestion
          case "$CHECK_UNTRACKED_REPOS_EXIT_CODE" in
            0) configurations_created=false ;;
            2) configurations_created=true ;;
            *) exit "$CHECK_UNTRACKED_REPOS_EXIT_CODE" ;;
          esac

          echo "configurations_created=$configurations_created" >> "$GITHUB_OUTPUT"
```
simplify the code

marcoieni (commented):
some initial comments. The first two detected with AI. All fixed with AI.

cuba0001:
> some initial comments. The first two detected with AI. All fixed with AI.

Made the commit including these improvements.

.github/workflows/check-untracked-repos.yml
```diff
@@ -6,24 +9,110 @@ on:
...
+            2) configurations_created=true ;;
+            *) exit "$CHECK_UNTRACKED_REPOS_EXIT_CODE" ;;
+          esac
+
+          echo "configurations_created=$configurations_created" >> "$GITHUB_OUTPUT"
+
+      - name: Check for an existing pull request
+        id: pull_request
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
```
marcoieni:
```suggestion
          GH_TOKEN: ${{ github.token }}
```
according to https://cli.github.com/manual/gh_auth_login this should be the correct way, right?

.github/workflows/check-untracked-repos.yml
```diff
@@ -6,24 +9,110 @@ on:
...
+              --jq 'map(select(.isCrossRepository == false)) | length > 0'
+          )
+
+          echo "Pull request exists: $PULL_REQUEST_EXISTS"
+          echo "exists=$PULL_REQUEST_EXISTS" >> "$GITHUB_OUTPUT"
+
+      - name: Close previous pull request
+        if: ${{ steps.check.outputs.configurations_created == 'false' && steps.pull_request.outputs.exists == 'true' }}
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
```
marcoieni:
```suggestion
          GH_TOKEN: ${{ github.token }}
```

.github/workflows/check-untracked-repos.yml
```diff
@@ -6,24 +9,110 @@ on:
...
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        id: app-token
+        with:
+          app-id: ${{ secrets.SYNC_TEAM_GH_APP_ID }}
+          private-key: ${{ secrets.SYNC_TEAM_GH_APP_PRIVATE_KEY }}
+
+      - name: Create pull request
+        if: ${{ steps.check.outputs.configurations_created == 'true' && steps.pull_request.outputs.exists == 'false' }}
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
```
marcoieni:
```suggestion
          GH_TOKEN: ${{ steps.app-token.outputs.token }}
```

marcoieni (commented):
fix token name

src/main.rs
```diff
@@ -186,7 +141,11 @@ enum CiOpts {
     /// Check if the .github/CODEOWNERS file is up-to-date
     CheckCodeowners,
     /// Check for untracked repositories in GitHub organizations
-    CheckUntrackedRepos,
+    CheckUntrackedRepos {
+        /// Create a TOML configuration file for each untracked repository and exit with code 2
```
marcoieni:
```suggestion
        /// Create a TOML configuration file for each untracked repository.
        /// If at least one repository was missing and the file was created, exit with code 2.
```

src/ci.rs
```diff
@@ -228,14 +244,22 @@ pub async fn check_untracked_repos(data: &Data) -> anyhow::Result<()> {
...
 
     warn!("❌ Found {} untracked repositories:", untracked.len());
     for repo in &untracked {
         warn!("  - {}/{}", repo.org, repo.name);
     }
 
+    if create_missing {
+        let created = create_missing_repo_configs(data_dir, &untracked)?;
+        for path in created {
+            info!("Created {}", path.display());
```
marcoieni: can't we emit the info in `create_missing_repo_configs` so that we can simplify the code here?
In theory there's no reason why this function should return the list of paths, right?
In tests we can check if the file we expect exists for example.

src/ci.rs
```diff
@@ -300,6 +324,189 @@ fn find_untracked_repos(
...
+        let contents = toml::to_string_pretty(&GeneratedRepoConfig {
+            org: &repo.org,
+            name: &repo.name,
+            description: &repo.description,
+            homepage: repo.homepage.as_deref(),
+            bots: Vec::new(),
+            access: GeneratedRepoAccess {
+                teams: BTreeMap::new(),
+            },
+        })
```
marcoieni: we can extract this into a From implementation to make this function shorter

src/ci.rs
```diff
@@ -300,6 +324,189 @@ fn find_untracked_repos(
...
+        .with_context(|| {
+            format!(
+                "failed to serialize configuration for {}/{}",
+                repo.org, repo.name
+            )
+        })?;
+
+        let repo_dir = data_dir.join("repos").join(&repo.org);
+        std::fs::create_dir_all(&repo_dir)
+            .with_context(|| format!("failed to create directory {}", repo_dir.display()))?;
```
marcoieni:
```suggestion
```
I think we can assume that the repository for the organization exists. If it doesn't, it's better to fail and add the directory manually, I think. So that we don't add a new directory for organizations by mistake with an automation.

src/ci.rs
```diff
@@ -300,6 +324,189 @@ fn find_untracked_repos(
...
+        let missing_repo_config_toml = repo_dir.join(format!("{}.toml", repo.name));
+
+        let mut file = OpenOptions::new()
+            .write(true)
+            .create_new(true)
+            .open(&missing_repo_config_toml)
+            .with_context(|| format!("failed to create {}", missing_repo_config_toml.display()))?;
+
+        file.write_all(contents.as_bytes())
+            .with_context(|| format!("failed to write {}", missing_repo_config_toml.display()))?;
```
marcoieni:
```suggestion
        std::fs::File::create_new(&missing_repo_config_toml)
            .with_context(|| format!("failed to create {missing_repo_config_toml:?}"))?
            .write_all(contents.as_bytes())
            .with_context(|| format!("failed to write {missing_repo_config_toml:?}"))?;
```
this is simpler.

cuba0001:
@marcoieni 
Sorry for the dealy. I force-pushed improvements.

marcoieni (approved):

PR #2631, opened 2026-07-26
Sync t-rustfmt/private with membership

fee1-dead:
The stream already exists, and Rust Owner Account has been added.

cc @jieyouxu


jieyouxu (approved):
Thanks, the private stream creation was discussed in DMs

cc @ytmimi @calebcartwright (for FYI)

PR #2468, opened 2026-05-14
create the funding team

nikomatsakis:
Initial charter from [rust-lang/leadership-council#294]:

* Get a first MiR contracted by June
* Establish procedures for ppl applying for becoming MiRs
* Establish procedures for finding sponsors
* Design a full charter by 31st of August and get it FCPed by the council
* Work within the merged RFMF RFC: https://github.com/rust-lang/rfcs/pull/3931
* wrt affiliation limits: only one employee per company is allowed on the funding team. If there are multiple (e.g. due to employment changes) either all but one must decide to leave or the one who has been there the shortest time will be removed automatically.

[rust-lang/leadership-council#294]: https://github.com/rust-lang/leadership-council/issues/294

cc @oli-obk

PR #2641, opened 2026-07-29
Archive hiring marker team and add their email to the funding team

oli-obk:
A left over cleanup forgotten in https://github.com/rust-lang/team/pull/2468

cc @rust-lang/funding

### DIFF
--- a/teams/archive/hiring.toml
+++ b/teams/archive/hiring.toml
@@ -3,7 +3,8 @@ kind = "marker-team"
 
 [people]
 leads = []
-members = [
+members = []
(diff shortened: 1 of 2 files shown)


Kobzol (commented):
I think that archiving is better than outright deleting it? Otherwise LGTM.

oli-obk:
> I think that archiving is better than outright deleting it? Otherwise LGTM.

I deleted it because it was just a marker team solely for that email address.

jieyouxu (approved):

PR #2650, opened 2026-08-04
Add spastorino to goal-owners

nxsaken:


nxsaken:
🤦 already done in #2646

PR #2651, opened 2026-08-04
Implement libs team refactor (RFC 3984)

clarfonthey:
Note: blocked on [RFC 3984](https://github.com/rust-lang/rfcs/pull/3984) merging, still has a few unanswered questions like renaming Zulip streams. Tried to split into a few commits to make reviewing easier.

Resolved items:
* [x] I'm not sure what the policy is for archived teams, especially when it comes to merging them into existing teams. I figured that since we have an explicit policy in the RFC for allowing alumni to be re-nominated, it would make sense to merge all the alumni here as well, but would also like some guidance on that. (this is fine)
* [x] I also am not sure whether we truly *need* to keep the library contributors team archived at all since it functionally has no difference from the new libs team. The libs-API team specifically feels like it's worth tracking, though. (this is also fine)
* [x] The [rust-lang/libs](https://crates.io/teams/github:rust-lang:libs) and [rust-lang-nursery/libs](https://crates.io/teams/github:rust-lang-nursery:libs) teams should not own any crates on crates.io. This requires contacting a few people first, since rust-lang-owner is not co-owning all of the crates.

Non-blocking items:
* [ ] Amanieu recommended just archiving t-libs/private.
* [ ] We will want to remain t-libs/reviewers to t-libs/private post-archive, but that can happen in a separate PR.


nia-e:
I think the Zulip stream setup proposed here makes sense, and I wouldn't change it. Same with merging the alumni, it checks out. The only bit I'm not 100% sure on archiving vs deleting libs-contributors & libs-api since functionally they're all just being merged, but I think this is the right call for historical preservation reasons.

lgtm, and I'm happy to merge this as-is as soon as the refactor FCP is done

clarfonthey:
Probably still gonna need a sign off from @Amanieu too as a team lead but otherwise, yeah, I figure Zulip rearranging can happen after the fact.

teams/libs-fcp.toml
```diff
@@ -0,0 +1,33 @@
...
+name = "Library FCP team"
+description = "Library team members with responsibility for signing off on API changes"
+zulip-stream = "t-libs"
+repo = "https://github.com/rust-lang/libs-team"
+
+[[zulip-groups]]
+name = "T-libs-fcp"
+
+[[zulip-streams]]
+name = "t-libs/private"
```
Amanieu, line 33: We should just drop the private channel. It's basically dead and all the internal discussions are happening on the t-libs/reviewers channel. I see no need to keep a separate channel for libs-fcp.

Amanieu (approved):

teams/libs-fcp.toml
clarfonthey: I'm fine with that, although my inclination is to assign libs-fcp to it first so it doesn't dump the new libs team in there immediately, then delete it in a future PR. Not that it matters, since it's not going to reveal old messages or anything, but to avoid confusion.

Since I believe deleting it from this PR will just cause its permissions to remain the same, which would dump all of the new libs into it.

nia-e: i believe perms are per-user, not per-group. either way this is smth infra can clarify, it seems we agree on the intent being to archive that channel without adding anyone new

clarfonthey: AFAIK the Zulip channel has group permissions, so the users would be added because their group changed, not because the channel changed. But either way, yeah, I'm happy to remove the channel from the PR if infra is fine coordinating that.

We would also want to name the reviewers channel to t-libs/private too, presumably.

repos/rust-lang/rust.toml
jieyouxu, line 1: NB: this touches `rust-lang/rust`

teams/libs-fcp.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[rfcbot]
+label = "T-libs"
+name = "Library"
+ping = "rust-lang/libs"
+
+[website]
+page = "library"
+name = "Library FCP team"
+description = "Library team members with responsibility for signing off on API changes"
```
jieyouxu, line 25: NB: I'm not sure if this is still accurate for what libs want to describe the FCP subteam as, but also this is easy to change so yeah.

teams/libs-fcp.toml
jieyouxu: Permissions are tied to per-user at least from I can tell.

jieyouxu (dismissed):
team-repo-admin approval given, this needs an infra-admin review as well since this modifies rust-lang/rust permissions.

teams/libs-fcp.toml
nia-e: FCPs are almost always for stably-visible changes, so if nothing else it's Close Enough

teams/libs.toml
```diff
@@ -6,14 +6,56 @@ leads = [
...
+    "Noratrieb",
+    "programmerjake",
+    "sayantn",
+    "SimonSapin",
+    "sunfishcode",
+    "tgross35",
     "the8472",
     "thomcc",
+    "workingjubilee",
+    "yaahc",
```
Mark-Simulacrum, line 43: This is adding everyone in this list with publish etc. access to the 20 crates currently owned (in part) by `libs` on crates.io - https://crates.io/teams/github:rust-lang:libs

I suspect we don't *really* want that. It's not an "umbrella privilege" laid out in the RFC. Some subset of those probably have restricted publishing to trusted publishing, which moves this to a question of repository access.

Should we maybe either remove that access on crates.io or move it to (say) crate-maintainers or libs-fcp? A few of those are controlled by team so we could do it here.

Mark-Simulacrum (commented):
It's unfortunately hard to tell what is actually changing permissions wise (i.e., where the libs team growing has influence). I think the crates.io thing is the primary one I see real risk in, but I may be missing other cases we granted libs access that may want to get removed as part of this transition.

teams/libs.toml
nia-e: libs-fcp is imo an acceptable option. we can always add perms to libs or crate-maintainers later, and libs-fcp should not consist of anyone we don't trust with that kind of access. we can always give libs write but not maintain perms for those crates

clarfonthey: cc @Amanieu in case you have a better idea short term, otherwise we can just move crate publishing to libs-fcp for now

tgross35: I brought this up at [#t-infra > crates.io owners of rust-lang crates](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/crates.2Eio.20owners.20of.20rust-lang.20crates/with/612234382) as well. But libc-fcp seems fine to me as well for now.

I think the solution mentioned there of having some bot handle yanks (release-plz?) and eventually removing libs-fcp would be the best long term option.

clarfonthey: I remember this discussion and wasn't 100% sure what the best long-term solution is, since write access to repos with CI publishing would mean that everyone effectively has crate-publishing privileges, but I also have no idea what the threat model here is anyway.

Amanieu: I just went through every single crate owned by rust-lang/libs on crates.io. I think we can safely just remove rust-lang/libs from all of them (and add rust-lang-owner to the ones that are missing it).

These all fall into one of these categories:
- Crates where release-plz is used for publishing, which uses rust-lang-owner and is controlled by write access in the team repo.
- Crates that are mostly maintained by a select number of people who are additional owners (e.g. dlmalloc-rs which is maintained by alexcrichton)
- Crates that are deprecated/unmaintained and haven't had a release in years.

tgross35: As long as infra can yank crates as needed until we get a better solution for that (I assume they can), this sounds reasonable to me.

Mark-Simulacrum: infra-admins can, it's just a bit of a hassle since we need to get into the owner account. I guess we can mint a token with yank permissions and stash it somewhere easier/less risky to get into.

clarfonthey: So, just to clarify what exactly would be necessary to make these changes in the repo: we just need to remove `teams = ["libs"]` from the `crates-io` section of repos wherever it's present, right? (I did this, but what Amanieu said implies there is more than just a handful of them.)

clarfonthey: Side note, there are a few crates that have this listed for compiler as well, and since that team is so large, I'm actually unsure if this was noticed when they did their team refactor either.

teams/libs.toml
```diff
@@ -88,4 +88,4 @@ address = "library-team@rust-lang.org"
 name = "T-libs"
 
 [[zulip-streams]]
-name = "t-libs/reviewers"
```
nia-e: why not leave this as reviewers? that way `private` can be archived entirely and won't need a rename

clarfonthey: We could do that, I just figured it shouldn't be too difficult to coordinate.

clarfonthey:
Just adding as a final remark: we seem to have fixed all outstanding issues except crates owned by the libs team. We're triaging that in this Zulip thread: [#t-infra > crates.io crate ownership @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/crates.2Eio.20crate.20ownership/near/616636135)

Note that as long as these two don't have any owned crates, we can merge:

* [x] https://crates.io/teams/github:rust-lang:libs
* [x] https://crates.io/teams/github:rust-lang-nursery:libs

clarfonthey:
Ready to merge and rebased. Hooray!

teams/libs.toml
```diff
@@ -6,14 +6,58 @@ leads = [
...
+    "dtolnay",
+    "folkertdev",
+    "hanna-kruppe",
+    "ibraheemdev",
+    "jhpratt",
+    "joboet",
+    "JohnTitor",
+    "joshtriplett",
+    "kennytm",
+    "KodrAus",
```
Kobzol, line 30: KodrAus is both an alumni and a member? I'll add a validation for that.

Mark-Simulacrum (dismissed):
Looks good to me modulo the comment from @Kobzol (https://github.com/rust-lang/team/pull/2651/changes#r3825133377)

Mark-Simulacrum (approved):

PR #2655, opened 2026-08-05
Generate GitHub App token on repository PR force push

marcoieni:
This allows CI to run when a "new repository PR" already exists, and the automation updates it.


ubiratansoares (approved):

Kobzol:
I thought that the important thing is which account/token opens the PR, not who pushes to it. If there's a PR open already, and the non-GH-app token force pushes to it, it also doesn't run PR CI?

marcoieni:
> If there's a PR open already, and the non-GH-app token force pushes to it, it also doesn't run PR CI?

yes, the rule is that actions from the default github actions token don't trigger workflows

PR #2660, opened 2026-08-06
Remove myself as Discourse mod

notriddle:
I haven't been putting the time in that I should, and mdBook is launching a new team that I'm probably going to be involved with.


notriddle:
CC @steffahn @oli-obk

oli-obk (approved):
Thanks for the time you have spent on it all these years!

steffahn:
Huge thank you from me as well!

jieyouxu (approved):

PR #2662, opened 2026-08-07
Configure branches for unrolled perf builds for bors

Kobzol:
Adds new branches that will be used by bors to unroll branches for perf. builds (this functionality will be ported over from rustc-perf).

Historically we had some issues with bors creating these branches, so it would be good if they could be pre-created manually first.

After the functionality is ported and we test that it works fine, I will remove the push access of rust-timer and the old `try-perf` and `try-tmp` branches.

### DIFF
--- a/repos/rust-lang/rust.toml
+++ b/repos/rust-lang/rust.toml
@@ -151,5 +151,19 @@ pattern = "perf-tmp"
 allowed-merge-teams = ["rust-timer"]
 prevent-update = true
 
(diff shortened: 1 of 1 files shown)


ubiratansoares (approved):

Kobzol:
Could you also create the branches on https://github.com/rust-lang/rust/branches/active? Thanks! Pushing latest `main` to them should be enough.

ubiratansoares:
Hopefully everything is set now :)

Issue #2654, opened 2026-08-05
Private repositories aren't discovered automatically


marcoieni:
With #2630 we periodically scan new repositories and create a PR when a repo is missing.

However, the scan step does not set GITHUB_TOKEN, so repository enumeration is anonymous and only sees public repositories.

Maybe we can use the read only github app that we use for the dry run to discover also private repositories?

cuba0001:
Looks like there is the .github/actions/generate-tokens/action.yml for that. We should re-use it and somehow pass all of those tokens to check_untracked_repos, right? The step of dry-run sets envs following the GITHUB_TOKEN_<ORG> naming pattern. Should it be the same in Check untracked repositories?

marcoieni:
yes, I think it should work 👍
I saw that we don't track all the org repos, so probably we should generate tokens just for the orgs that we use for the repositories.

cuba0001:
Then I am up to this.

> I saw that we don't track all the org repos, so probably we should generate tokens just for the orgs that we use for the repositories.

I will keep that in my mind.

PR #2666, opened 2026-08-09
Added org github tokens to check untracked private repositories

cuba0001:
Related #2654

### DIFF
--- a/src/ci.rs
+++ b/src/ci.rs
@@ -202,7 +202,7 @@ pub async fn check_untracked_repos(
     data_dir: &Path,
     create_missing: bool,
 ) -> anyhow::Result<CheckUntrackedReposResult> {
-    let github = crate::api::github::GitHubApi::new();
+    let github = crate::api::github::GitHubApi::new_with_org_tokens();
 
     // Get allowed GitHub organizations from config instead of hardcoding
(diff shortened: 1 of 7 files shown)


src/ci.rs
```diff
@@ -295,6 +306,13 @@ async fn fetch_all_github_repos(
     Ok(all_repos)
 }
 
+fn github_token_env_var_name(org: &str) -> String {
+    format!(
+        "GITHUB_TOKEN_{}",
+        org.to_ascii_uppercase().replace('-', "_")
+    )
+}
```
marcoieni: We already parse and select `GITHUB_TOKEN_{ORG}` credentials in 

https://github.com/rust-lang/team/blob/0a7e85b49a19d79d64293b19e8031a834b2b7f23/src/sync/github/api/tokens.rs#L194

 Could we reuse that mechanism here?

cuba0001: It was a bit tricky because of two github clients. I needed to expose `tokens.rs` to reuse it in `api/github`.

marcoieni:
Sorry I didn't have time to look into this. I'm unassigning this from me because I'm going to rustconf and after that I'll be on PTO for two weeks.
Hopefully someone else has time to look into it!

PR #2673, opened 2026-08-13
fix rustwide release publish

syphar:
From what I understand, this is what's needed to fix the failure? 

- https://github.com/rust-lang/rustwide/actions/runs/31682741361 
- https://github.com/rust-lang/rustwide/blob/a3eecbc214a907f04ab4e0bd36708d323f660c8d/.github/workflows/release.yml

### DIFF
--- a/repos/rust-lang/rustwide.toml
+++ b/repos/rust-lang/rustwide.toml
@@ -8,7 +8,8 @@ infra = "write"
 docs-rs = "write"
 
 [environments.publish]
-branches = ["main"]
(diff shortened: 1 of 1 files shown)


repos/rust-lang/rustwide.toml
```diff
@@ -9,6 +9,8 @@ docs-rs = "write"
 
 [environments.publish]
 branches = ["main"]
```
Kobzol: If releases only happen from tags, then this can be removed.

marcoieni (approved):

PR #2680, opened 2026-08-15
Remove `rust-timer` branches from `rust-lang/rust`

Kobzol:
The unrolling functionality has been migrated over to bors, so this access is not needed after https://github.com/rust-lang/rustc-perf/pull/2524. This removes one ad-hoc account with push privileges to `rust-lang/rust` :tada:.

After this is merged, the `perf-tmp` and `try-perf` branches should be removed from `rust-lang/rust`.

### DIFF
--- a/repos/rust-lang/rust.toml
+++ b/repos/rust-lang/rust.toml
@@ -28,8 +28,6 @@ rust-analyzer = "write"
 style = "write"
 types = "write"
 triage = "write"
-# needed for the branch protection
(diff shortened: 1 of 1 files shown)


ubiratansoares (approved):

ubiratansoares:
Those branches were deleted as requested!

screenshot-2026-08-17 at 21 17 02

Kobzol:
Thank you very much!

PR #2683, opened 2026-08-16
Partial libs team refactor: merge libs and libs-api, don't move libs-contributors into libs yet

clarfonthey:
Was a request from @Mark-Simulacrum.

Essentially, this is #2651 without the merging of libs-contributors into libs, which is blocked by removing team access to crates for libs. It *does* technically give access to two additional people for these crates (BurntSushi and dtolnay) as they were only part of libs-api before, but this is fine comparing to giving 21 additional people access (assuming I counted right).


Mark-Simulacrum:
Hm, I was expecting just the crates.io permission removals -- I think that's the only piece that needs doing before we can just merge the other PR? I guess we can merge this though since it's a more minimal expansion in practice...

clarfonthey:
Sure: #2684

clarfonthey:
Closing since we've unblocked the main PR.

PR #2701, opened 2026-08-21
remove duplicate code in validation

marcoieni:
I extracted a variable.

## AI disclosure

I used GPT5.6-Sol with the codex harness to generate this change. I reviewed its output and changed it where necessary.

### DIFF
--- a/src/validate.rs
+++ b/src/validate.rs
@@ -322,12 +322,14 @@ fn validate_duplicate_team_entries(data: &Data, errors: &mut Vec<String>) {
             errors.push(e.to_string());
         }
 
+        let alumni = || {
+            team.explicit_alumni()
+                .iter()
+                .map(|alum| alum.github.as_str())
+        };
+
         // Check alumni for duplicates
-        if let Err(e) = check_duplicates(
-            team.name(),
-            "alumni",
-            team.explicit_alumni().iter().map(|a| a.github.as_str()),
-        ) {
+        if let Err(e) = check_duplicates(team.name(), "alumni", alumni()) {
             errors.push(e.to_string());
         }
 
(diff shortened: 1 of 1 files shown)


src/validate.rs
```diff
@@ -322,22 +322,22 @@ fn validate_duplicate_team_entries(data: &Data, errors: &mut Vec<String>) {
             errors.push(e.to_string());
         }
 
+        let alumni = team
```
Kobzol: Cloning iterators is a bit icky, I usually create a closure instead (`let alumni = || ...`) and then call it repeatedly.

steffahn:
Alternatively, to reduce some of the duplication already, the `"leads + alumni"` check isn't really necessary in the first place, given that we also check `leads` is a subset of `members`.

steffahn:
Also actually, with the `"members + alumni"` check in place, the individual checks on `members` and on `alumni` become redundant, because those also create duplicates in `members + alumni` combined

steffahn:
so you just need 2 checks total; one on `members + alumni` and one on `leads`. And no duplicate code is left at all

Kobzol:
Yeah, but this situation is so rare that I didn't want to bother with it. And if we only check members + alumni together, then the error message will be misleading (it would have to be generalized to something like "there is duplication either amongst members or members + alumni").

steffahn:
This is what I thought it can look like: #2702

marcoieni:
Thanks both!
I implemented Kobzol's suggestion. I'll let the reviewer decide which approach they prefer. If they prefer https://github.com/rust-lang/team/pull/2702, feel free to close this PR. 👍

ubiratansoares (approved):

PR #2704, opened 2026-08-23
Add `theemathas` to cloud compute team

theemathas:
The dev desktops would be helpful for, among other things, running `cargo-bisect-rustc` (which, when ran locally, often runs into some fastly issue that's geographically local)


jieyouxu (approved):
Happy to vouch for @theemathas myself.

(I'm surprised release does not have Dev Desktops, but yeah.)

PR #2709, opened 2026-08-25
feat: add `wg-build-script` working group

weihanglo:
We've discussed this on Zulip DMs and t-cargo meemting (see [note](https://github.com/rust-lang/cargo-team/blob/bfc106cb25acca505d6183634b90423f350e839b/meetings/sync-meeting/2026-08-11.md?plain=1#L113-L119) and we'd like to form a new working group to discuss about the issues and evolutions around Cargo's build scripts.

Initially we expect to focus on topics like security, build determinism, and build script performance (supply chain security being of particular interest to me). The scope of this working group can be expanded, shrunk, or split when needed.

The way of working is not yet decided. It will be the first topic on the new Zulip channel (expect on `#t-cargo/wg-build-script`)

cc @rust-lang/cargo @LawnGnome @NobodyXu @celinval

### DIFF
--- a/teams/wg-build-script.toml
+++ b/teams/wg-build-script.toml
@@ -1,0 +1,27 @@
+name = "wg-build-script"
+subteam-of = "cargo"
+kind = "working-group"
+
+[people]
+leads = ["celinval", "weihanglo"]
+members = [
+    "LawnGnome",
+    "NobodyXu",
+    "celinval",
+    "epage",
+    "joshtriplett",
+    "ranger-ross",
+    "weihanglo",
+]
+alumni = []
+
+[[github]]
+orgs = ["rust-lang"]
+
+[website]
+name = "Build Script Working Group"
+description = "Improving the security, determinism, and performance of Cargo build scripts"
+zulip-stream = "t-cargo/build-script"
+
+[[zulip-groups]]
+name = "WG-build-script"


teams/wg-build-script.toml
```diff
@@ -0,0 +1,27 @@
+name = "wg-build-script"
+subteam-of = "cargo"
+kind = "working-group"
```
weihanglo, line 3: What is the difference between working group and project group? Which should we use?

teams/wg-build-script.toml
```diff
@@ -0,0 +1,27 @@
...
+]
+alumni = []
+
+[[github]]
+orgs = ["rust-lang"]
+
+[website]
+name = "Build Script Working Group"
+description = "Improving the security, determinism, and performance of Cargo build scripts"
+zulip-stream = "t-cargo/wg-build-script"
```
weihanglo: Is `wg-` prefix necessary? Some other WG don't seem to use that.

teams/wg-build-script.toml
```diff
@@ -0,0 +1,27 @@
+name = "wg-build-script"
+subteam-of = "cargo"
+kind = "working-group"
+
+[people]
+leads = ["celinval", "weihanglo"]
```
weihanglo, line 6: WG leads are currently me and @celinval, and me being a the liason to t-cargo, though probably unnecessary as we have many t-cargo members in this WG already.

teams/wg-build-script.toml
jieyouxu: It's not so clear cut in practice. WG seems fine for this purpose.

teams/wg-build-script.toml
jieyouxu: I don't think it's necessary for the zulip stream name, I think we only intend to use `WG-*` prefix for the zulip group name

weihanglo: Fixed. Thanks for the feedback!

jieyouxu (approved):

weihanglo:
@jieyouxu there is no Zulip channel created after this merged. Is there something wrong or did I overlook anything?

weihanglo:
Ah. I guess that is website display only.

jieyouxu:
Yeah, you'll need to create it :D (Or I can create it)

weihanglo:
>  You do not have permission to create channels. 

😭

jieyouxu:
@weihanglo created it seeded with WG-build-script members, see [#t-cargo/build-script > channel events @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/628857-t-cargo.2Fbuild-script/topic/channel.20events/near/619419520)

PR #2724, opened 2026-09-01
mailgun: enable mangling and syncing with subdomains

ubiratansoares:
## AI disclosure

I used Claude Code to get clarifications about these changes and deep search in the wild info on Mailgun's API rate limits.

---

Towards https://github.com/rust-lang/infra-team/issues/297

We tweak the mangling logic to handle a special subdomain we'll introduce to enable the migration (of lists provisioned by `team`) from Mailgun to Google Groups.

`sync` does not support dry-runs for Mailgun in CI, so I had to test these changes manually. 

```shell
echo "Preparing team static api"
cargo run --quiet static-api build
echo

echo "Dry-run changes against Mailgun"

MAILGUN_API_TOKEN=$(op read <path/to/secret>) \
EMAIL_PRIVATE_KEY=$(op read <path/to/secret>) \
  cargo run --quiet -- sync dry-run --services mailgun --src team-api
echo
```

A part of the output looks like this, I won't share it full since it'd disclose encrypted emails:

```
# .
# .
# previous output
#
[INFO  rust_team::sync::mailgun::api] deleting route with ID 5c5b479fa4764a000139448d
[INFO  rust_team::sync::mailgun::api] deleting route with ID 6321fe4584c515511ba78994
[INFO  rust_team::sync::mailgun::api] deleting route with ID 6321fe4565c9ab17d6be343e
[INFO  rust_team::sync::mailgun] creating list ^leah(?:\+.+)?@(?:legacy-lists\.)?rustconf\.com$
[INFO  rust_team::sync::mailgun] creating list ^infra-team(?:\+.+)?@(?:legacy-lists\.)?rust-lang\.org$
[INFO  rust_team::sync::mailgun] creating list ^community(?:\+.+)?@(?:legacy-lists\.)?rust-lang\.org$
[INFO  rust_team::sync::mailgun] creating list ^project-vision-doc-2025(?:\+.+)?@(?:legacy-lists\.)?rust-lang\.org$
#
# more output
# .
# .
```

This approach will delete and (re)create all 96 lists we currently have, rather than updating them in-place. This should be fine from the rate-limiting perspective for a single deploy (no official numbers, but 300 rpm mentioned on [issues like this](https://github.com/phillbaker/terraform-provider-mailgunv3/issues/15)). 

After lots of pain, I think this path worths compared with trying to update existing rules in-place and keeping guardrail checks to ensure consistency between the current and the new state (that was my first attempt). 

Happy to get input in case I missed something.

### DIFF
--- a/src/sync/mailgun/mod.rs
+++ b/src/sync/mailgun/mod.rs
@@ -15,6 +15,10 @@ const DESCRIPTION: &str = "managed by an automatic script on github";
 // Limit (in bytes) of the size of a Mailgun rule's actions list.
 const ACTIONS_SIZE_LIMIT_BYTES: usize = 4000;
 
+// Helper subdomain to use while we migrate lists from Mailgun
+// See https://github.com/rust-lang/infra-team/issues/297
+const MIGRATION_HELPER_SUBDOMAIN: &str = "legacy-lists";
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 struct List {
     address: String,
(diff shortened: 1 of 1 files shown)


src/sync/mailgun/mod.rs
```diff
@@ -223,11 +238,31 @@ mod tests {
...
+
+    #[test]
+    fn test_mangle_address_matching() {
+        for addr in [
+            "list@example.com",
+            "list@legacy-lists.example.com",
+            "list+bot@example.com",
+            "list+bot@legacy-lists.example.com",
+        ] {
+            let re = regex::Regex::new(&mangle_address("list@example.com").unwrap());
```
marcoieni: let's move this above the for loop so we construct the regex once

ubiratansoares: Done!

marcoieni (approved):

PR #2732, opened 2026-09-03
allow setting zulip stream history visibility

marcoieni:
In [#t-infra > ✔ DM groups in Zulip @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/.E2.9C.94.20DM.20groups.20in.20Zulip/near/621384411) they asked me to modify a zulip stream visibility.
However, it's a private stream and I'm not part of it, so I can't.
I also don't want to log in with rust-lang-owner.

Merging this PR should set this field via the API.

## Manual test
Unfortunately the dry run of github actions doesn't test zulip, because you cannot have a read-only Zulip token. 

I tested this by running `cargo run -- sync --services zulip print-plan` locally by provisioning an API key for my account.
I verified that no streams that I'm in are modified.

I also verified that a stream I'm in would be updated when editing the file locally:
```
    💻 Stream History Diffs:
    📝 Updating stream history:
      Name: t-crates-io/operations
      ID: 346296
      Public history: true
```

I had to edit the program locally to avoid errors like:

```
[ERROR rust_team] failed: Id of stream 'funding/private' not found. The stream probably doesn't exist and sync-team doesn't support creating it yet. Please create the stream manually and add the rust-lang-owner user to it.
```

## AI disclosure

I used GPT5.6-Sol with the codex harness to generate this change. I reviewed its output.

### DIFF
--- a/src/sync/zulip/api.rs
+++ b/src/sync/zulip/api.rs
@@ -265,6 +265,38 @@ impl ZulipApi {
         Ok(())
     }
 
+    /// Set whether a Zulip stream's history is public to new subscribers.
+    ///
+    /// See <https://zulip.com/api/update-stream#parameter-history_public_to_subscribers>.
+    pub(crate) async fn update_stream_public_history(
+        &self,
+        stream_id: u64,
+        public_history: bool,
+    ) -> anyhow::Result<()> {
+        log::info!("updating stream {stream_id} to public_history={public_history}");
+
+        if self.dry_run {
+            return Ok(());
+        }
+
+        let history_public_to_subscribers = public_history.to_string();
+        let mut form = HashMap::new();
+        form.insert(
+            "history_public_to_subscribers",
+            history_public_to_subscribers.as_str(),
+        );
+
+        self.req(
+            reqwest::Method::PATCH,
+            &format!("/streams/{stream_id}"),
+            Some(form),
+        )
+        .await?
+        .error_for_status()?;
+
+        Ok(())
+    }
+
     /// Perform a request against the Zulip API
     async fn req(
         &self,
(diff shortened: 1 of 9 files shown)


src/sync/zulip/api.rs
```diff
@@ -265,6 +265,38 @@ impl ZulipApi {
...
+        stream_id: u64,
+        public_history: bool,
+    ) -> anyhow::Result<()> {
+        log::info!("updating stream {stream_id} to public_history={public_history}");
+
+        if self.dry_run {
+            return Ok(());
+        }
+
+        let history_public_to_subscribers = public_history.to_string();
```
ubiratansoares, line 282: [Official docs ](https://apt.zulip.com/api/update-stream#parameter-history_public_to_subscribers )say

> It's an error for this parameter to be false for a public or web-public channel and when is_private is false.

Our TOML schema docs say we can set if to `false` but do not mention this failure mode; do we need to guard here?

Kobzol: We currently only support private streams in the `team` DB though, so it is kinda implied.

src/sync/zulip/mod.rs
```diff
@@ -210,14 +253,19 @@ impl Diff {
...
+        }
         for stream_membership_diff in &self.stream_membership_diffs {
             stream_membership_diff.apply(sync).await?;
         }
         Ok(())
     }
 
     pub(crate) fn is_empty(&self) -> bool {
-        self.user_group_diffs.is_empty() && self.stream_membership_diffs.is_empty()
+        self.user_group_diffs.is_empty()
```
Kobzol: While we're modifying this, could you please use destructuring here, to make it harder to forget when we add new fields?

Kobzol (approved):

src/sync/zulip/mod.rs
marcoieni: done in https://github.com/rust-lang/team/pull/2732/changes/1f92a3d61f695e60bc0c5784e7540b2289e6469e

ubiratansoares (approved):

PR #2741, opened 2026-09-10
Remove `jieyouxu` from the `rustfmt-contributors` subteam

jieyouxu:
Follow-up to https://github.com/rust-lang/team/pull/2739, I completely forgot about the contributors subteam split after compiler/libs got rid of `-contributors` subteams lol.

cc @ytmimi

### DIFF
--- a/teams/rustfmt-contributors.toml
+++ b/teams/rustfmt-contributors.toml
@@ -7,7 +7,6 @@ members = [
     "Manishearth",
     "fee1-dead",
     "camsteffen",
-    "jieyouxu",
     "estebank",
     "matthewhughes934",
 ]


ytmimi (approved):
Oh I didn't realize I needed to remove you from `rustfmt-contributors`.

jieyouxu:
> Oh I didn't realize I needed to remove you from `rustfmt-contributors`.

Yeah don't worry about it, I forgot about it too 😆

marcoieni:
you don't want to add yourself to alumni?

ytmimi:
Oh yeah, we should probably do that instead.

jieyouxu:
> you don't want to add yourself to alumni?

https://github.com/rust-lang/team/pull/2739 moved me to the `rustfmt` team, this PR just removes me from the contributors subteam so I'm not in both 😆 

When a member gets moved from the contributors subteam -> the parent team, we generally just do a move, and not also mark them as alumni (it doesn't really make sense, lol). See https://github.com/rust-lang/team/pull/2729 for an example of r-a-contributors -> r-a team move.

ytmimi:
First time Iv'e been involved with moving someone from a subteam to the parent team, so if we typically just move them then I think it's fine to follow that precedent.

jieyouxu:
> First time Iv'e been involved with moving someone from a subteam to the parent team, so if we typically just move them then I think it's fine to follow that precedent.

Yeah, that's usually what we do for contributors subteam => parent team moves 👍

marcoieni (approved):
approving even if I'm not a team repo admin because it seems an easy change.

PR #2726, opened 2026-09-02
Ensure that only `rust-lang-owner` owns our tracked crates

Kobzol:
This is in preparation for https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Crate.20yanking.2Funyanking.20on.20Zulip/with/620911997.

This removes all individual and team ownership of crates.io crates tracked in `team`, and keeps only `rust-lang-owner` as their sole owner.

PR #2742, opened 2026-09-10
Assign owner teams to crates.io crates

Kobzol:
In preparation for https://github.com/rust-lang/team/pull/2726, this PR assigns owner teams to all crates.io crates currently recorded in the `team` DB. I assigned mostly the teams that had access to the repo from where the crate is published.

### DIFF
--- a/repos/rust-lang/annotate-snippets-rs.toml
+++ b/repos/rust-lang/annotate-snippets-rs.toml
@@ -16,5 +16,6 @@ required-approvals = 0
 crates = ["annotate-snippets"]
 publish-workflow = "release.yml"
 publish-environment = "publish"
+teams = ["cargo"]
 
 [environments.publish]
--- a/repos/rust-lang/ar_archive_writer.toml
+++ b/repos/rust-lang/ar_archive_writer.toml
@@ -14,3 +14,4 @@ tags = ["*"]
 crates = ["ar_archive_writer"]
 publish-workflow = "publish.yml"
 publish-environment = "publish"
+teams = ["compiler"]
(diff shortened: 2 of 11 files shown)


repos/rust-lang/annotate-snippets-rs.toml
```diff
@@ -16,5 +16,6 @@ required-approvals = 0
 crates = ["annotate-snippets"]
 publish-workflow = "release.yml"
 publish-environment = "publish"
+teams = ["cargo"]
```
fmease, line 19: This crate is not only used by Cargo but also by rustc to render diagnostics.

Obviously, WG-diagnostics members still have write access and personally speaking I don't have enough context to know what qualifies a crater owner. I guess WGs can't be made owners? In any case, I'm wondering whether this should be T-cargo+T-compiler (I guess that would also require updating `[access.teams]`)?

It's somewhat recent that annotate-snippets is finally used in rustc, too, so the existing access rules might need updating in general? After all, nowadays a bug in annotate-snippets can negatively affect rustc just as badly as Cargo.

That's of course not for me to decide.

Kobzol: Yeah I was thinking the same, that this should also be owned by t-compiler. Regarding WGs, we can make them owners, but I'd rather not, if there is an existing team that could serve as the owner.

marcoieni (approved):

repos/rust-lang/ar_archive_writer.toml
```diff
@@ -14,3 +14,4 @@ tags = ["*"]
 crates = ["ar_archive_writer"]
 publish-workflow = "publish.yml"
 publish-environment = "publish"
+teams = ["compiler"]
```
bjorn3, line 17: :+1:

PR #2745, opened 2026-09-10
enable crabwatch on all repositories except rust

marcoieni:
As discussed in [#t-infra > crabwatch on infra repos](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/crabwatch.20on.20infra.20repos/with/620361183), crabwatch seems to work well, so now we want to enable it for more repositories.

before merging:
- [x] announce this change in `t-infra/announcements` because people might need to trigger CI by pushing a commit in order to get this job running (and therefore their PR can be merged).
- [x] verify `GITHUB_TOKEN=$(gh auth token) cargo run -- analyze --org rust-lang` works without errors in the crabwatch repo.

## Alternative

Another approach is to:
* run crabwatch on all repositories (i.e. with `crabwatch = true` or where `crabwatch` is undefined)
* don't run crabwatch in repositories where `crabwatch = false`.

This approach has the pro that we don't set the crabwatch repository everywhere.
I guess it doesn't change much, but if the reviewer has a preference we can change the ruleset rule.

## AI disclosure

I used GPT6-Astra with the codex harness to generate this change.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -21,6 +21,11 @@ independent-github-orgs = [
     "rust-embedded",
 ]
 
+# Crabwatch is enabled for all non-archived repositories except those listed here.
+crabwatch-deny-list = [
+    "rust-lang/rust",
+]
+
 permissions-bors-repos = [
     "bors-kindergarten",
     "rust",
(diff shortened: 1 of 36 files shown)


ubiratansoares (approved):

marcoieni:
when we enable this job also in rust-lang/rust, we can just remove this property and run this job in every repository, with no mechanism to opt-out.

PR #2747, opened 2026-09-12
Add TechnoPorg to the RISC-V team

TechnoPorg:
I'd like to get more involved in Rust, and since this is related to my work projects, I think it's a good place to start!

### DIFF
--- a/teams/risc-v.toml
+++ b/teams/risc-v.toml
@@ -8,4 +8,5 @@ members = [
     "kraj",
     "tblah",
     "sweiglbosker",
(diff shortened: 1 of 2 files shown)


teams/risc-v.toml
```diff
@@ -8,4 +8,5 @@ members = [
     "kraj",
     "tblah",
     "sweiglbosker",
```
jieyouxu, line 10: cc @denisvasilik @kraj @tblah @sweiglbosker

PR #2752, opened 2026-09-15
Archive `packed_simd`

clarfonthey:
This was superceded by rust-lang/portable-simd.

Note that there is also the `packed_simd` crate on crates.io, although that one is also under `rust-lang-owner` so I'm not sure if there's anything extra we need to do.


Kobzol:
It would be good, before archiving, to modify its README to provide a warning about the deprecation and the link to the `portable-simd` repo.

clarfonthey:
I can do that, actually, since libs currently has write access.

(done)

calebzulawski (approved):
I think formally deprecating it via archiving is a good idea.

clarfonthey:
I think that qualifies as team lead approval?

jieyouxu (approved):

PR #2755, opened 2026-09-15
Add `libs-ping` group

clarfonthey:
Relevant discussion: [#t-libs/meta > Pinging all of libs for libs FCPs](https://rust-lang.zulipchat.com/#narrow/channel/253831-t-libs.2Fmeta/topic/Pinging.20all.20of.20libs.20for.20libs.20FCPs/with/624405709)

Since `lang-advisors` has a bit more meaning than just people who want to be pinged (due to the presence of a private chat), decided to just name this `libs-ping` and make it only a marker group for now. Also added a comment so people are aware of who exactly should be in the group.

For now, I've added:

* Existing `libs` lead
* Existing `libs-fcp` team
* All libs members on the review rotation for `rust-lang/rust` (note: waiting on below)

Which should be sufficient for now.


clarfonthey:
Ran into the reflexivity problem and forgot to include self as one of the members, hence the force-push.

clarfonthey:
Also, I guess, to verify for the people who are on the review rotation who didn't express a desire to be on the FCP team or this list, to verify that you still want to be @-mentioned. Please react with a :+1: or :-1: on this comment to indicate if you want to be on the list. (I don't think we can do regular checkboxes here, since I don't think you can edit comments on this repo.

Verified:

* [x] @aapoalas
* [x] @Darksonn
* [x] @joboet
* [x] @JohnTitor
* [x] @LawnGnome
* [x] @workingjubilee

Darksonn:
What volume of pings would I be signing up for here?

clarfonthey:
FCPs and backports, basically.

Amanieu (approved):
LGTM! Have you updated the message from rfcbot to ping this team instead?

teams/libs-ping.toml
```diff
@@ -0,0 +1,38 @@
+# should include:
+# * libs-fcp
+# * libs leads
+# * any other libs member who wants to be included
```
jieyouxu, line 5: Can you leave a quick TL;DR for the intended purpose of this marker team as a comment?

hanna-kruppe:
Please add me as well

LawnGnome:
I'm happy to receive pings.

clarfonthey:
Everyone on the list is confirmed so we can merge now.

tgross35:
Not worth doing here but I wonder if this could somehow include libs members by default with opt-out rather than opt-in, so we don't forget to add new members or ask if they're interested. Then they can of course leave the group if it's too noisy.

clarfonthey:
So, to be clear, I think it would be a good policy to mention this for new people, but also part of the goal with the comments on the file is so that people are comfortable accepting libs members just adding/removing themselves. Similarly to how listing GitHub sponsors on the website is also opt-in, for valid reasons.

jieyouxu (approved):

PR #2758, opened 2026-09-16
Give the Rustup team access to `rustup-components-history`

Kobzol:
Suggested [here](https://github.com/rust-lang/rustup/issues/5085#issuecomment-5694807481).

Since the repo has a lot to do with Rustup, I think that the team should have access.

CC @rust-lang/rustup


ubiratansoares:
@Kobzol should we make the most of this PR and eventually add simple branch protection for this repo?

Kobzol:
Yeah, good point, I was thinking the same. While at it, I also switched GH pages to the modern workflow mechanism, I will switch the GH pages workflow to use it.

ubiratansoares (approved):

PR #2760, opened 2026-09-16
Remove libs from wg-allocators repo

clarfonthey:
Since this appears to be the exception among WG repos. cc @nia-e as WG lead


nia-e:
yep, lgtm for me

nia-e (approved):

tgross35:
For what it's worth this was intentional https://github.com/rust-lang/team/commit/c603db1aefeb3b2f4092b5154a765b541a7676d1 libs members wanted to help triage the huge backlog of issues

clarfonthey:
That makes a lot of sense, although I also know that the working group has been inactive for a while and that we have recently added a few new people onto it. Considering the low barrier to entry I would just say that libs members should add themselves to the WG if they're interested in helping out.

I mostly proposed this as I was looking into what repos libs maintains, and this felt like an outlier.

jieyouxu:
Let us know the conclusion. I feel like it's fine leaving libs on the repo? But anyway if the conclusion is to proceed with the removal we can merge this.

nia-e:
either is fine, but let's merge this i think. the motivating reasoning for having libs is gone and the barrier to entry to the wg is quite low

jieyouxu (approved):

